PR #29, opened 2016-06-01
setColumns doesn't update table state in cookie

MxFulder:
fixed


olifolkerd:
Good spot!, i have just committed a further update to optimise that bit of code so the ***_colLayout*** and ***_setColCookie*** functions are only called once at the end of the ***setColumns*** function rather that the duplicates that were at the end of each branch of the if statment.

Issue #273, opened 2017-04-21
Bug: Editable is false, but it still lets me edit


vasilevich:
Hello! thanks again for your work, there seems to be a bug in the cells,
here are my columns:
```
var columns = [ 
    {
        title: "תאריך",
        field: "date",
        align: "right",
        sorter: "date",
        sortable: true,
        headerFilter: true,
        editor: dateEditor,
    },
    {
        title: "מצב",
        field: "state",
        align: "right",
        sortable: false,
        headerFilter: true,
        mutator: stateMutator,
        editor: stateEditor,
        accessor: function (value, data) {
            for (var key in states) {
                if (states[key] == value)
                    return parseInt(key);
            }
            return 0;
        }
    },
    {
        title: "פוליסה",
        field: "polisa",
        align: "right",
        sorter: "number",
        sortable: true,
        headerFilter: true,
        editor: numberEditor,
    },
    {
        title: "הזמנה",
        field: "order",
        align: "right",
        sorter: "number",
        headerFilter: true,
        editor: numberEditor
     
  
    }
];
```

as you can see, editable is not set(thus its false
,
it seems that whenever the editor function is set by me, the cells become editable regardless of weather the "**editable**" variable is set to true or false, which is an undesired behavior because the editor is also used as the **filter**, so the fact that I set the editor to my own made function should not make columns editable, can you please look into that? thank you!

vasilevich:
I found a fix, will do a pull request shortly.

olifolkerd:
Hey,

If you only want to use a custom editor for the header filters but not for editing the cells then you should set the ***headerFilter*** value to that of the editor and not set the ***editor*** value at all.

```js
    {
        title: "הזמנה",
        field: "order",
        align: "right",
        sorter: "number",
        headerFilter: true,
        editor: numberEditor
   }
```

Full documentation on this can be found [here](http://olifolkerd.github.io/tabulator/docs/#header-filters)

Cheers

Oli

vasilevich:
sorry about that, guess I missed it, thank you!

PR #274, opened 2017-04-21
see issue: https://github.com/olifolkerd/tabulator/issues/273

vasilevich:
Hello! thanks again for your work, there seems to be a bug in the cells,
here are my columns:
```
var columns = [ 
    {
        title: "תאריך",
        field: "date",
        align: "right",
        sorter: "date",
        sortable: true,
        headerFilter: true,
        editor: dateEditor,
    },
    {
        title: "מצב",
        field: "state",
        align: "right",
        sortable: false,
        headerFilter: true,
        mutator: stateMutator,
        editor: stateEditor,
        accessor: function (value, data) {
            for (var key in states) {
                if (states[key] == value)
                    return parseInt(key);
            }
            return 0;
        }
    },
    {
        title: "פוליסה",
        field: "polisa",
        align: "right",
        sorter: "number",
        sortable: true,
        headerFilter: true,
        editor: numberEditor,
    },
    {
        title: "הזמנה",
        field: "order",
        align: "right",
        sorter: "number",
        headerFilter: true,
        editor: numberEditor
     
  
    }
];
```

as you can see, editable is not set(thus its false
,
it seems that whenever the editor function is set by me, the cells become editable regardless of weather the "**editable**" variable is set to true or false, which is an undesired behavior because the editor is also used as the **filter**, so the fact that I set the editor to my own made function should not make columns editable, can you please look into that? thank you!

### DIFF
--- a/tabulator.js
+++ b/tabulator.js
@@ -2534,7 +2534,7 @@
 			self._generateTooltip(cell, item, column.tooltip);
 
 			//match editor if one exists
-			if (column.editable || column.editor){
+			if (column.editable){
 				if(column.editor){
 					var editor = column.editor;
 				}else{


olifolkerd:
Hey,

If you only want to use a custom editor for the header filters but not for editing the cells then you should set the ***headerFilter*** value to that of the editor and not set the ***editor*** value at all.

```js
    {
        title: "הזמנה",
        field: "order",
        align: "right",
        sorter: "number",
        headerFilter: true,
        editor: numberEditor
   }
```

Full documentation on this can be found [here](http://olifolkerd.github.io/tabulator/docs/#header-filters)

Cheers

Oli

PR #280, opened 2017-04-25
Added define of jQuery to work as other jQuery plugins.

vasiliyaltunin:
So now if even jQuery is not loaded when tabulators loads, but loads later all works fine. It works fine in Joomla component.

### DIFF
--- a/tabulator.js
+++ b/tabulator.js
@@ -23,7 +23,6 @@
 }(function ($, undefined) {
  	'use strict';
 
-
  	//polyfill for Array.find method
  	if (!Array.prototype.find) {
  		Array.prototype.find = function (predicate, thisValue) {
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey,

Sorry for the delay in getting back to you on this one.

I like the idea there, that is a clever little widget, i will have a play this weekend and approve the merge if all goes well.

Thanks for contributing!

olifolkerd:
Do you have any examples of this working correctly? i have just tried it out and it dosn't see, to work if jQuery is loaded after the function.

vasiliyaltunin:
Sorry at moment i wrote commit i may be thinkin about other thing or look wrong place. Point is, that with this code factory will correctly create object at runtime and without this code in joomla tabulator not works at all.

Issue #369, opened 2017-06-15
paginatorDataReceived should also support object properties


liviu-vasut:
When parsing remote data the current code tries to get the number of pages like this:

    this.max = parseInt(data[this.paginationDataReceivedNames.last_page]);

This only accepts a page count specified as:

    { ..., count: 4, data: [...], ... }

where the data is in response.data and the page count is in response.count. If one would like to load data from elasticsearch, for example, it would get this response:

    { ..., hits: { total: 16, ... }, ... }

where the data is in response.hits.hits and the record count is in response.hits.total. Note that no page count is returned and only the record count. This format is not supported now.
The code could be modified so that the configuration also accepts a function that takes the response and the page size and returns the page count.

olifolkerd:
You are in luck, there is already a function available to help you there. The ***ajaxResponse*** callback lets you reformat your data into the format expected by Tabulator:
```js
$("#example-table").tabulator({
    ajaxResponse:function(url, params, response){
        //url - the URL of the request
        //params - the parameters passed with the request
        //response - the JSON object returned in the body of the response.

        return response; //return the response formatted response object to tabulator
    },
});
```

Documentation for this feature can be found in the [Callbacks Documentation](http://olifolkerd.github.io/tabulator/docs/3.0/#callbacks)

liviu-vasut:
Well, I have read the documentation for the **ajaxResponse** but it says that the response should be an array, so that's where my confusion should have started from. Somehow, when checking the **paginationDataReceived** documentation I missed the remote server response format part, or I just did not make the connection with the **ajaxResponse**. I reviewed the documentation today and you are right, no changes were required, so sorry for the noise...

olifolkerd:
no problem, I will see if i can update the documentation to make it a bit clearer in the remote pagination documentation what you should do.

liviu-vasut:
Great, thanks!

PR #370, opened 2017-06-15
fix for #369

liviu-vasut:
added option to configure a function that provides the last page. For example this should work:
paginationDataReceived:{ "last_page": "last" }
and also:
paginationDataReceived:{ "last_page": function(data, pagesize){ return Math.ceil(data.length/pagesize); } }

### DIFF
--- a/src/js/extensions/page.js
+++ b/src/js/extensions/page.js
@@ -360,9 +360,16 @@ Page.prototype._getRemotePageAuto = function(){
 
 
 Page.prototype._parseRemoteData = function(data){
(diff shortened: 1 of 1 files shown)


olifolkerd:
There is already a more general solution to this problem in the form of the ***ajaxResponse*** callback.

I have responded to you issue with more details.

Thanks for getting involved, i look forward to hearing from you again.

Cheers

Oli

liviu-vasut:
Thanks for the reply, like I said in the issue, I've got confused by the documentation of **ajaxResponse**. Anyway, great library, thanks for sharing it ;).

PR #539, opened 2017-08-28
invisible columns / persistentLayout bug

bbrand84:
hidden colums caused error when "persistentLayout: true".
In my case, the concerned invisible field was an index as well, I did not test if this is important.
(the change should have happened in "src" instead of "dist" I think, sorry for that)


bbrand84:
just tried to update the "src/js/column.js". Seems to have bee fixed already. Dist folder seems out of sync ??

PR #600, opened 2017-09-28
if size is a string then concatenation occurs

scarabus:
If the user passes in a String as the paginationSize then the value of Start is calculated correctly but the value of End is incorrect as string concatenation occurs.


olifolkerd:
nice addition, Thanks :)

Let me know if you have any other tweaks :D

PR #653, opened 2017-10-20
feat: add dataReceiver option to parse remote data directly

CeceWall:
In my case, pagination response data only has a property indicate total rows instead of total pages. thus I add a dataReceiver function to modify data.

### DIFF
--- a/src/js/extensions/page.js
+++ b/src/js/extensions/page.js
@@ -14,6 +14,7 @@ var Page = function(table){
 	this.page = 1;
 	this.max = 1;
 	this.paginator = false;
+	this.dataReceiver = false;
 };
 
 //setup pageination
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @CeceWall,

There is already a callback to handle processing of data received over ajax, the ***ajaxResponse*** callback, which is called after any successful callback and lets you manipulate the data before it is read into the table if you need too:
```js
$("#example-table").tabulator({
    ajaxResponse:function(url, params, response){
        //url - the URL of the request
        //params - the parameters passed with the request
        //response - the JSON object returned in the body of the response.

        return response; //return the response data to tabulator
    },
});
```

What additional functionality does your callback add?

Cheers

Oli

CeceWall:
i'm too careless to notice this function, it's enough for me, thanks very much.

PR #821, opened 2018-01-10
Adding optional argument whether to refresh grid after filter change

Tomkaa:
Use case:
With single button click you need both: to remove filter and add new filter. Since it is needed to call both `addFilter` and `removeFilter` - 2 HTTP requests will be made. 
This pull request would improve this situation by adding boolean property, whether to refresh data after filter change. Usage:
`
$("#example-table").tabulator("removeFilter", "age", ">", 15, false); /* <- false indicates not to refresh the data */
$("#example-table").tabulator("addFilter", "age", ">", 10);
`
Ajax request would only be done after the `addFilter`.
Parameter to refresh data by default is true (represent the way it is right now), so it is backwards compatible.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -732,10 +732,12 @@
 	 		},
 
 	 		//add filter to array
-	 		addFilter:function(field, type, value){
+	 		addFilter:function(field, type, value, triggerRefresh = true){
 	 			if(this.extExists("filter", true)){
 	 				this.extensions.filter.addFilter(field, type, value);
-	 				this.rowManager.filterRefresh();
+					if(triggerRefresh) {
+					 	this.rowManager.filterRefresh();
+					}
 	 			}
 	 		},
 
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @Tomkaa 

Thanks for your suggestion, and sorry for taking so long to get back to you.

I don't feel that adding additional arguments is necessary in this case, as it complicates the system and i would prefer to keep the arguments in each function to a minimum.

You should use the existing ***setFilter*** method to replace the current filter set in one go.

If you would like to update the current filter value of an existing filter then can i suggest the creation of a ***replaceFilter*** function that would replace a filter with a matching field and filter with the new value.

That would me a more intuitive way of managing it and more inline with the rest of Tabulators API.

If you would like to implement a ***replaceFilter*** function i  would be happy to receive the pull request.

Cheers

Oli :)

PR #870, opened 2018-02-09
Make validators optional by default

lgraziani2712:
This PR allows to have validators like min, max in optional columns.

E.g.

```js
$('#example-table').tabulator({
    columns:[
        {
            field: 'optional_1',
            title: 'My optional prop',
            editor: true,
            validator: ['integer', 'min:1', 'max:12']
        },
        {
            field: 'required_1',
            title: 'My required prop',
            editor: true,
            validator: ['required', 'integer', 'min:1', 'max:12']
        },
    ],
});
```

If `optional_1` is an empty string, null or undefined, it won't complain.

### DIFF
--- a/src/js/extensions/validate.js
+++ b/src/js/extensions/validate.js
@@ -87,76 +87,102 @@ Validate.prototype.validate = function(validators, cell, value){
 	return valid.length ? valid : true;
 };
 
-Validate.isEmptyOrValidate = function(callback) {
-	return function(cell, value, parameters){
-		// If it isn't empty then validate
-		return typeof value !== "undefined" || value === null || value === ""
-			|| callback(cell, value, parameters);
-	};
-};
 Validate.prototype.validators = {
 
 	//is integer
-	integer: Validate.isEmptyOrValidate(function(cell, value, parameters){
+	integer: function(cell, value, parameters){
+		if(value === "" || value === null || typeof value === "undefined"){
+			return true;
+		}
 		value = Number(value);
 		return typeof value === 'number' && isFinite(value) && Math.floor(value) === value;
-	}),
+	},
 
 	//is float
-	float: Validate.isEmptyOrValidate(function(cell, value, parameters){
+	float: function(cell, value, parameters){
+		if(value === "" || value === null || typeof value === "undefined"){
+			return true;
+		}
 		value = Number(value);
 		return typeof value === 'number' && isFinite(value) && value % 1 !== 0;;
-	}),
+	},
 
 	//must be a number
-	numeric: Validate.isEmptyOrValidate(function(cell, value, parameters){
+	numeric: function(cell, value, parameters){
+		if(value === "" || value === null || typeof value === "undefined"){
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @lgraziani2712 

Thanks for your suggestion, but i would prefer to keep validator function isolated between validators, especially as the ***required*** validator checks for an empty string, which needs to be considered as an option for the ***minLength*** validator

Under what circumstance are you managing to pass a null value into a validator?

Cheers

Oli

lgraziani2712:
Hello Oli, thank you for your response!

I was trying to make a table with an optional column but with restricted values. Just integers from 1-12. I cannot enforce a default value, and I need to allow empty values even if they had one before.

Here is an example: https://codepen.io/lgraziani/pen/KQWQwa?editors=0010

>  but i would prefer to keep validator function isolated between validators

Would you be OK if instead of `!this.required(value)` be `!value`?

Luciano.

olifolkerd:
Hey @lgraziani2712 

I would go a step further than you have done, use the code from the required formatter in each of the validators, including string, numeric float, integer and unique, an empty value should be allowed through them all, the ***required*** validator should be used to say they cant be empty.

The reason i ask not to call another validator from within that validator is that all other callbacks (editors, formatters, etc) are called in the scope of their table, not the object they are defined in, and i will be making the same change to the validators in tomorrows 3.4 release.

If you can get the changes in today i would be happy to include it in the official release tommorow.

Cheers

Oli :)

lgraziani2712:
Hello Oli! Like this?

I made a codepen with the fix: https://codepen.io/lgraziani/pen/bLqOYP?editors=0010

olifolkerd:
Hey @lgraziani2712 

You are there in theory, though you haven't put the ***isEmptyOrValidate*** function on the prototype. and the design pattern dosn't follow the rest of Tabulator, if you could just literally put the null/undefined check code  in each validator i would be happy to approve the pull request. eg:

```js
//maximum string length
maxLength:function(cell, value, parameters){
        if(value === "" || value === null || typeof value === "undefined"){
            return true;
        }else{
            return String(value).length <= parameters;
        }
},
```

The exception to this would be the minLength formatter which doesnt want to check for an empty string in the null check.

I know if isnt great for code reusue, but the point is that each of the validators is functionally isolated from one another and dosn't try and access external functionality. it also makes it easier for other users to tweak them as they see fit.

Cheers

Oli

lgraziani2712:
Understood!

> The exception to this would be the minLength formatter which doesn't want to check for an empty string in the null check.

Hmm, if I don't check for empty string, it will make the column required.

lgraziani2712:
I updated the pen with that constraint https://codepen.io/lgraziani/pen/bLqOYP?editors=0010

olifolkerd:
Hey @lgraziani2712 

That is looking great, Pull request approved!

I will include it in tomorrows 3.4 release

Thanks for the contribution, keep up the good work!

Cheers

Oli :)

PR #979, opened 2018-03-21
Add cell instance verification to element's focus evt

### LINKED ISSUE #966: Too much recursion error from v3.4.3
Hello!

When the TAB key is pressed while editing an input and the validation fails, tabulator enters in an infinite loop.

Here is a pen with the steps to reproduce the error: https://codepen.io/lgraziani/pen/EEyxdY

I think this line creates the infinite loop:

https://github.com/olifolkerd/tabulator/blob/f88eb7262b315c3e8e14b21fca40a79b46b06a5a/src/js/extensions/edit.js#L113-L117

Thank you! Luciano.

lgraziani2712:
Fix #966 

On `element.on('focus')` and before invoking the `self.edit()`, it checks if `self.currentCell` exists and if is a different cell than the `cell` variable.

- If exists and it's a different cell it means is the previous cell, so I store this variable locally.
- If not exists or if both points to the same cell, I store the `cell` variable locally.

Then I set `self.currentCell = false;`

Last I call `self.edit` with the local variable.

Codepen: https://codepen.io/lgraziani/pen/LdyEKP

It works in Firefox and Chrome.


lgraziani2712:
By the way, this "bug" it also happens when the editor field never calls the `success` or `cancel` functions. Quite common on custom editors.

olifolkerd:
Hey @lgraziani2712 

Thanks for the suggestion, unfortunately this fix doesn't really address the problem which was due to the way the edit function was being called through a looping focus event.

The proposed solution negated several of the functions of the validator.

Calling the ***success*** or ***cancel*** functions correctly is a requirement of a custom editor. I will make sure the documentation is updated to reflect this. 

I have just pushed an update to the master branch to address a number of issues with the editor/validator looping (which also fix this issue) and will include it in a patch release this afternoon. I have also added in a catch all that will cancel an edit when another edit is called if the editor has not handled the loss of focus correctly.

As a side note, I notice you were comparing the `column.field` property to check they were the same cell.  you could just directly compare the cells `self.currentCell === cell` to avoid situations where they are cells in the same column.

Thanks for the contribution.

Cheers

Oli :)

PR #1022, opened 2018-04-09
Include all columns of the definition

fortrax:
If the definition columns are merged with the persistentLayout columns, we need to see newly added columns. They are simply added to the right.

### DIFF
--- a/dist/js/tabulator.js
+++ b/dist/js/tabulator.js
@@ -15403,6 +15403,14 @@ var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol
         }
       });
 
+      oldCols.forEach(function (column, to) {
+          var from = self._findColumn(newCols, column);
+
+          if (!from) {
+              output.push(column);
+          }
+      });
+
       return output;
     };
 
--- a/src/js/extensions/persistence.js
+++ b/src/js/extensions/persistence.js
@@ -89,6 +89,16 @@ Persistence.prototype.mergeDefinition = function(oldCols, newCols){
 		}
 
 	});
+	oldCols.forEach(function (column, i) {
+		var from = self._findColumn(newCols, column);
+		if (!from) {
+			if(output.length>i){
+				output.splice(i, 0, column);
+			}else{	
+				output.push(column);
+			}
+		}
+	});
 
 	return output;
 };


fortrax:
Hi Oli,

I think, we can agree that the present behaviour is worst case - the user is not informed that the model shows a new column now. For me, it's not acceptable that the user should loose all configuration just because a new column is available. So changing the persistentLayoutID is a no-go. It feels only natural to add the columns in the rear. I've seen a lot of table control implementations in my life and this was the general way out of this problem. Do you think the new hash comparison will cover my needs? Or can you think of another way to realize that without patching your code?

Thanks & cheers,
Andreas

olifolkerd:
Hey @fortrax 

Don't get me wrong I agree there should be some improvement to the system, I just have to approach things holistically looking at all possible usage cases.

And would be happy to approve this pull request with a couple of tweaks.

In the case of this pull request I think we can get there if you make it so that instead of adding the missing columns on at the end of the column list, then are inserted at their previous index or on the end if the number of columns has been reduced, eg:
```js
 oldCols.forEach(function (column, i) {
     var from = self._findColumn(newCols, column);
        if (!from) {
            if(output.length>i){
                 output.splice(i, 0, column);
            }else{
                output.push(column);
            }
        }
});
```

You would also need to make the change in the `/src/js/extensions/persistence.js` file instead of on the `dist/js/tabulator.js` file. this is because everything in the ***dist*** folder is generated by a gulp command.

So the procedure to follow would be to make your changes to the `persistence.js` file, then run the ***gulp*** command in the tabulator root directory, then submit the pull request that should contain the `peristence.js`, `tabulator.js` and `tabulator.min.js` file.

If you are able to make these tweaks in the next couple of days, I can include them in the official v3.5 release which will be coming out towards the end of the week.

Thanks for your input, and it is great to see people wanting to contribute towards Tabulator!

Cheers

Oli :)

fortrax:
Ok, sounds great. I have to get into the Github/Gulp world, though. Can do it tomorrow.

fortrax:
I am quite proud of myself as I managed to install node.js and the necessary gulp modules :-). Gulp generated the tabulator.js file successfully but then crashes in the uglify: 
[Error] GulpUglifyError: unable to minify JavaScript
Caused by: SyntaxError: Unexpected token: name (parentIndex)
C:\tabulator-master\dist\js\tabulator.min.js
Line: 217

Looked it up in StackOverflow and it might be connected with ES6?!

olifolkerd:
Hey @fortrax 

Glad to hear you are succeeding with the new tools.

All code on the master branch compiles successfully so either npm hasnt installed the correct dependencies or there is isuses in the newly added code.

once you have installed npm on your computer, there are onlt two commands you should need to run to get tabulator compiled:
```npm install```
then
```gulp```

any other commands would be unnecessary, can i ask what commands you have been running?

Cheers

Oli :)

fortrax:
Oops, I got that wrong. I installed all the single modules with -save-dev like I found in an arbitrary web page. Now I tried
`npm uninstall`
in several variations. I also uninstalled the gulp-uglify and
`npm install`
re-installed it automatically. But the error I quoted above remains. Maybe it's some version of npm or gulp-uglify or anything? Sorry for the trouble. In the end it's more work for you than just including the fix you proposed yourself ;-)

olifolkerd:
Hey @fortrax 

Not a problem, it is a good learning experience for you.

I would recommend you delete the repo (keeping a copy of your changes elsewhere) then re-clone it, and run `npm install` just to make sure you have a clean slate. as your previous command may well have altered some other files which is causing the issue.

Cheers

Oli :)

fortrax:
That did the trick! When I replace the code of tabulator.js in my fork, though, he says there are to many changes to be shown. I better wouldn't commit that, right?

olifolkerd:
The only file you want to manually change is persistence.js. the let gulp
update tabulator.js and tabulate.min.js. then you should commit all 3 files
togeather.

Make sure you are on the latest version of the branch before committing.

Cheers

Oli

fortrax:
Sorry, Oli, I miss the link. What piece of software "commits" files on my file system to GitHub?

fortrax:
Ok, I think I've anwered that question myself, the missing link is GIT. So I installed the Github desktop which works fine. But still gulp creates strange deltas in tabulator.js. The attached screenshot will clarify this.
githubdesktop

olifolkerd:
Hmm, that is odd.

Could you try making your changes on the 3.5 branch instead of master and see if that works?

I can 100% guarantee that the 3.5 branch should compile correctly.

Cheers

Oli :)

olifolkerd:
Hey @fortrax 

Thanks for updating the persistence.js, the tabulator .js is still showing the old changes, but i will accept the pull request at the tabulator.js will be overwritten when i merge in 3.5 in a couple of days.

Thanks for your contribution, it should hopefully make persistent tables that bit more predictable.

Cheers

Oli :)

fortrax:
Pleasure! Now that I learned more or less how to do it, you will hear from me again. Thank you!

Issue #1047, opened 2018-04-15
cell.getRow().delete() deleting the row but erroring too


answerquest:
Hi, this this happening since v3.5.0, tested with source files downloaded just half an hour ago.

For this column definition:
```
{formatter:"buttonCross", align:"center", title:"del", headerSort:false, cellClick:function(e, cell){
	cell.getRow().delete();
	// other commands 
}}
```
On clicking the red cross the row *is being deleted*, but the function errors out at the same time and other commands after it do not run anymore. In the browser console, an error is reported:

```
Uncaught ReferenceError: index is not defined
    at Row.delete (tabulator.js:4855)
    at RowComponent.delete (tabulator.js:4272)
    at Object.cellClick (misc.js:65)
    at HTMLDivElement.<anonymous> (tabulator.js:5104)
    at HTMLDivElement.dispatch (jquery-3.3.1.min.js:2)
    at HTMLDivElement.y.handle (jquery-3.3.1.min.js:2)
```
I'm seeing the same effect happening from two different pages/scripts using the `cell.getRow().delete()` command this way. All errors lead to ` at Row.delete (tabulator.js:4855)`.  
Looking at the next function's lines my guess is `index` variable has not yet been defined in this function.

answerquest:
Update: yes, adding this line just before L4855 fixes the issue:
`var index = this.table.rowManager.getRowIndex(this);`

answerquest:
Created a pull request. Added the `var index...` line on top. Also made minified .min.js .

olifolkerd:
Hey @answerquest 

Thanks for that, i have added a note to the pull request, as the change will need to made on the source file too.

Cheers

Oli :)

olifolkerd:
Hey @answerquest 

I have pushed the update for this to the master branch, and will include it in the patch release later in the week.

Thanks for letting me know.

Cheers

Oli :)

PR #1048, opened 2018-04-15
declare index in delete function before using

answerquest:
fix for https://github.com/olifolkerd/tabulator/issues/1047


olifolkerd:
Hey @answerquest 

Thanks for the fix, 

If i am going to accept this request, you will need to you make the change in the `/src/js/row.js` file, in the delete function on line 497 and then run the ***gulp*** command to generate the tabulator.js and tabulator.min.js

Because both files are generated by gulp, i cannot accept pull request that only change those files, as they will be overritten next time the file is generated from source.

If you can make the change in the row.js file i would be happy to accept this pull request.

Cheers

Oli :)

olifolkerd:
Also as a side note,

it only needs to be `var index` to define the variable, rather than `var index =...`  could you leave out the equals part so that it dosnt get run if it isnt needed. it is just a bit more efficient when deleting large numbers of rows if it works that way.

Cheers

Oli :)

answerquest:
@olifolkerd  no need to accept this pull request then.. please do the change at your end itself, it's just a small thing. I want to save you trouble, not cause more!

PR #1057, opened 2018-04-17
New callback ajaxRequest

fortrax:
Hi Oli,

I promised you will hear from me again ;-)

My proposal here is to add a powerfull callback for the ajax call. You added a lot of auxiliary functions like paginationDataReceived to map web APIs that do not comply 100% to the requirements of the ajax call you generate. I found e.g. issue 164. You could solve this now elegantly with the new callback:
                    `var options = {`
`                        ajaxConfig: {`
`                            type: "POST",`
`                            contentType: 'application/json'`
`                        },`
`                        ajaxRequest: (config) => {`
`                            config.data = JSON.stringify(config.data);`
`                            return $.ajax(config);`
`                        }`
`                    };`
For issue 165, you found a work-around but also there, my function could have solved the problem elegantly. But more fundamentally, it allows to call any existing web interface. If you cannot change the server code, you need the full power of the callback here. What do you say?

Cheers,
Andreas


olifolkerd:
Hey @fortrax 

Nice to hear from you again, glad to see Tabulator is keeping you on your toes.

I can see where you are coming from there. and that is quite an interesting idea.

There are a couple of details that would need to be worked through first though. 

Im not sure exactly what your example is showing, as you can already alter all the config variables and parameters in the existing system so this extra function doesn't seem to be doing anything new. Could you provide a more specific usage case that you feel cant be achieved through the existing system, with an explanation of the benefits of this approach in that situation.

As i am looking to move away from jQuery dependency ion the 4.0 release in the next couple of months (there will still be a jQuery wrapper for those that want to stick with jQuery) i would be reluctant to add and functionality directly based off of the jQuery ajax functionality.

You also don't seem to be basing your code of the latest version of the Tabulator source as there are merge conflicts.

Let me know your thoughts.

Cheers

Oli :)

fortrax:
Hi Oli,

Yes you can change config variables, but in my example, the *container* of the variables is changed. If you want to do the Ajax call with JSon, there is no other way to get the JSON in the request, at least I could not find one.
But if you do not like the JQuery, we could replace it with any other http call. We cannot use my concrete scenario as example, it is too complex. Would you like me to make something up?

But the principle is clear, isn't it? You can use any communication protocol on any existing interface as long as you deliver the result in the form Tabulator expects.

Cheers,
Andreas

fortrax:
What about the following abstract example:
```
ajaxRequest: (config) => {
  var res1 = callWebServiceA(config);
  var res2 = callWebServiceB(config);
  return mergeResultsOfAandB(res1, res2);
}
```

olifolkerd:
Hey @fortrax 

Thanks for the example,

Unfortunately, that wouldn't work with your submitted pull request. the callback you have created is expecting a return of the ajax request object, and not the returned data itself, the ***done*** and ***fail*** functions are called on the return from this function.

Because calls to an external data source must be treated asynchronously it is not possible to use the callback in this way as it will likely return before the data is retrieved from an asynchronous source. *(hence it passing back the ajax object which then handles the async return and then calls the done or fail callbacks)*.

The correct approach to handle this would be for the callback to return a Promise which Tabulator would then be able to bind ***then*** and ***catch*** actions to.

At the moment Tabulator does not handle promises, and there is no suitable transpiling in place to make it backwards compatible for windows 10 devices so at this stage i would not accept a pull request on it.

As of version 4 i will be implementing a promise based architecture across Tabulator that will allow chaining of functions in this way  as well as on functions like setData and on the table constructor itself, which will give users much more flexibility in their code.

I think that your idea is great, and please don't take this as a criticism of your code as it is certainly useful in your usage case. But i have to take in consideration all users of the table and in its current form it isn't generally applicable enough to offer the functionality that you are proposing.

With a major update on the way i would prefer to wait till the promise functionality is in, and the present it to the users as a finished solution rather than offering them a new callback in this release, that will be completely changed in the next.

It would be great if you wanted to expand on this idea for the 4.0 release. and i would love to work with you more on it. If you are up for it then when i start work on the 4.o release in a months time i would love to collaborate with you on this part of the ajax system.

Cheers

Oli :)

fortrax:
Hey Oli,

No need for polite apologies. I use ES6 promises myself. (I don't need to support IE8, otherwise there is a polyfill on Github). 

How did you know mergeResultsOfAandB did not return a promise?  ;-) Another try:

```
ajaxRequest: (config) => {
  callWebServiceA(config).then(function(dataA) {
    callWebServiceB(config).then(function(dataB) {
      return mergeResultsOfAandB(dataA, dataB);
    });
  });
}
```

I see that it is a better timing to do it for 4.0 only. In the meantime, I will treat as a hotfix in my code.

Cheers,
Andreas

Jayu1234:
Hey Oli,
 I want to read only filter rows in html table.Please help me!! waiting for your reply

olifolkerd:
Hey @Jayu1234 

Your issue does not seem related to this pull request, please create a new issue to deal with your request.

When you create the issue, please give more information about what you mean by read only filter. the normal [Filtering Documentation](http://tabulator.info/docs/3.5#filtering) should guide you through what you need.

Many Thanks

Oli :)

fortrax:
Hi Oli, are you still prepared to do this? Cheers, Andreas

olifolkerd:
Hey Andreas,

Yes the promise based functionality will be comming in the version 4.0 release.

Cheers

Oli :)

fortrax:
Oh great! Just to make sure - promises **and** the proposed callback ("ajaxRequest")?

olifolkerd:
Hey @fortrax 

It will be a complete rebuild of the ajax system, until i have finished it i wont be able to tell you exactly how it will be made up.

keep an eye on the 4.0 documentation when it is released to find out more.

Cheers

Oli :)

fortrax:
Hi Oli,

I just had a look in 4.0. Am I right that the "ajaxPromiseFunc" in the options is what came out of this?

Cheers,
Andreas

olifolkerd:
Hey @fortrax 

That is correct. let me know your thoughts

Cheers

Oli:)

fortrax:
Hey Oli,

Today I tried ajaxPromiseFunc with pagination. Unfortunately, it's not being called then. With progressive load everything's fine.

Cheers,
Andreas

olifolkerd:
Hey @fortrax 

Could you post a copy of your constructor.

The Promise function is loaded when the table is initialized and replaces the inbuilt one, the pagination system uses the ajax module to make requests when it is in remote mode so there is no way for it not to use the function, so i am a bit confused,

Could you provide any more information about what you are experiencing?

Cheers

Oli :)

Issue #1145, opened 2018-05-22
rowVal check in 'like' custom filter


dpatel20:
Hi Oli - with regards to issue #1111, I have my filter working but as mentioned in the other issue I have had to make the following change your code:

> Also, I had to add the following check to the "like" filter as rowVal could be undefined or null:
>`if(typeof rowVal != 'undefined' && rowVal)`

Would you consider this a bug or an issue in my implementation or how I have set up my table?

Here's my function:

```JS
	function filterEverywhere(input, exact_match) {
		var cols = $("#admin-fixture-data").tabulator("getColumnDefinitions");

		var filterData = [];
		cols.forEach(function (col) {
			if (col.field) {
				if (col.sorter === "string" || col.sorter === "alphanum") {
					if (isNaN(input)) {
						if (exact_match)
							filterData.push({ field: col.field, type: "=", value: input });
						else
							filterData.push({ field: col.field, type: "like", value: input });
					}
				}
				else if ((col.sorter === "number"))
					filterData.push({ field: col.field, type: "=", value: Number(input) });
			}
		});

		$("#admin-fixture-data").tabulator("setFilter", [
			filterData
		]);
	}
```

olifolkerd:
Hey @dpatel20 

The like filter is currently only designed to handle strings so i would consider your update to the like filter to be an enhancement.

Would you like to submit a pull request to have the update included in Tabulator?

Cheers

Oli :)

dpatel20:
Done...please see #1148.

PR #1148, opened 2018-05-23
Check rowVal is valid in 'like' filter.

dpatel20:
Changes as per #1145.

I am submitting my first *ever* pull request (ever means on any project!! Go gently if I have done something wrong).

### DIFF
--- a/src/js/extensions/filter.js
+++ b/src/js/extensions/filter.js
@@ -559,8 +559,8 @@ Filter.prototype.filters ={
 		if(filterVal === null || typeof filterVal === "undefined"){
 			return rowVal === filterVal ? true : false;
 		}else{
-			if(typeof rowVal != 'undefined' && rowVal)
-				return rowVal.toLowerCase().indexOf(filterVal.toLowerCase()) > -1 ? true : false;
+			if(typeof rowVal !== 'undefined' && rowVal !== null)
+				return String(rowVal).toLowerCase().indexOf(filterVal.toLowerCase()) > -1 ? true : false;
 			else
 				return false;
 		}
--- a/src/js/extensions/filter.js
+++ b/src/js/extensions/filter.js
@@ -559,7 +559,10 @@ Filter.prototype.filters ={
 		if(filterVal === null || typeof filterVal === "undefined"){
 			return rowVal === filterVal ? true : false;
 		}else{
-			return rowVal.toLowerCase().indexOf(filterVal.toLowerCase()) > -1 ? true : false;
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @dpatel20 

Great to hear you have chosen Tabulator for your first pull requests!

I have a couple of bits of feedback for you on the changes, that will need updating before i can accept the request.

First of all you have used a "!=" in your code, could you please make sure it is a "!==" instead as it is more specific.

In the second comparison in that if just checks to see that rowVal is truthey, which will call it to fail if the value of the row is 0 or "" both of which are valid values.

what you could do instead is check that the value is not null ```rowVal !== null``` and then on the row below cast any other value to a string:
```js
String(rowVal).toLowerCase()
```
instead of
```js
rowVal.toLowerCase()
```

I hope that makes sense.

Let me know if you have any other questions.

Cheers

Oli :)

dpatel20:
Hi Oli -

I see my secret amateur programmer status is now out! ;)

I took my changes based on what I had seen online assuming that `!=` had to be used rather than `!==`. I seem to remember coming across an issue where I used `!==` and it did not work as expected (and see this [comment](https://stackoverflow.com/questions/5515310/is-there-a-standard-function-to-check-for-null-undefined-or-blank-variables-in/28475133#comment38025689_5515385)). And [here](https://stackoverflow.com/questions/5515310/is-there-a-standard-function-to-check-for-null-undefined-or-blank-variables-in/28475133) I had seen about checking if truthy using `if(someVar)`.

And then the following links that I have seen after checking again suggest otherwise:
https://stackoverflow.com/questions/3390396/how-to-check-for-undefined-in-javascript
https://stackoverflow.com/questions/2647867/how-to-determine-if-variable-is-undefined-or-null

So I am totally confused as to the best way (in 2018) to test for `null` or `undefined`. I am willing to bow to your vastly superior knowledge on this...

olifolkerd:
Hey @dpatel20 

Great stuff, that is looking much better! The final change i would ask for is that you use curly braces in your if/else statement to match the coding style of the rest of Tabulator.

Once that is in i will be happy to merge in the pull request.

Cheers

Oli :)

dpatel20:
Take 3 - bet you wish you'd just done it yourself! :)

Thanks for the patience and tutorial.

olifolkerd:
Approved!!!

Great work, thanks!

Pull requests are very rarely simple, you will always have a bit and back and forth with the people who maintain the library.

Keep up the good work, let me know if you have any other suggestions.

Cheers

Oli :)

PR #1170, opened 2018-05-31
Make row's getData method returns a new object

lgraziani2712:
This PR fixs the problem of returning the reference of the row's internal data object.

With this, the user code doesn't affect the row's data and vice versa.

### DIFF
--- a/src/js/row.js
+++ b/src/js/row.js
@@ -426,7 +426,7 @@ Row.prototype.getData = function(transform){
 			return self.table.extensions.accessor.transformRow(self.data, transform);
 		}
 	}else{
-		return this.data;
+		return Object.assign({}, this.data);
 	}
 
 };


olifolkerd:
Hey @lgraziani2712 

Thanks for the pull request

Are you finding this to be an issue?  please describe a usage case for this update?

Cheers

Oli

lgraziani2712:
Sure!

I cache the data while is being edited so I can send to the server only what I add/change/delete. This cache listens the Tabulator event, so everytime I edit a cell, I get the row data, then I process it and finally I store it in the cache.

I process the data because some custom editors saves an object or an array.

Once I click the persist button, I send that cached data to the server. The thing is, my data object is reverted to the original state because is a reference of the same object manipulated by Tabulator.

If something isn't clear, ask me! English isn't my first language.

Thank you, Luciano.

lgraziani2712:
BTW, what I'm doing right now to solve it is this

```js
const rawData = Object.assign({}, row.getData());

//process rawData

return cleanData;
```

Issue #1186, opened 2018-06-08
Cell edition crashes when pressing enter if validation fails and if html instance is wrapped in a form tag


lgraziani2712:
Updated description:

### Scenario

1. Tabulator html instance is wrapped with a `<form>` tag.
2. Edit a cell.
3. Add an invalid value.
4. Press enter.
5. Validation fails (but nothing is printed).
6. Edition stop working.

This doesn't happens with the `tab` key or a click.

### Versions

- jQuery: v3.3.1
- jQueryUI: v1.12.1
- Tabulator: v3.5.2

### Demo

https://codepen.io/lgraziani/pen/dKvVyE?editors=0010

Thank you! Luciano.

olifolkerd:
Hey @lgraziani2712 

Could you post a copy of your table constructor, or build a JS fiddle to show how your table is setup.

Which version of Tabulator are you using?

Cheers

Oli :)

lgraziani2712:
Sure! Here is the configuration: https://codepen.io/lgraziani/pen/dKvVyE?editors=0010

~~I tested it there and everything works. I still don't know why it doesn't work~~

I found the error!!! It happens when tabulator is inside a form tag. The pen is updated with the bug

lgraziani2712:
I tried to fix this by myself, but I failed :(

@olifolkerd, if you can tell me which files/functions I could check, I would try to fix this in no time heh :)

PR #1206, opened 2018-06-18
Fix firefox bug with buttons

lgraziani2712:
Following this bug #1186, when Tabulator is wrapped in a form tag, there is another bug in firefox.

When you click a navigation button something crashes without logging anything, and breakes the page. This is because the buttons aren't being configured as simple buttons and instead they trigger the submit event.

This PR replace the `role='button'` attribute with `type='button'`.

### DIFF
--- a/src/js/extensions/page.js
+++ b/src/js/extensions/page.js
@@ -219,7 +219,7 @@ Page.prototype._setPageButtons = function(){
 
 Page.prototype._generatePageButton = function(page){
 	var self = this;
-	var button = $("<button class='tabulator-page " + (page == self.page ? "active" : "") + "' data-page='" + page + "' role='button' type='button' arpagea-label='Show Page " + page + "'>" + page + "</button>");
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @lgraziani2712 

Thanks for the pull request, that is a great solutions, though to accept the pull request i will need to you leave the `role='button'` bit in as well as they are used for screen readers as part of accessibility.

If you can do that i will accept this request.

Cheers

Oli :)

lgraziani2712:
image

Wow didn't saw this until now. That's why that buttons worked but those of the specific pages didn't!

I would fix that!

lgraziani2712:
Done! @olifolkerd

olifolkerd:
While you are at it, would you mid fixing another tag on that element, it looks like one of the aria tags has been corrupted at some point.

the `arpagea-label` tag should be `aria-label`

Cheers

Oli :)

lgraziani2712:
Done @olifolkerd !

olifolkerd:
Merged! thanks for you contribution and very rapid response!

olifolkerd:
it will be included in the patch release at the weekend

Issue #1223, opened 2018-06-26
Regex validator with ':'


mateddy:
I ran into an issue related to regex.  My column looks like this:

{ field: 'time', title: 'Time', editor: 'input', validator:'regex:^([01][0-9]|2[0-3]):[0-5][0-9]$' },

It seems the hour part /([01][0-9]|2[0-3])/ and min part /[0-5][0-9]/ works on its own, but the colon itself messed things up.

mateddy:
Proposing fix in #1228

olifolkerd:
Approved!

Sharkimus:
Hi, I know this is closed, but I'm having the same problem, slightly different situation.

I had the regex working when Tabulator was instanced in the document ready function, but when I moved it into an onclick event for a bootstrap 4 modal to make a table in the modal editable the validator kept failing in the same way as described in the initial post.

Upon calling the validationFailed callback it shows the regex is only being applied up to the colon character, you may be splitting the string on each colon character instead of just the first somewhere? or something about how bootstrap manages the modal is breaking it possibly?

I've gotten around this by creating a custom validator to do the regex but thought I'd let you know the problem my be still there.

If anyone else is struggling with this:
`		// Custom validator for tabulator
		var timeValidator = function(cell, value, parameters){
			var r = /^([0-9]|0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]$/
			return r.test(value)
		};
`

wdehoog:
This bug is probably caused by this line:

```
        parts = value.split(":", 2);`
```

which throws away the part of the regexp after the ':'. Better would be 

```
        pos = value.indexOf(':')
        type = value.substring(0,pos)
        params = value.substring(pos+1)

```
Should this bug be reopened or require a new one?

olifolkerd:
@wdehoog If it is the result of a different problem it should be a new issue, but if it is a small fix and you think you have a solution a Pull Request with the suggested fix in would be better.

Cheers

Oli :)

wdehoog:
@olifolkerd I'll try to explain.

I have this in my table: `validator: ["required", "regex:(0|1|2)?\\d:\\d\\d"]`
split results in three parts: 'regex', '(0|1|2)?\\d' and  '\\d\\d'
split(":", 2) would limit the result to two parts: 'regex' and '(0|1|2)?\\d'
resulting in type = 'regex' and param = '(0|1|2)?\\d'

[...]

PR #1228, opened 2018-06-27
Fix for Regex validator with ':' #1223

mateddy:
Fix for #1223 
It seems that multiple validators has to be passed as array, so if the input is string, it contains only one validator.
split(':', 1) only break the string into two parts.


mateddy:
Fix for #1223

olifolkerd:
good solution.

Sorry for the delay in approval. i will include it in the 4.0 release in the next couple of weeks

Cheers

Oli

PR #1292, opened 2018-07-23
Finding rows and columns by jQuery doesn't work

lavigor:
Everytime you call `jQuery` a new object is created.

https://stackoverflow.com/a/6985990
> Using the `$()` function will always create a new object, so no matter what, your equality check there will always fail.

The correct way is to use the [`.is()`](http://api.jquery.com/is/) jQuery method.


olifolkerd:
Hey @lavigor,

Sorry for the delay in getting back to you.

That is a very good catch, thanks for submitting!

Cheers

Oli :)

PR #1296, opened 2018-07-24
Fix textarea editor: renders null values

lgraziani2712:
### Bug

It renders null values into textarea cells.

How to replicate it: https://codepen.io/lgraziani/pen/yqbQgY?editors=0010

Click on the cell to see how it renders the null value.

This PR fix the issue: https://codepen.io/lgraziani/pen/JBNerO?editors=0010

### Why `value == null`?

http://adripofjavascript.com/blog/drips/equals-equals-null-in-javascript.html

> Despite the fact that null is a falsy value (i.e. it evaluates to false if coerced to a boolean), it isn't considered loosely equal to any of the other falsy values in JavaScript. In fact, the only values that null is loosely equal to are undefined and itself.
>
> Because of this interesting property, it has become somewhat conventional to use == null as a more concise way of checking whether a given value is "nothing" (null/undefined) or "something" (anything else).

Thank you for your work!

### DIFF
--- a/src/js/extensions/edit.js
+++ b/src/js/extensions/edit.js
@@ -296,7 +296,7 @@ Edit.prototype.editors = {
 	textarea:function(cell, onRendered, success, cancel, editorParams){
 		var self = this,
 		cellValue = cell.getValue(),
-		value = String(cellValue == null ? "" : cellValue),
+		value = String(typeof cellValue == "undefined" || cellValue === null ? "" : cellValue),
 		count = (value.match(/(?:\r\n|\r|\n)/g) || []).length + 1,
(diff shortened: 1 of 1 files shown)


lgraziani2712:
Hmm, it won't be necessary to use the undefined handling.

```js
console.log(null == null); // true
console.log(undefined == null); // true
```

The `or` handler wont be evaluated if `cellValue` contains `null` or `undefined`.

olifolkerd:
@lgraziani2712 Undefined checks are used throughout tabulator and to fit with that coding style it must be included.

If you can restore that back to the statement i would be happy to accept the pull request.

Cheers

Oli :)

PR #1426, opened 2018-09-20
Update select_row.js

erez86:
Fix multiple/single row selection to be compatible with web / common tables conventions.

### DIFF
--- a/src/js/extensions/select_row.js
+++ b/src/js/extensions/select_row.js
@@ -39,7 +39,7 @@ SelectRow.prototype.initializeRow = function(row){
 		if(self.table.options.selectable && self.table.options.selectable != "highlight"){
 			element.on("click", function(e){
 			if (e.shiftKey) {
-                            var rows = self.table.getRows();
+                            var rows = self.table.getRows().slice(0);
                             if(self.focusedRow===undefined) {
                                 self.focusedRow = row;
                             }
--- a/src/js/extensions/select_row.js
+++ b/src/js/extensions/select_row.js
@@ -1,12 +1,12 @@
 var SelectRow = function(table){
 	this.table = table; //hold Tabulator object
-	this.selecting = false; //flag selecting in progress
+	this.focusedRow = undefined; //last clicked row
 	this.selectPrev = []; //hold previously selected element for drag drop selection
 	this.selectedRows = []; //hold selected rows
 };
 
 SelectRow.prototype.clearSelectionData = function(silent){
-	this.selecting = false;
+	this.focusedRow = undefined;
 	this.selectPrev = [];
 	this.selectedRows = [];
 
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @mikila85  an @erez86 

Thanks for your pull request,

While i do appreciate the functionality you are after (and it is great that you have sumbitted a pull request with your ideas),

Im afraid your suggestion completely replaces the current selection method, which while i am sure it is great for your usage case it goes completely against what other users have requested.

The drag selection mechanic was specifically requested a number of times when this section was being developed.

I would be happy to look at adding a option to choose between a between drag select and shift range select to allow both approaches, but i would not be looking to remove the drag select from the code.

Cheers

Oli :)

mikila85:
I will fix it to allow both approaches with old one by default and resubmit, are there any other comments on the code that I should fix while I'm at it?

olifolkerd:
Hey @erez86,

Great to hear your are still interested in submitting a pull request.

A few pointers for your next submission:

Please dont assign ***undefined*** to variables, it is bad practice and isnt used elsewhere in Tabulator, please use ***false*** instead.

Please do not completely rewrite the structure of the module, lots of other things depend on this and so it could have unintended consequences. the existing event infrastructure is fine as it is, you only need to add some if statements to the ***mousedown*** event  and ***endSelect*** function to chose between the different modes of operation.

Please only add the update for the range selection on shift, the rest is already handled.

Could you call the new option ***selectableRangeType*** with options of ***drag*** as the default and ***click*** for your option.

When using comparison operators please put spaces between the variables and the operators for clarity . it should be `x ? y : z` not `x?y:z`

you don't want to be calling the ***getRows*** function on the table, that is designed for use external to the table, you want to be calling the ***self.table.rowManager.getDisplayRows()*** function which returns the array of internal Row objects that Tabulator uses, not the Row Component this is for external use only.

you can find the index of the display row (bare in mind things are a lot more complicated internally than just looking things up using the external API) using the `self.table.rowmanager.getDisplayRowIndex(row)` passing in the row are looking for, rather than calling filter and findIndex functions.

You seem to be going to a lot of effort to repeatedly splice and filter rows, on large tables this is going to consume a lot of memory and computation time, it would be far easier to find the index's you need from the display rows using the above two functions and then itterate between the start and end values with a for loop using the existing array. it would be much more efficient that way.

Can you make your updates on the 4.0 branch rather than master as there have been some major updates and the structure of a lot of the core files has changed considerably since the master brnach was last changed so i am not accepting changes on it till after the 4.0 merger next weekend.

Cheers

Oli :)

erez86:
Hey @olifolkerd,
The new pull request is [https://github.com/olifolkerd/tabulator/pull/1448](https://github.com/olifolkerd/tabulator/pull/1448)

PR #1512, opened 2018-10-16
Downloading grouped tables and group clearing

resonatingfury:
I needed some additional grouping functionality for my usage of Tabulator, and though I'd share them:

-added 'fileGroupheader' and 'downloadGroupedData' to options, allowing a custom header function to be passed in for downloaded files(to use different syntax for a report, or use HTML for the actual table, etc.) and a toggle for whether or not the data is to be downloaded grouped (enabled by default and exists merely for freedom)

-added the 'getGroupedData' in the data manipulation section, which returns a data object of the same format that 'getData' produces, but with group headers as part of the object. they are tied to the first column field, so they are placed correctly

-edited 'download.processData' such that it handles the grouping toggle and passes the correct data through

-added a 'clearGrouping' function, which cleanses the group managers of grouping data. if an empty string is passed into 'setGroupBy', this function is called automatically, allowing for grouping switches(like what I needed for my implementation).

I only edited tabulator and tabulator-min for now, and have not tested it with nested groups as I don't use them, but I think nested groups should be fine.


olifolkerd:
Hey @resonatingfury 

Thanks for sharing your updates to tabulator.

I think that is a really interesting approach you have taken. As it happens im in the process of completely rebuilding the download module to make it handle row grouping, column grouping and column calculations, so i wouldn't be looking to pull any changes in on that side of things at the moment.

I really like the idea of your ***getGroupedData*** function and would be happy to accept a pull request on that idea, but im afraid i would need there to be a few changes to the code before i could accept it:

You have made your changes on files in the `/dist` folder, these files are automatically generated by gulp from the files in the `/src` folder and would be overwritten the moment an changes are made to the source code.

When making additions to Tabulators API the correct methodology would be to define the function in the '/src/js/core.js` file and then put the module specific code in the correct module, in this case the '/src/js/modules/group_rows.js` file.

You could then use the internal structure of the groups module, to iterate through the groups to build the output data, this would also allow you to automatically handle nested data. Internal Tabulator functionality as a rule does not call any of the public API functions, keeping the internal functionality isolated from the public API.

If you would be interested in adding this function let me know and i can guide you through some of the inner workings of Tabulator.

Cheers

Oli :)

resonatingfury:
Hi @olifolkerd 

Thank you for politely correcting me- I have no prior experience with github, or collaborating on existing code! What you said makes sense in hindsight. I did everything in a day and rushed through a bit.

I tried to take your points and properly update the correct files. Please let me know if it is still not correct!

olifolkerd:
Hey @resonatingfury 

Great to hear you have chosen Tabulator as your first chance to collaborate!

You will need to raise a new pull request with your updated commits in as this one still contains all of your previous code on in the dist files.

I just have a couple of other pointers for you, just to fit in with the coding style of the rest of Tabulator:

When you are assigning properties to objects, could you use dot notation instead of array notation:
```js
groupHeader["level"]
```
should be 
```js
groupHeader.level;
```

Where you are trying to get the row data, you are just using the data property of the row. there are a load of transforms that need to happen to data before it leaves the table. to make sure you get the right data you should call ***getData***, so:
```js
row.data
```
should be
```js
row.getData(true);
```

This ensures that any accessors etc defined by the user are applied to the data before it leaves the table. and isolates the objects to prevents accidental pass by reference issues.

Could you define your ***pullGroupListData*** function on the prototype instead of as a variable in a function, on the whole functions should not be declared inside other functions in Tabulator as it prevents their reuse. The only exception to this is where that function is used as a closure or is assigned elsewhere and its scope isolation is necessary, but in this case that isnt so.

The logic structure of your if else is a bit strange, you seem to have some of the code that can only be executed if group.hasSubGroups is true inside the if statement and the other half after the else. if it can only be executed inside of the if could you place it in the if.

For the time being could you avoid using spread operators in the code, I completely agree with their use in the long term, but at the moment Tabulator has not made the move to using most of the ES6 syntax so i just want to keep things consistent till it is introduced.

Aside from those small tweaks, it is looking good!

Cheers

Oli :)

PR #1523, opened 2018-10-22
Fix table bg color in Bootstrap theme

polomasta:
While using the bootstrap theme I found that when trying to set a custom `$backgroundColor` variable for the table it was being overridden by something.

On investigating further I found @ line 331:
```
.tabulator-table{
   position:relative;
   display:inline-block;
   background-color:$rowBackgroundColor;
   white-space: nowrap;
   overflow:visible;
```

Here we can see that the `.tabulator-table` wrapper has its background-color being set to the `$rowBackgroundColor` variable... which then overrides  `.tabulator` background color being set on line 43 with the `$backgroundColor` variable.

Unless I'm mistaken, line 334 isn't needed, hence this change to remove it and allow the table background color to actually be set by the `$backgroundColor` variable.

Thanks!

### DIFF
--- a/src/scss/bootstrap/tabulator_bootstrap4.scss
+++ b/src/scss/bootstrap/tabulator_bootstrap4.scss
@@ -331,7 +331,6 @@ $footerSeperatorColor:$table-border-color !default; //footer bottom seperator co
 		.tabulator-table{
 			position:relative;
 			display:inline-block;
-			background-color:$rowBackgroundColor;
 			white-space: nowrap;
 			overflow:visible;
 


olifolkerd:
Hey @polomasta 

Thanks for getting in touch.

The ***rowBackgroundColor*** variable is used throughout all of the themes, i keep the variables the same across all of them to make it easier to maintain all the themes as the codebase gets adjusted.

You will see at line 22 of the .scss file that the ***rowBackgroundColor*** is actually set to be the ***backgroundColor*** variable.

The ***backgroundColor*** variable is actually for the background of the table itself (which is only visible when there are not enough rows or columns to fill the table entirely.

The ***rowBackgroundColor*** then sets the background color of the rows themselfs.

I hope that clears things up.

Cheers

Oli :)

polomasta:
@olifolkerd 

> (which is only visible when there are not enough rows or columns to fill the table entirely)

This is not correct in my case, where we've modified the design of our table to have margin between the rows for aesthetic purposes.

In this case, `rowBackgroundColor` causes and undesirable override of `backgroundColor`

*example screenshot*
gig wage - easy payments for 1099 independent contractors 1099 payroll 2018-10-25 15-26-37


What would be your suggested method for setting a table background color different from the row background color?

olifolkerd:
Hey @polomasta 

Thanks for clarifying your usage case,

I wouldn't recommend using margins on rows, as tabulator does not expect there to be margins on rows, which could throw out the virtual DOM when it comes to adjusting for scroll position or updating table data.

Instead i would suggest making the rows bottom border thicker and using the ***rowBorderColor*** variable to set this to what you want it to be.

If at that point you still want a "border" around the row you could set that on the bottom border of the cells.

I hope that helps,

Cheers

Oli :)

polomasta:
hmm that seems a very odd way to go about things... but ok.

olifolkerd:
Feel free to include the margins if you want, im just warning you that Tabulator does not expect them and it will likely cause rendering glitches for you as it will throw out the maths of the virtual DOM.

Another option would be to add padding to your rows and then use the borders round the cells to define the row edge. there are any number of options to achieve what you are after.

You can of course override any CSS you like outside the theme.

As your are manipulating the style beyond the scope of the variables in this theme i would not be looking to pull this one change in.

Cheers

Oli :)

PR #1534, opened 2018-10-26
Fix deselect "row" when do ".deleteRow()"

onikienko:
Looks like `this` is `row component` itself. Fixes deselecting when you are removing selected *row*.


olifolkerd:
Hey @onikienko 

Good spot,

Thanks for the fix, i will include it in the next patch release.

Cheers

Oli :)

PR #1545, opened 2018-10-30
clearHeaderFilter fix for checkboxes

resonatingfury:
when clearing filters, even with `clearFilter(true)`, the filter would clear from the data but the checkbox in the header would stay checked. setting the checked value to `false` did not work either; it would de-register the input and force me to check, uncheck and recheck it just to use the filter again. `click()` handler works just fine.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -1349,6 +1349,13 @@ Tabulator.prototype.getGroups = function(values){
 	}
 };
 
+// get grouped table data in the same format as getData()
+Tabulator.prototype.getGroupedData = function(){
+	if (this.modExists("groupRows", true)){
+		return this.options.groupBy ? 
+				this.modules.groupRows.getGroupedData() : this.getData()
+	}
+}
 
 ///////////////// Column Calculation Functions ///////////////
 Tabulator.prototype.getCalcResults = function(){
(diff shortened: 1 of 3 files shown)


olifolkerd:
Hey @resonatingfury 

Thanks for getting in touch.

While agree with your premise im afraid the solution you suggest does not take advantage of some of the latest features offered since the 4.0 release. 

Sadly it looks like the old clearHeaderFilter function got missed out in the update so it didnt give you a great starting point

Because we have no way of knowing how editors are setup it would be bad to try and interact directly with the elements they contain, by setting values or clicking *(in 4.1 there is a tristate checkbox which will make it even more complicated)*.

Instead you want to have a look at the ***setHeaderFilterValue*** function in the filter which sets a value then completely regenerates the header element to avoid any undesirable interaction issues.

I would suggest that actually what you want to be doing in the ***clearHeaderFilter*** function is calling ***setHeaderFilterValue*** passing in a value of ***null***

If you would be willing to approach it that way i would be happy to consider the pull request.

Also you appear to have accidentally submitted to other commits in your pull request that are unrelated to the issue, please resubmit without these extra commits

Cheers

Oli :)

olifolkerd:
Hey @resonatingfury 

Due to the release of v4.1 tomorrow i have made the changes myself.

Please feel free to make pull requests in future with any ideas you may have.

Cheers

Oli :)

PR #1549, opened 2018-10-31
Data loss occurs when calling getData if column's datatype is date

SpecmoComptonJames:
If a date data type is bound to a column, the data is lost when calling getData while inside the deepClone() function.  Dates evaluate as an object causing the function to recursively call itself.  When this happens the Date value is lost.  In this pull I add an additional test to deepClone to allow Dates to copy like all the other non-object types, else it works as previous versions for 'objects'.  I am unsure if other date types may be affected and may need to be examined.  However, this pull addresses issues my application is encountering with DateTimes getting mangled when loaded into tabulator, modified, and then saved after upgrading to version 4x.

Also while building my fix, I found an issue with filename case sensitivity which was causing the gulp script to fail on linux.

Thanks for a great library.


olifolkerd:
Hey @SpecmoComptonJames 

Thanks for getting in touch, 

Thanks for pointing out the issue, i would be happy to accept your pull request, but there is a tweak i would need you to make.

The point of the deep clone is to isolate the data inside the table from outside the table and the current method of handling this does not do that, someone altering that date object would also alter it in the actual row data..

To avoid that you can use something like `copiedDate = new Date(date.getTime());`

If you could take that approach that would be great!

also could you please remove the verbose comments, i appreciate that you put them in to explain what your changes were but the coding style of Tabulator does nor really go in for that, a single short line above the inserted line of code would suffice.

Good spot on the kebindings case there!

Cheers :)

SpecmoComptonJames:
Thanks Oli... Glad I could help out...  tabulator has been a major joy to work with and happy to help it grow and improve any way I can.

Removed the comments and changed just the new date case to wrap as 'new'.... I think this meets your request, if not let me know and I'll tweak again.   Your request makes total sense once I thought about it.

olifolkerd:
Hey @SpecmoComptonJames 

It is great to hear you are enjoying working with Tabulator :D 

Thanks for making those tweaks, that looks great.

I will pull this in as part of the 4.1 release this weekend

Cheers

Oli :)

olifolkerd:
all merged in!

Thanks for your contribution :D

PR #1581, opened 2018-11-08
Fixing 'this' context issue for responsive module

jjwilly16:
Adding a self variable to avoid using 'this' inside of forEach function


olifolkerd:
Good Spot!

PR #1611, opened 2018-11-29
setPage missing return for promise

erez86:


erez86:
Hi Oli,
[Documentation shows that it should return promise:](http://tabulator.info/docs/4.1/page#manage)

table.setPage(1)
.then(function(){
    //run code after table has been successfuly updated
})
.catch(function(error){
    //handle error loading data
});

PR #1622, opened 2018-12-04
fixed wrong element variable assingment

dwolff:
Hi,
I've fixed a small typo on the "element" variable assignment, inside the image formatting parameters.


olifolkerd:
good catch, i will merge this in with other patch update for this weekends release

PR #1625, opened 2018-12-06
fix cssClass can not have multi classes using space

puffsnow:
Hi Oli,
As document said, cssClass value should be a string containing space separated class names, but in current there will be an error said 「Uncaught DOMException: Failed to execute 'add' on 'DOMTokenList': The token provided ('aaa bbb') contains HTML space characters, which are not valid in tokens.」.
It looks like you can not put the string with space to the list directly, it should be separated before add to classList.
I try to fix this problem and please check if it is appropriate, thanks!

### DIFF
--- a/src/js/cell.js
+++ b/src/js/cell.js
@@ -123,7 +123,10 @@ Cell.prototype._configureCell = function(){
 	}
 
 	if(self.column.definition.cssClass){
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @puffsnow 

Thanks for the submission, that looks great!

Cheers

Oli :)

PR #1658, opened 2018-12-17
1618: On selectRow/deselectRow, call rowSelectionChanged once

gillez:
- Bug fix: Any calls to selectRow and deselectRow should call
rowSelectionChanged only once.
- This was fixed in SHA 010b37eaec for cases when an array is passed.
However, the problem also exists when passing true or no parameter,
which should select active rows or all rows.
- Also, calls to deselectRows had the same issue so this is
fixed as well.

### DIFF
--- a/src/js/modules/select_row.js
+++ b/src/js/modules/select_row.js
@@ -139,7 +139,7 @@ SelectRow.prototype.selectRows = function(rows){
 	switch(typeof rows){
 		case "undefined":
 		self.table.rowManager.rows.forEach(function(row){
-			self._selectRow(row, false, true);
+			self._selectRow(row, true, true);
 		});
 
 		self._rowSelectionChanged();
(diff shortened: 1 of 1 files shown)


olifolkerd:
Thanks for the submission @gillez, i will review this at the weekend.

Cheers

Oli :)

olifolkerd:
All merged in, i will include it in this evenings pull request.

Thanks for your contribution

Cheers

Oli :)

dcanning:
I just ran into this issue trying to select 10k+ rows, and lo and behold someone already fixed it, thanks so much @gillez

dcanning:
I just ran into this issue trying to select 10k+ rows, and lo and behold someone already fixed it, thanks so much @gill
@olifolkerd it looks like this is fixed in the source for 3.5, but isn't pushed to npm, any chance of that happening?

PR #1685, opened 2018-12-27
Properly dispose of old header filter elements

ddetkovs:
---
Old header filter elements are not removed from DOM
---

**Describe the bug**
Old filter elements are not removed from column content element. As a result, if you use setHeaderFilterValue or other methods that use generateHeaderFilterElement it creates this kind of DOM

image

If header filters have some styling it becomes more apparent, since this styling applies to old filter elements as well.

**To Reproduce**
1. call setHeaderFilterValue method on the table with some new value
2. look up the filter element in DOM or add this CSS 
```
.tabulator-header-filter {
    background-color: red;
    padding: 5px;
}
```

**Expected behavior**
old filter elements to be removed from DOM

**Screenshots**
image

### DIFF
--- a/src/js/modules/filter.js
+++ b/src/js/modules/filter.js
@@ -106,7 +106,18 @@ Filter.prototype.generateHeaderFilterElement = function(column, initialValue){
 	function cancel(){}
 
 	if(column.modules.filter.headerElement && column.modules.filter.headerElement.parentNode){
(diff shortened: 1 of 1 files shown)


olifolkerd:
All merged in, i will include it in this evenings patch release.

Thanks for your contribution

Cheers

Oli :)

PR #1686, opened 2018-12-27
Allow persistence ID to be changed

gillez:
Allow a new persistence ID to be set, allowing different column layouts for the same table.

- New function in persistence module: setID
- Tabulator.setColumns now reloads persistent column
setup before setting columns.
- Also save persistent columns on header double-click resize.

My use case for this is that the columns in my table can change based on certain external events (new setup docs are loaded from a database).  When this happens, I reload the columns in my table using setColumns.  However, as the columns are now different from the previous persistent set that have been saved for this table, the layout gets reset to the default.  To overcome this, I change the ID when the number of columns change, saving a new table layout for each set of columns.

This is not possible with the master Tabulator code as the persistent ID is set on initialisation and the columns are only loaded once, at that point.

When the columns change I now call the following:

```
          if (table.modExists("persistence")) {
            table.modules.persistence.setID(tableId + newTabColumns.length);
          }  
          table.setColumns(newTabColumns);
```

I think that this may be a useful addition for others, hence the merge request.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -991,7 +991,11 @@ Tabulator.prototype.copyToClipboard = function(selector, selectorParams, formatt
 /////////////// Column Functions  ///////////////
 
 Tabulator.prototype.setColumns = function(definition){
-	this.columnManager.setColumns(definition);
+  	if(this.options.persistentLayout && this.modExists("persistence", true)){
+    	definition = this.modules.persistence.load("columns", definition);
+  	}
+
+  	this.columnManager.setColumns(definition);
 };
 
 Tabulator.prototype.getColumns = function(structured){
--- a/src/js/modules/persistence.js
+++ b/src/js/modules/persistence.js
@@ -14,6 +14,10 @@ Persistence.prototype.initialize = function(mode, id){
 	this.id = "tabulator-" + (id || (this.table.element.getAttribute("id") || ""));
 };
 
+Persistence.prototype.setID = function(id) {
+	this.id = "tabulator-" + id;
+}
+
 //load saved definitions
 Persistence.prototype.load = function(type, current){
 
--- a/src/js/modules/resize_columns.js
+++ b/src/js/modules/resize_columns.js
@@ -44,6 +44,9 @@ ResizeColumns.prototype.initializeColumn = function(type, column, element){
 		handle.addEventListener("dblclick", function(e){
 			if(self._checkResizability(column)){
 				column.reinitializeWidth(true);
+				if(self.table.options.persistentLayout && self.table.modExists("persistence", true)){
+					self.table.modules.persistence.save("columns");
+				}
 			}
 		});
 


gillez:
Hi,
Are you interested in this change at all?  If so, I'll update the merge request to resolve the conflict that has appeared with the latest master.  If not I'll close the merge request.

olifolkerd:
Hey gillez,

Sorry for not getting back to you on this, i will review it at the weekend.

Cheers

Oli

olifolkerd:
Hey @gillez 

Sorry for the late reply,

Have a look at the updated persistence module in 4.5 and see if it meets your needs.

Im unsure as to the purpose of this, why would you need to update the id on the fly? if you change the persistence id then when the table next loads it will still load the old persistence data if you havnt changed the constructor object.

also you dont expose the set ID function in the core.js so it isnt actually available for anyone to use?

Cheers

Oli :)

olifolkerd:
Closed due to lack of activity.

@gillez if you want to pickup discussion on this again i would be happy to reopen.

Cheers

Oli :)

gillez:
Hi Oli,

Thanks for the reply back in November.  I missed the comment at the time. 

My reason for doing this was because I change the number of columns in the table after the table is created and I didn't see any other way to get what I wanted.  However, this was a while ago and, off the top of my head, I don't remember the details.  I will look into it this week and write a better explanation.  We can then decide whether to reopen this.  It may be that there's a better way to do what I require.

Thanks,
Giles

gillez:
Hi Oli,

I haven't yet checked out Tabulator 4.5 but wanted to get back to you this week.  This is based on my current version of Tabulator, which is 4.3  I'll try Tabulator 4.5 with the updated persistence module asap, but it will probably be next week now.

"I'm unsure as to the purpose of this, why would you need to update the id on the fly?"

My situation is as follows:

1. my page loads and sets up a table with, say, 5 columns
2. Persistence is on so the current setup for this table is loaded
3. Sometime after this (when an async load from a database completes) another column is added to the table.  Let's call this col6.  
4. This column gets added to the end of the column array and I update the columns in the table.  The new column layout (with col6 at position 6) is saved.
5. The user resizes col6, and maybe moves it to position 3, using the fanatastic column drag-n-drop features of Tabulator (which I love).
6. This new column layout is automatically saved by the persistence module.

7. The page is now refreshed.
8. As step 1 above - the page loads and table is set up with 5 columns
9. As step 2 - The current setup for the table is read from Local Storage (via persistence module).
10. One column in the persistence setup (col6) does not exist in the table and so this column is removed from the state. The five columns that currently exist use the setup.
11. I think at this point the state is automatically saved, now with five columns.
12. As step 3 - col6 is loaded from the database.
13. As step 4 - col6 is added to the columns array and the columns are updated in the table.
14. As the persistent state of the table does not contain col6, col6 is added to the end with it's default size.

This is the problem! As col6 is not in the original table columns when the page loads, it's position and size cannot be customised (or at least they can, but the saved state will be removed when the table is loaded without the column).

To overcome this with the change in this merge request, I have a different saved state for this table with 5 columns and 6 columns, which I do by persistent ID.  At step 2, persistent ID is set to "table5" and so the table layout with 5 columns is loaded and saved.  At step 4 I change the perisistent ID to "table6" and so the table layout with 6 columns is loaded and can be modified  and saved.  

With the standard Tabulator code, I think the only way to do this would be to destroy the table and create a new one.

I hope this explanation makes sense.  I realise it's quite long but I thought I should explain exactly what is going on and my reason for this as my short summary before obviously wasn't sufficient. :-)

An alternative would be to keep the state for columns that don't currently exist when loading the peristent state for the table, in case those columns are added in future.  This would avoid storing multiple states for the same table.

"also you dont expose the set ID function in the core.js so it isnt actually available for anyone to use?"

In my code I do the following:

```
          if (table.modExists("persistence")) {
            table.modules.persistence.setID($ctrl.id + newTabColumns.length);
          }
```

So use the function from the module directly.  I appreciate this is not the way to do it and it should be exposed in code.js.

Thanks,
Giles

gillez:
Hi Oli, 

I know you get inundated and it may be that you're well aware of my issue and will respond in good time, but just want to check that you read my last comment/explanation on why I need this feature.  As this PR is closed, I'm a bit concerned that you may have missed it and now won't go back to it.

It may be that there's another way to fix the above situation, or you may have an idea of how to do it in a different way, in which case I'm happy to keep it closed.  However, in the absence of an alternative, this PR does solve my problem and seems a relatively small change with no negative side-effects.

Thanks,
Giles

gillez:
Hi Oli,

As this PR is closed, I just want to check that you read my previous comment. You said in November that if I want to pickup discussion you would be happy to reopen. Please could you comment and let me know if there is another way to fix my situation described above or whetehr it is worth reopening this PR.

Thanks,
Giles

olifolkerd:
Hey @gillez 

Sorry for the delay in getting back to you, as you can imagine things have been a bit hectic of late.

That certainly is a complex arrangement you have there. As it stands i would consider the table to be persisting correctly in those circumstances, it is only intended to persist the current active state of the table.

Asking any more of it than that an i fear things would become very confusing very fast.

I dont think this is a scenario that the core Tabulator library should be responsible for handling as i would class it as an edge case operation.

That being said, i think that there is already the facility to do what you ask for if you take a slightly different approach.

Tabulator provides the ***persistenceWriterFunc*** and ***persistenceReaderFunc*** callbacks to allow you to take manual control of the reading/writing process. you could easily add a bit of logic in these callbacks to tweak the id depending on if it is initial load or after column update.

More info on these callbacks can be found in the [Persistence Mode Documentation](http://tabulator.info/docs/4.7/persist#mode)

Cheers

Oli :)

gillez:
Hi Oli,

Thanks for the reply. I will check out those functions as it sounds like htey may do the trick. If I can take control of the reading and writing I may be able to persist the state of columns that are not currently in the table, and then recall the state when the columns are added. If I can do that then I won't need to change the ID.

Should I delete this branch?

Thanks,
Giles

PR #1692, opened 2019-01-02
Three State Sort Mode

erez86:
Hi Oli! :)

We added the ability to support both single and  multiple sort modes with the third (first load) sort mode which is none.

Today: if you do a sort on a column you cannot turn it off.
Fix: 1'st click is ASC, 2'nd click is DESC, 3rd click is back to NONE.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -57,6 +57,7 @@ Tabulator.prototype.defaultOptions = {
 	initialFilter:false, //initial filtering criteria
 
 	columnHeaderSortMulti: true, //multiple or single column sorting
+	threeStateSortMode: false, //three state sorting (asc, desc, none)
 
 	sortOrderReverse:false, //reverse internal sort ordering
 
--- a/src/js/modules/sort.js
+++ b/src/js/modules/sort.js
@@ -51,8 +51,12 @@ Sort.prototype.initializeColumn = function(column, content){
 			match = false;
 
 			if(column.modules.sort){
-				dir = column.modules.sort.dir == "asc" ? "desc" : (column.modules.sort.dir == "desc" ? "asc" : column.modules.sort.startingDir);
-
+				if(self.table.options.threeStateSortMode && column.modules.sort.dir === "desc") {
+					dir = "none";
+				}else{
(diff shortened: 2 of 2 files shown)


olifolkerd:
Im afraid i dont see the point in this addition.

If you click on a different column to sort the data is resorted by the new column, there is no state where the data is "not sorted"

Cheers

Oli :)

mikila85:
The server's default sort

mikila85:
https://www.ag-grid.com/example.php#/
Look at this main example of ag-grid
Three state sort as default.
I proposed here an optional three state sort for backward compatibility..

Please reconsider.

mikila85:
You are saying tabulator spec has only client side tables in mind? 
Because our tables are mostly ajax server side tables and by default they are sorted by ID which is not visible by any column and all fields are sortable. But if a user want to go back to the server's original sort he can be desorting.

mikila85:
And about what you said: there is no state where the data is "not sorted"

The initial state of tabulator is none. Every column start as none and no arrow is black. Once you click sort, you can never go back to the original table state where no arrow is black.

olifolkerd:
Hey @mikila85 

Im not saying that at all, what i am saying is that in tables without ***ajaxSorting*** enabled your suggested functionality would do nothing because the table has no intrinsic memory of the initial order of data when it was entered into the table and that should a feature like this be implemented it would need to be consistent across all versions.

And to answer your second comment, yes there is a default state of unsorted, but once that data is sorted there is no way to unsort it, the initial order has been lost.

When i approve pull requests i have to consider usage across all my users, and a very large proportion of those do not use ajax to pull data into their table

You option has also only been set to be enabled global for the table, which is not consistent with other column sort options which can be defined at the column level.

To enable this functionality across all modes of the table would involve considerable reworking of the row management pipeline to remember the initial state, and while i would certainly be happy to look into this in future it is not something i would be happy to accept a pull request on as it would affect the core functionality of the table.

Im not disagreeing that in your usage case it offers useful functionality, and you are of course welcome to keep this functionality in your fork which is the joy of open source coding, what i am saying is that it has wider implications beyond your usage case that this pull request does not cover.

Thankyou for your submission, but in this instance i would not be looking to pull it in to Tabulator. I will stick it on the roadmap so that when i do the next update to the row management pipeline i will look at adding in the functionality you request in a way that will work for all users.

Many Thanks

Oli :)

PR #1697, opened 2019-01-03
Fix scrollToRow bug when near the bottom row.

gillez:
There is a bug in ScrollToRow, when requested a row near the end of the table.  To fix this, an offset needs to be added to the scrollTop position.

To see the problem, set up a table with more rows than can fit on a single screen then call:

```
const rows = this.table.getRows(true);
this.table.scrollToRow(rows[rows.length - 1], "bottom", false);
```

This change request fixes this bug.

### DIFF
--- a/src/js/row_manager.js
+++ b/src/js/row_manager.js
@@ -189,7 +189,8 @@ RowManager.prototype.scrollToRow = function(row, position, ifVisible){
 	var rowIndex = this.getDisplayRows().indexOf(row),
 	rowEl = row.getElement(),
 	rowTop,
-	offset = 0;
+	offset = 0,
+	scrollOffset = 0;
 
 	return new Promise((resolve, reject) => {
 		if(rowIndex > -1){
@@ -232,7 +233,7 @@ RowManager.prototype.scrollToRow = function(row, position, ifVisible){
 				this.element.scrollTop = Tabulator.prototype.helpers.elOffset(rowEl).top - Tabulator.prototype.helpers.elOffset(this.element).top + this.element.scrollTop;
 				break;
 				case"virtual":
-				this._virtualRenderFill(rowIndex, true);
+				scrollOffset = this._virtualRenderFill(rowIndex, true);
 				break;
 			}
 
@@ -240,11 +241,11 @@ RowManager.prototype.scrollToRow = function(row, position, ifVisible){
 			switch(position){
 				case "middle":
 				case "center":
-				this.element.scrollTop = this.element.scrollTop - (this.element.clientHeight / 2);
+				this.element.scrollTop = this.element.scrollTop - (this.element.clientHeight / 2) + scrollOffset;
 				break;
 
 				case "bottom":
-				this.element.scrollTop = this.element.scrollTop - this.element.clientHeight + rowEl.offsetHeight;
+				this.element.scrollTop = this.element.scrollTop - this.element.clientHeight + rowEl.offsetHeight + scrollOffset;
 				break;
 			}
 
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @gillez 

Could you describe this bug in more detail as im not currently aware of an issue, what happens when you scroll to the bottom using the current codebase?

Cheers

Oli :)

gillez:
Hi Oli,

When I get a chance I'll try to create a CodePen to show this.  It may not be today but I'll try to do it this week if I can.

Thanks,
Giles

alainfrisch:
As written in #1864, this seems to fix the issue I observed as well.  At least under Chrome.  Under Edge, I still get a weird behavior for the same test:  as long as I don't resize the browser's window, scrollToRow still fails to make rows close to the end visible; also when scrolling to rows in the middle, some offset to be slightly off.   Probably some CSS derived property which is not updated at the correct time.  Video below (same code as in example in #1864):

tabulator_bug_edge

alainfrisch:
Same problem with IE12 and Firefox 60.0b11.

olifolkerd:
Hey @gillez 

I can see now that this is an issue, but im afraid that your pull request has a couple of issues at the moment.

It addresses the symptom not the problem, if resizing the table fixes the issue it indicates that the issue is located elsewhere in the table and that it is an initialization problem and not a render issue. I would prefer that the table is initialised correctly in the first place rather than that trying to work around another issue

Also the proposed solution does rather hijack the ***_virtualRenderFill*** function to return the offset value. since the purpose of the function is to fill the view it should not be returning the offset value, as that is a side effect of it working and only applicable in this edge case.

I will have a look into the issue in more detail this week and see if i c an isolate the root cause of the initial mis-configuration.

Cheers

Oli :)

gillez:
Hi Oli,

Thanks for looking at the pull request.  I hadn't actually realised that resizing the table fixes the issue but I agree that that would indicate that it's not a render issue.

If you can investigate and fix the mis-configuration soon that would be great.  Otherwise, I'll try to investigate the issue further when I can although not sure this will be in the next few days.

Cheers,
Giles

alainfrisch:
@gillez : yes, the video is after applying your patch.

alainfrisch:
I think there is a render issue (plus perhaps something else).  With my (limited) understanding of the current logic, one can get away without adding a return value to _virtualRenderFill  (or an extra argument).  IIUC, that function puts the requested row at the top of the view, and the caller then tweak the scroll offset to move it to the desired center/bottom position (under the assumption that the row was indeed at the top).  But since there is no content after the last row, it cannot be at the top after the call to _virtualRenderFill, and the caller will calculate a wrong offset.

alainfrisch:
Gentle ping to @olifolkerd ... Have you had a chance to investigate the root cause for the problem?

olifolkerd:
Hey @alainfrisch 

Im afraid i havnt had a chance to look into it yet, ive had a family member in hospital so ive been a bit distracted lately.

I appreciate that it is important that this gets resolved, but could i ask that you dont "ping" me. the issue is in the list and i will work through them when i can set aside a couple of days in the next few weeks to work through on the next bug release.  If an issue is open it is because i am aware it needs my focus and it will be prioritised for me to look at.

When i get users "pininging" me all that happens is that my message box gets full of "pings" and it makes what is already significant task of replying to the many messages i get each day that much harder because im having to also filter through peoples nudges and reminders.

Cheers

Oli :)

olifolkerd:
I have pushed a fix for this to the master branch, it will be included in this evenings patch release.

Cheers

Oli :)

PR #1714, opened 2019-01-09
fixed readOnly bug in select editor

galuha:


olifolkerd:
Hi@galuha 

I think you may have misunderstood things there.

If you don't want a cell to be editable simply don't include the ***editor*** in the column definition. if you want a cell to only be editable under certain circumstances you can set the ***editable*** callback.

Under no circumstances should an editor be read only, then it has no use as an editor, editors are solely for editing data.

Full details can be found in the [Editor Documentation](http://tabulator.info/docs/4.1/edit)

Cheers

Oli :)

galuha:
No, Oli. You don't understand. Playing with editable is not going to alter this
There is a bug.
If I don't include "select" editor, I won't get the select behavior.
But if I include, I do get not only select behavior, but a plaintext editor behavior on input field also because of this.
Like here: https://drive.google.com/open?id=1cRil7zgsReEzWv5IyzYRnBiVoLPWTF_r
There shouldn't be a cursor

galuha:
Sorry, Oli) I was in a rush, and I am not a native English speaker.
Thanks) I have edited my comment here and in the Issues not to look obrupt)

PR #1715, opened 2019-01-09
added return so that page load promises can be accessed above module scope

Eguthrie3214:
When trying
var table = new Tabulator('#table', {//config});
table.nextPage().then(function(){
})

nextPage() returns undefined.
you have to use table.modules.page.nextPage() in order to return the promise.

### DIFF
--- a/dist/js/tabulator_core.js
+++ b/dist/js/tabulator_core.js
@@ -6191,7 +6191,7 @@ Tabulator.prototype.previousPage = function () {
 	if (this.options.pagination && this.modExists("page")) {
 		return this.modules.page.previousPage();
 	} else {
-		return false;
+		return Promise.reject(new Error('Module does not exist'));
 	}
 };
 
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @Eguthrie3214 

Thanks for your pull request.

I agree with the principal, and on the whole i would be happy to approve, but to be a proper implementation it would need to always return a promise rather than sometime return a promise and some time return a boolean.

Could you make it so that it always returns a promise and that promise either resolves passing on the resolution from the module promise, or rejects if the module promise rejects or if the current iff statement returns false.

Cheers

Oli :)

Eguthrie3214:
Hi Oli, thanks for the quick reply. I used the static Promise.reject() in order to return a rejected promise when the module does not exist. If you would like different verbiage in the reject message or something else, please don't hesitate to say. 

Thanks, 
Evan

PR #1725, opened 2019-01-14
Number sorter: parse numbers with multiple commas

### LINKED ISSUE #1724: Number sorter breaks on numbers with more than 2 commas
**Describe the bug**

When the value in a column is "1,000,000" or more, the number sorter breaks.

**Tabulator Info**

4.1.4

**To Reproduce**

```javascript 
  var table = new Tabulator("#inventory-table", {
    data: [
      {"inventory": "1,000,000"},
      {"inventory": "950,000"},
      {"inventory": "500"},
    ],
    columns: [
      {title: "Inventory", field: "inventory", sorter: "number"}
    ]
  });
```

**Expected behavior**

Descending Sort: 1MM, 950k, 500
Ascending Sort: 500, 1MM, 950k

**Actaul behavior**

Descending Sort: 950k, _1MM_, 500
Ascending Sort: 500, _1MM_, 950k

**The culprit?**

[String.replace()](https://github.com/olifolkerd/tabulator/blob/master/src/js/modules/sort.js#L295-L296) only replaces the first instance (unlike in python). A combo of split/join must be used.

Therefore, "1,000,000" is evaluating to `1000`

Brobin:
Resolves #1724 So that numbers with more than one comma in them will be parsed correctly.

I also added 2 parameters `decimalSeparator` and `thousandSeparator` for the number format, but could not find the docs in this repo to update.

### DIFF
--- a/src/js/modules/sort.js
+++ b/src/js/modules/sort.js
@@ -291,11 +291,11 @@ Sort.prototype.sorters = {
 	number:function(a, b, aRow, bRow, column, dir, params){
 		var alignEmptyValues = params.alignEmptyValues;
 		var decimal = params.decimalSeparator || ".";
-		var separator = params.thousandsSeparator || ",";
+		var thousand = params.thousandSeparator || ",";
 		var emptyAlign = 0;
 
-		a = parseFloat(String(a).split(separator).join("").split(decimal).join("."));
-		b = parseFloat(String(b).split(separator).join("").split(decimal).join("."));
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @Brobin 

Thanks for the pull request, i think that is a great idea. Could i ask that you change the name of the parameters to ***decimal*** and ***thousand*** to match with the same parameters used with the money formatter, just to keep things consistent.

I would then be happy to approve the request.

Cheers

Oli :)

Brobin:
@olifolkerd No problem! Updated

Issue #1721, opened 2019-01-11
setHeaderFilterValue() Render Problem


OOnogul:
**Description**
When setHeaderFilterValue() is called with a value on a column that has the following definitions, nothing is changing in header filter input element. But if clicked/focused the value is rendered/shown.

The definition part with the header filter:
```
headerFilter: "select"
headerFilterParams: {values: true}
```

**Tabulator Info**
- Tabulator version 4.1.5

**To Reproduce**
1. Call setHeaderFilterValue() with valid parameters on a column that has select headerfilter.
2. See that there is no change in the text of header filter input element.
3. Click the header filter and see it as rendered.

**Expected behavior**
The value should be visible when its set on header filter as the other columns that is not `headerFilter: "select" `

Browser: Chrome v67.0.3396.87

Thanks in advance.

olifolkerd:
Hey @OOnogul 

Thanks for getting in touch, sorry to hear you are having trouble.

Could you create a JS Fiddle or a CodePen that demonstrates how your are using the table and how you are calling the function. It is hard to offer support without seing the code you are using.

Cheers

Oli :)

OOnogul:
Yeah of course, I have recreated the problem with a demo in the link below.

https://jsfiddle.net/vmq6eosn/1/

You can see that after constructing the table there is two lines of code that sets the header filter values.
One for a input type header filter, second one is setting a enum type header filter. As you can see after setting headerfilter the value is seen in the filter element.
But in the second one, altough the filter is applied to data, the value is not visible until you click in the header filter input element.

I hope I have been able to express the issue that I have come across. 

Many thanks

olifolkerd:
Hey @OOnogul 

Thanks for the fiddle, it was very helpful in isolating the issue. That is definitely a bug, i will look at fixing it for the 4.2 release.

Cheers

Oli :)

olifolkerd:
The issue was caused by the editor not correctly displaying the initial value of the cell, when used as an editor or header filter.

I have pushed a fix for this to the 4.2 branch and will include it in mondays release.

Cheers

Oli :)

PR #1729, opened 2019-01-16
Parse select editor options and set current item when creating editor

ddetkovs:
---
name: Select editor header filter input box not populated with initial value on creation.
---

**Describe the bug**
When using setHeaderFilterValue on a column with select editor tabulator doesn't calculate the displayItems and doesn't set the current item/initial value on the input text box.
This is different from select editor in cells, since editor is only visible when user clicks onto the cell, which causes it to calculate display options, create a dropdown and consequently set the value on the input text box.

**To Reproduce**
1. use setHeaderFilterValue on a column with select editor header filter.

Alternatively, I think this can be reproduced with persistent config.

### DIFF
--- a/src/js/modules/edit.js
+++ b/src/js/modules/edit.js
@@ -650,8 +650,6 @@ Edit.prototype.editors = {
 
 			dataItems = dataList;
 			displayItems = displayList;
-
-			fillList();
 		}
 
 		function fillList(){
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @ddetkovs 

This is a duplicate of the same issue #1721 which is scheduled to be resolved in the 4.2 release.

Cheers

Oli :)

olifolkerd:
Hey @ddetkovs 

Sorry for not getting back to you on this for so long, i usually prefer to engage more with users submitting pull requests.

While your fix does handle the specific issue you were facing, it was actually rooted in a deeper issue of setup values in the select editor. i have completed a more comprehensive update to the select editor so this pull request is no longer needed.

Thanks for you suggestion, i will endeavour to engage with you faster next time you make a pull request.

Cheers

oli :)

Issue #1880, opened 2019-03-08
Export xlsx doesn't work


porkydenny:
Tryng to create an export in xlsx of a table it return: 
Uncaught ReferenceError: i is not defined
    at tabulator.min.js:6
    at Array.forEach (<anonymous>)
    at c (tabulator.min.js:6)
    at tabulator.min.js:6
    at Array.forEach (<anonymous>)
    at a (tabulator.min.js:6)
    at r (tabulator.min.js:6)
    at w.xlsx (tabulator.min.js:6)
    at w.download (tabulator.min.js:6)
    at t.d.download (tabulator.min.js:5)

Link with simple test tryng the lib, hope it will help!
https://jsfiddle.net/7ehp0bto/

**To Reproduce**
Click on xlsx button

gravjean:
The function parseRow use a variable i.

But this variable dont exist. I think the variable to use is value event if i dont understand why the value cant be date.

[
rowData.push(!(**i** instanceof Date) && (typeof value === 'undefined' ? 'undefined' : _typeof(value)) === "object" ? JSON.stringify(value) : value);
}
To
[
rowData.push(!(**value** instanceof Date) && (typeof value === 'undefined' ? 'undefined' : _typeof(value)) === "object" ? JSON.stringify(value) : value);
]

in tabulator.

porkydenny:
Hi gravjean,

modifing in that way the export work correctly.
Awesome!!!

Thank you for support!

PR #1859, opened 2019-03-04
[bugfix] ReferenceError: i is not defined - when exporting to Excel

runawaycoin:
Sorry, the variable with the PR was incorrect
https://github.com/olifolkerd/tabulator/pull/1841

PR #1882, opened 2019-03-08
Export xlsx doesn't work #1880

gravjean:
Correction of this issue : https://github.com/olifolkerd/tabulator/issues/1880

But im not surr on the logic of this new condition


gravjean:
Oops sorry i just see the bug already have a pull request here : 

https://github.com/olifolkerd/tabulator/pull/1859

PR #1912, opened 2019-03-18
Fix border-radius in Semantic UI css

Primajin:
using the right border radius from Semantic UI


olifolkerd:
Hey @Primajin 

Thanks for your pull request, unfortunaly you have made the changes to the CSS files in the ***dist*** directory.

All files in this directory are automatically generated when new code is added, so your changes will be overritten next commit.

All pull requests should be made on the files in the ***src*** directory, which are then used to generate the dist files.

if you can make your change to the appropriate SCSS files in 'src\scss\semantic-ui` then i would be happy to accept this pull request.

Cheers

Oli :)

Primajin:
Of course! I must have been not completely awake :)

olifolkerd:
Hey @Primajin 

Im confused as t why this is needed, the variables file was built straight from code from semantic-ui so unless there was an issue with their sass variable it should be working.

Primajin:
Ahh I was checking their (semantic UI's) code and I only saw that they are using less with `unit({something}, rem)` - which is not available in sass.

You are telling SASS to calculate `integer divided by pixel plus rem`.

Using mixed units in SASS based on the implementation will either

- cause an error right away (vanilla SASS)
- cause something weird like `0.2857142857px^-1` and an error (Dart SASS)
- cause it to being handled as a pure string and result in `0.28571/pxrem` (Lib SASS) which is bad because it silently swallows the error anticipating stuff for you

Screenshot 2019-03-26 at 09 16 24

Screenshot 2019-03-26 at 09 15 39

Screenshot 2019-03-26 at 09 16 02

Since the latter is what I saw in your compiled code, I assume you are using Lib Sass.

You can play around in https://www.sassmeister.com/
That's why I was introducing the new function to strip away the units, so it works regardless of the compiler implementation.

If you say you just took the source from Semnatic UI directly, then either they are using a fourth implementation that is fine with the syntax, they might use some special framework like COMPASS that handles that differently or they have overloaded some functionality included in an earlier file that is not throwing up during compile time at that point.

Or I'm completely overlooking something here, also possible since I was working on the wrong files in the beginning 😅

Primajin:
Can we go on with this?

- [ ] Yes 🎉 
- [ ] No 😞 
- [ ] Maybe 🤷‍♂️

olifolkerd:
Hey @Primajin 

Sorry for not getting back to you on this, one of my family has been in hospital so ive been a bit distracted the last week.

You make a very valid point, i would be happy to merge this pull request.

Cheers

Oli :)

Primajin:
Oh no worries! I hope they get well soon!

PR #1942, opened 2019-03-29
closes #1940: arrow keys in "select editor" scrolls the editor grid

### LINKED ISSUE #1940: Using up/down arrows when using the "select editor" scrolls the entire grid
**Describe the bug**

**Tabulator Info**
- Which version of Tabulator are you using? 4.2
- Post a copy of your construct object if possible so we can see how your table is setup

**Working Example**
tabualtor-scroll-bug

**To Reproduce**
1. Go to the edit demo on the docs page (http://tabulator.info/docs/4.2/edit)
2. Click on any cell that has a dropdown/select editor (such as "gender")
3. When the editor opens... use the up/down keys to select a different value
4. Notice how value in the select box does change but the grid also scrolls unexpectedly 

**Expected behavior**
The table should not scroll

**Screenshots**
attached previously

 - OS: MacOS and Windows
 - Browser: Chrome
 - Version: latest

**Additional context**
The issue is also present in the documentation demo.
If the inner grid is not scrollable then the entire windows ends up scrolling.

morficus:
* root cause: when using the up/down keys, the `keydown` event were bubbling up out of the select editor and being caught by either containing grid (or the window, if the grid doesn't have internal vertical scroll bars)
* solution: catch the event in the `keydown` event from bubbling up when using the arrow keys with in the editor
* also adding code to deal with the left/right keys to prevent the table from scrolling horizontally while the select editor is in focus
* extending the bug fix to the "autocomplete" editor as well

### DIFF
--- a/src/js/modules/edit.js
+++ b/src/js/modules/edit.js
@@ -795,6 +795,7 @@ Edit.prototype.editors = {
 				case 38: //up arrow
 				e.stopImmediatePropagation();
 				e.stopPropagation();
+				e.preventDefault();
 
 				index = dataItems.indexOf(currentItem);
 
@@ -806,6 +807,7 @@ Edit.prototype.editors = {
 				case 40: //down arrow
 				e.stopImmediatePropagation();
 				e.stopPropagation();
+				e.preventDefault();
 
 				index = dataItems.indexOf(currentItem);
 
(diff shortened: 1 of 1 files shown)


morficus:
Hey @olifolkerd - not sure what the "PR etiquette" is here, but I only made changes to the src and did not create an updated `dist` output. Let me know if you would like me to generate that as well.

olifolkerd:
Hey @morficus 

Great suggestion!

Could i ask if you have checked through all the editors you have changed to ensure that the foucs still shifts to the cell below in the case of input editors etc.

Cheers

Oli :)

morficus:
@olifolkerd - yes sir.
Here is a Fiddle with the test which is using my new changes: https://jsfiddle.net/morficus/h0g3x9er/2/

olifolkerd:
great stuff! thanks for the fiddle, i will include this in this weekends patch release.

Cheers

Oli :)

Issue #1959, opened 2019-04-04
virtualDomBuffer setting has no effect


sibsibsib:
**Describe the bug**
setting `virtualDomBuffer` as [specified in the docs](http://tabulator.info/docs/4.2/layout#virtual-dom-buffer) has no effect. 

**Tabulator Info**
- Which version of Tabulator are you using? **4.2.3**

**Additional context**
On investigation, I found that this is because the setting is not present in `defaultOptions` and is thus not copied into the table options on init. Adding it fixes the issue.

*PR to follow*

PR #1960, opened 2019-04-04
add virtualDomBuffer to defaultOptions

sibsibsib:
hello, here is a fix for issue #1959


olifolkerd:
Great spot! I must have accidentally removed that at some point.

I will include the fix in the next patch release over the weekend

Cheers

Oli :)

PR #1971, opened 2019-04-10
Replaced entity nbsp with UTF code (\u00a0)

TommySedin:
Using entity name "&nbsp" breaks initialization in XHTML pages in browsers that do not load the XHTML DTD (Chrome, in my case). I've simply replaced all occurrences of &nbsp with the UTF character code. Could not find any other html entities that aren't supported (eg. &lt; and &gt; works fine).

Note that I have **not** run any tests other than my own use-case. ("Works fine for me!")

### DIFF
--- a/src/js/column.js
+++ b/src/js/column.js
@@ -593,19 +593,19 @@ Column.prototype._buildColumnHeaderTitle = function(){
 
 		if(def.field){
 			table.modules.localize.bind("columns|" + def.field, function(text){
-				titleElement.value = text || (def.title || "&nbsp");
+				titleElement.value = text || (def.title || "\u00a0");
 			});
 		}else{
-			titleElement.value  = def.title || "&nbsp";
+			titleElement.value  = def.title || "\u00a0";
 		}
 
 	}else{
 		if(def.field){
 			table.modules.localize.bind("columns|" + def.field, function(text){
-				self._formatColumnHeaderTitle(titleHolderElement, text || (def.title || "&nbsp"));
+				self._formatColumnHeaderTitle(titleHolderElement, text || (def.title || "\u00a0"));
 			});
 		}else{
-			self._formatColumnHeaderTitle(titleHolderElement, def.title || "&nbsp");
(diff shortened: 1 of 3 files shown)


olifolkerd:
Hey @TommySedin 

Im a bit confused, i have never seen a scenario when it dosn't work. Chrome respects &nbsp, i use it all the time.

It sounds like a misconfiguration of the character encoding of your site unless you have a very niche usage case.

I would need to understand more about why this is needed before moving away from a standard practice.

Cheers 

Oli :)

TommySedin:
Hello,
Done some more experimenting and it turns out that the problem is with XHTML 1.1 specifically - if you specify XHTML 1.0 it works fine - Chrome defaults to v1.1 when rendering XHTML.

There are actually only 5 html entities specified in XHTML 1.1: https://cweiske.de/tagebuch/xhtml-entities.htm

Here's a very simple test case:
`<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1 //EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<body><script>document.body.innerHTML = "f&nbsp;oo";</script></body>
</html>`

If you save that with extension .xhtml, Chrome gives an error "error on line 3 at column 49: Entity 'nbsp' not defined". Rename it to .html, it works fine.

Replace the &nbsp; with the UTF code (\u00a0), it works in XHTML 1.1 as well.

That said, I should of course have used &#160; instead since that's more common. (I'm not primarily a web-developer, my brain goes "Replace everything with UTF!" whenever I run into anything related to character representations. :))

/Tommy

olifolkerd:
Hey @TommySedin 

Can i ask what environment you are trying to use this in, under normal web development you would be saving the files as html not xhtml?

Cheers

Oli

TommySedin:
It's just a personal preference for me - I find it easier to catch any
mistakes with tags and such with the stricter syntax of xhtml.

/Tommy

Den lör 13 apr. 2019 kl 20:39 skrev Oli Folkerd <notifications@github.com>:

PR #1886, opened 2019-03-11
fix: TickCross Filter

akki-jat:
fix: pull request (#1875) made with changes done in dist folder ( which will be overwritten on rebuild)  instead of from src folder

PR #1980, opened 2019-04-15
fix: tickCross & autocomplete Filter

akki-jat:
fix condition changed in #1886 (because of changed in condition live update were also applying on tickCross & autocomplete which we were avoiding)

### DIFF
--- a/src/js/modules/filter.js
+++ b/src/js/modules/filter.js
@@ -258,12 +258,12 @@ Filter.prototype.generateHeaderFilterElement = function(column, initialValue){
 
 				if (
 					!(
-						(column.definition.headerFilter === 'autocomplete' ||
-							column.definition.editor === 'autocomplete' ||
-							column.definition.headerFilter === 'tickCross' ||
-							column.definition.editor === 'tickCross') &&
-						column.definition.headerFilter === true
-					)
+                        column.definition.headerFilter === 'autocomplete' ||
+                        column.definition.headerFilter === 'tickCross' ||
+                        ((column.definition.editor === 'autocomplete' ||
+                            column.definition.editor === 'tickCross') &&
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @akki-jat 

i do not follow you, what is this fixing exactly?

src/js/modules/filter.js
```diff
@@ -258,12 +258,12 @@ Filter.prototype.generateHeaderFilterElement = function(column, initialValue){
 
 				if (
 					!(
-						(column.definition.headerFilter === 'autocomplete' ||
-							column.definition.editor === 'autocomplete' ||
-							column.definition.headerFilter === 'tickCross' ||
-							column.definition.editor === 'tickCross') &&
-						column.definition.headerFilter === true
```
akki-jat: Hi @olifolkerd 

Here what happening is we **applying and (&) conditions with all of or (||) conditions** (**not wrapping the and conditions so would not affect all or (||) conditions**) which returns false in case of 'autocomplete' and 'tickCross' but it should be true.

Because of this `column.definition.headerFilter === true` condition which returns false  in case of 'autocomplete' and 'tickCross' and we are doing and condtion with all of the or conditions and the applying not (!)

src/js/modules/filter.js
```diff
@@ -258,12 +258,12 @@ Filter.prototype.generateHeaderFilterElement = function(column, initialValue){
...
-					)
+                        column.definition.headerFilter === 'autocomplete' ||
+                        column.definition.headerFilter === 'tickCross' ||
+                        ((column.definition.editor === 'autocomplete' ||
```
akki-jat: So here I am just wrapping the and (&) condition so it would not affect the all conditions.

Issue #1618, opened 2018-12-03
Selecting all rows calling rowSelectionChanged repeatedly


gillez:
Version: Tabulator v4.1.3

When selecting multiple rows with table.selectRow(), I would expect the rowSelectionChanged callback to be called once, after all rows are selected.  However, it is being triggered repeatedly as all the rows are selected.  This takes much longer than I think it needs to.

e.g. I have a table with 5000 rows.  If call table.selectRow() to select all the rows, the rowSelectionChanged callback is called 5000 times, first with one row, then two rows, then three rows, etc.

I would expect the callback to ba called once, after all rows have been selected, and passed 5000 rows.

I have fixed this by modifying the SelectRow.prototype.selectRows function to call
```self._selectRow(row, true, true);``` at the three points in selectRows where selectRow is called for multiple rows.  The middle parameter (silent) is now true which stops _rowSelectionChanged from being called as each row is added.  With this change, selectRow() took about 10 ms rather than 600 for me, but obviously the time taken depends on how much work is being done in rowSelectionChanged.

Is this intended behaviour or do you agree that it is a bug?  If so, I can submit this as a pull request.

olifolkerd:
Hey @gillez 

Sorry for the delay getting back to you on this one, ive had a busy couple of weeks at work.

yes i would say this is a bug, please feel free to submit a pull request, that would be great,

Thanks

Oli :)

olifolkerd:
Hey @gillez 

I have pushed a fix for this to the master branch and will include it in this evenings patch release.

Cheers

Oli :)

gillez:
Hi Oli,

Apologies for not submitting a pull request for this.  I was busy last week and didn't get a chance to do it.  

Thanks for including a fix in the latest release.  However, In the latest code the fix is not complete.  You have changed the code only for the "default" case, which applies if a range of rows is passed.  The same fix also needs to be applied to the "undefined" and "boolean" cases, which are used when selecting all rows or all active rows.  Also, the same change should be applied to deselectRows() otherwise the callback is called multiple times when  removing the selection.

I will check the new code out today, make these changes and submit a pull request.

Cheers,
Giles

gillez:
Hi Oli,

I have created a pull request for this.

Cheers,
Giles

PR #1986, opened 2019-04-16
issue #1618 for 3.5

dcanning:
fix for issue #1618 repeated_rowSelectionChanged issue in the 3.5 branch


olifolkerd:
Im afraid I only accept pull requests against the active branch.

As i have explained in other posts there simply isnt the time and resources to maintain main active versions of Tabulator and i only support the latest version.

Tabulator 3.5 had many issues which were resolved in the move to 4.0 for that reason it would be a massive burden to continue supporting it.

Cheers

Oli :)

PR #2066, opened 2019-05-15
[FEATURE] Add json string support for select editor params

jonasreymen:
Useful when using tabulator on an existing table so you can pass options as a JSON string.


olifolkerd:
I dont understand why this would be needed, why can you not parse your string before passing it into Tabulator?

PR #2084, opened 2019-05-21
Update to allow flags for add / remove performance

hookedupjoe:
This proposed update is to allow a flag to be set when updating data in bulk.  
*** Specifically when using the reactiveData option

table.blockRedraw = true;
 ... do mass updates
table.blockRedraw = false;

Note: To do mass deletes, do not splice them individually.

How I was able to remove hundreds in a high performance way ..

1) Sort data by remove flag or not.   
2) Find first to be removed.  
3) Splice from there on ... was very fast this way

YOURTABLEDATA.splice(tmpFirstPos, tmpToClean);

### DIFF
--- a/src/js/row_manager.js
+++ b/src/js/row_manager.js
@@ -344,6 +344,10 @@ RowManager.prototype._wipeElements = function(){
 }
 
 RowManager.prototype.deleteRow = function(row, blockRedraw){
+	//WJF - Allow global flag to block redraw
+	if( this.table.blockRedraw === true ){
+		blockRedraw = true;
+	}
 	var allIndex = this.rows.indexOf(row),
 	activeIndex = this.activeRows.indexOf(row);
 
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @hookedupjoe 

Thanks for getting in touch, while i appreciate your idea, and can certainly see some good usage cases for it, however i do have some concerns about your approach.

At no point does Tabulator ever allow users to manipulate properties directly on the table object, all interaction with Tabulator must be through functions as defined in the table API, this makes the table more extensible by allowing actions to be triggered as a result of the action, rather than just using un watched properties. for example there should be some functions of ***disableRedraw*** and ***enableRedraw***

Also the functionality you are suggesting is handled exclusively in the row_manager, so i would expect the variable to be held there after being set by external api functions.

Please do not leave custom comments with your initials in the code, your code should either be self explanatory through correct variable naming, or need consistent comments. If you have made a specific change that you want to explain that should be included in your commit message or as a note in your pull request

I would be happy to consider this for inclusion if you could make the above updates.

Cheers

Oli :)

hookedupjoe:
I'll attempt to rework that in the latest version, but really was reporting that without this minor change, the system is so slow as to not be usable in some cases.  Before this was a show stopper and after a full day of working down to the root cause, this was it.  I figured the best way to report the changed needed was to show code that fixed the problem.   I really didn't expect you would just "pop it in there" but instead would look at the write up and code changes and make them in the way you would do in your system, that works with the rest of it as expected, I couldn't say if the changes would cause issues somewhere else, etc .. so to make them "perfect" would have been a waste of time.

Again, I"ll see if I can understand what your saying I need to change to make this code change the same way you would so you can accept it as I would like to use upgrades without having to manually fix this issue every time myself.

Thanks for the great tool and reply.   Cheers.

patissi:
good like

olifolkerd:
Hey @hookedupjoe 

I have pushed a more wholistic fix for this to the 4.5 branch which will be released in the next few days.

There are now a ***blockRedraw*** and ***restoreRedraw*** function that can be used to pause the table redrawing action.

The updated functionality will take account any actions that happen that would need a redraw and their level in thee data processing pipeline, as well as any render in place actions. On calling the ***restoreRedraw*** function it then triggers the appropriate level of redraw.

Cheers

Oli :)

Issue #2064, opened 2019-05-14
row.update() doesn't seem to update child/sub values


jlschaper:
I think I have found a small(ish) bug to do with how row.update(JSON) doesn't update children/sub values.

I'm grabbing (quite happily via AJAX) a JSON in the style of:
```javascript
[{"ID":1,
  "clientID":89,
  "client":  { 
     "ID":89,
    "clientName":"ABC (Test Client)",
    "clientAddress":null
  },
  "jobName":"Test Joby",
  "jobAddress":"Example\r\ny",
  "jobDetails":"etc etc etc"
},
{"ID":2,
  "clientID":89,
  "client":  { 
     "ID":89,
    "clientName":"ABC (Test Client)",
    "clientAddress":null
  },
  "jobName":"This is test",
  "jobAddress":"123 Fake Street",
  "jobDetails":"etc etc etc"
},
{"ID":3,
  "clientID":89,
  "client":  { 
     "ID":89,
    "clientName":"ABC (Test Client)",
    "clientAddress":null
  },
  "jobName":"Is ABC (89) but to change",
  "jobAddress":"Please Enter Job Address..",
  "jobDetails":"etc etc etc"
}]
```
Which is JSON serialized from two objects classes in PHP without any issues.  Initial load is perfect (and any manual refresh with table.replaceData('AJAX.php') calls etc).
But when I use row.update(myXMLHttpRequest.responseText) only the top level (ID, jobName, jobAddress, & jobDetails) update in the grid leaving the child ones unchanged (client.ID, client.clientName, & client.clientAddress) displaying the old values.

```myXMLHttpRequest.responseText``` being something akin to:
```javascript
{"ID":3,
  "clientID":81,
  "client":  { 
     "ID":81,
    "clientName":"Test Client",
    "clientAddress":null
  },
  "jobName":"Was ABC (89) now another",
  "jobAddress":"Please Enter Job Address..",
  "jobDetails":"etc etc etc"
}
```

I know the underlying row object *has* replaced the values with the new one (I've console logged the row and shows correctly updated JSON including client sub JSON) but not updated the cells that display that data.  If I resort the columns (any column) it *does* update.

Using Tabulator v4.2.7 (latest)

Table columns (abridged for clarity)
```javascript
            {title:"ID", field:"ID", visible:false},
            {title:"Job Name", field:"jobName", editor:"input", validator:["required", "maxLength:32"]},
            {title:"Client", field:"client.clientName"},
[...]

PR #2089, opened 2019-05-22
Issue #2064 row.update() doesn't seem to update child/sub values.

aklaver:
Modify row.js to descend into nested data on row.update() and update the cells.
I am still relatively new to Javascript, so suggestions as to applicability of this code are welcomed.  The comments inline explain what I am doing.

### DIFF
--- a/src/js/row.js
+++ b/src/js/row.js
@@ -534,20 +534,50 @@ Row.prototype.updateData = function(data){
 			this.table.modules.reactiveData.unblock();
 		}
 
-		//update affected cells only
+		//update affected cells only. First build an array(cellNameArray) of cell
+		//names from the row data, descending down into any nested data and
+		//building cell names for those in dotted notion name.name. Then loop over
+		//the array and find cell for each name. Split cell name, if splits into more
+		//then one value use both names to look up value in row data. Otherwise
+		//use just the single name. Test to see if data has changed, if so update
+		//cell and render.
+		var cellNameArray = [];
+		var found = false;
 		for (var attrname in data) {
-			let cell = this.getCell(attrname);
+			{if (data[attrname].constructor == Object ){
+				for (var nestedAtt in data[attrname]){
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @aklaver 

Im a bit unsure as to the usage case of this. if you update the ***_children*** property of a row with the ***update*** function it already refreshes the child rows?

What am i missing?

Cheers

Oli :)

aklaver:
This is not for Data Tree:
http://tabulator.info/docs/4.2/tree
but for nested data in a field:
http://tabulator.info/docs/4.2/columns#field-nesting
See:
https://github.com/olifolkerd/tabulator/issues/2064

olifolkerd:
Ahh with you now.

There is actually already a system for handling this built into the column management system (it is a smidge more complicated than exploding on the "." as tabulator allows for custom delimiters and other such fun).

You have definitely spotted an area that needs improvement, but it actually should only take a line of code to tie it into the correct part of the system.

I will make a commit with the fix in tomorrow and post a link to it here so you can see how it is done incase you come across any others like it.

Cheers

Oli :)

aklaver:
Did you ever make the change above? If so can you point me at it. Thanks.

psmitty7373:
I am suffering form this same issue, updateRow is failing to update fields that are object properties.  Redraw works, but causes other issues for me.

olifolkerd:
@psmitty7373 

Please post a JS Fiddle that demonstrates your issue and we can offer you some advice.

If you feel this is a new issue please create a new issue.

Cheers

Oli :)

olifolkerd:
Ahhhhh I have just finally cottoned on to what it was you are trying to achieve here.

Yes that was definitely a needed change.

Looking at your proposed solution it was a bit verbose and revolved around build an array that already exists in the column manager and then it only handled one level of nesting, where there are already systems built into the columns objects to handle data management of nested data.

I have just pushed a fix for this to the master branch and will include it in a patch release today.

Cheers

Oli :)

PR #2090, opened 2019-05-23
Update jquery_wrapper.js

pmulreid:
Change AMD dependencies


olifolkerd:
Looks good to me, i will merge it in now and include it in the next release.

Thanks for the contribution.

Cheers

Oli :)

olifolkerd:
Just a quick not on this update. it turns out this cause a regression that stopped the wrapper working in web browsers, in the else factory function you passed the Tabulator object to the 2nd instead of 3rd argument

Issue #2109, opened 2019-05-30
true for getDataCount() not respected


mestaritonttu:
1. Open https://codepen.io/anon/pen/zQmWRx
2. Open dev tools console
3. Input a number in the Progress filter, f.ex. 10

The total number of rows (5) keeps appearing in the console instead of the expected filtered number.

tentus:
First off, this highlights that the `getDataCount` function operates the opposite of the documentation (passing a truthy value counts all rows, while omitting the param or passing a falsey value counts only the active rows). 

More bewilderingly, this pen also highlights that `dataFiltered` seems to be step offset from where it should be (showing the count from the previous filter results, not the current results).

@mestaritonttu in the meantime, it seems like you could probably utilize ```renderComplete: function() { console.info(this.getDataCount()); }``` to achieve the desired behavior.

mestaritonttu:
Thanks a lot, @tentus ! I can't believe I did not test without true :D
Yep, renderComplete is fine as that is what I am using in my actual project.

aklaver:
Submitted PR for getDataCount(true) not resolving correctly:

https://github.com/olifolkerd/tabulator/pull/2113

Have not figured out lag in count results yet.

aklaver:
Figured it out. In filter.js is 

` 
if(self.table.options.dataFiltered){

                activeRows.forEach(function(row){
                        activeRowComponents.push(row.getComponent());
                });

                self.table.options.dataFiltered.call(self.table, self.getFilters(), activeRowComponents);
        }

        return activeRows;
Which means the rows are returned after dataFiltered. As it turns out that does not matter as the signature for a dataFiltered function is; dataFiltered:function(filters, rows) where rows are the active rows.
If you change your code to:
`
dataFiltered:function(filters, rows) {
        console.info(rows.length);
`
it will work.

olifolkerd:
Hey @aklaver 

Thanks for the pull request, i have just merged that in.

Cheers

Oli :)

PR #2113, opened 2019-05-30
Issue #2109 true for getDataCount() not respected.

aklaver:
Order of evaluation for active is backwards in row_manager.js for
getDataCount(). Fix by swapping order.


olifolkerd:
Good Spot!

PR #2119, opened 2019-06-04
Declare textRange

kenjiuno:
- Fix ASP.NET Core Bundler (BuildBundlerMinifier) problem:
  `Bundler & Minifier error 0: Strict-mode does not allow assignment to undefined variables: textRange`


olifolkerd:
Good Spot!

PR #2138, opened 2019-06-10
Fix missing reference to scroll header elements.

### LINKED ISSUE #2137: Header row scroll position not updating
**Describe the bug**
The header row does not properly update its horizontal scroll position when the body scrolls.

**Tabulator Info**
- Which version of Tabulator are you using?
4.2.7

**Working Example**
https://jsfiddle.net/haferje/q7o8b0mv/

**To Reproduce**
1. Run the jsfiddle

**Expected behavior**
Header row should scroll along with the data rows.

**Additional context**
When the user has horizontally scrolled the grid, and we remove and re-add a row, the body scrolls all the way to the left, but the header horizontal scroll is not updated to match.

haferje:
Closes #2137.

I know there were changes recently to update the header horizontal scroll position, but I think there was a reference that was missed.  The functionality of RowManager.scrollHorizontal() should match that of the "scroll" event handler.  The JSFiddle example in #2137 shows a particular situation where the scroll is not updated properly.

### DIFF
--- a/src/js/row_manager.js
+++ b/src/js/row_manager.js
@@ -864,6 +864,8 @@ RowManager.prototype.scrollHorizontal = function(left){
 	this.scrollLeft = left;
 	this.element.scrollLeft = left;
 
+	this.columnManager.scrollHorizontal(left);
+
 	if(this.table.options.groupBy){
 		this.table.modules.groupRows.scrollHeaders(left);
 	}


haferje:
Oli,
In the related issue, you said you would re-evaluate. Should we reopen this PR?

PR #2148, opened 2019-06-13
Issue #2114 - Row mulit-select on Mac using Command key

### LINKED ISSUE #2114: Row mulit-select on Mac using Command key
**Describe the bug**
Users are conditioned to use the Command key when selecting multiple rows on Mac computers, but the code only matches against `ctrlKey`.

**Tabulator Info**
- Which version of Tabulator are you using?
v4.2.7

**Expected behavior**
The Command key on Macs should work as Ctrl key in Windows to multiselect rows.

**Additional context**
I do not have a ton of experience with Mac computers, but in most cases, the Cmd key does in Mac OS, what the Ctrl key does in Windows.  Users alerted me that the Command key does not work with selecting multiple rows in the grid.  The code in `modules/select_row.js` only checks for `ctrlKey` on the keypress event object.

After researching and testing how to capture the Command key, I have found out that different browsers report different data for the event.  The Cmd key on Mac is equivalent to the Win key on Windows, which reports 91 as the keycode (`event.which`) in Chrome, Firefox, and Edge, and IE11.  On Mac OS, the Cmd key reports keycode 91 under Safari and Chrome, but 224 under Firefox.  The OSes and browsers report other vastly different values for `event.key` and `event.code`, so your best bet is to probably code off of `event.which` to capture multiselect intent on Mac.

I don't believe keycode 224 is mapped to anything else, so you could probably just write:
`var ctrlKey = e.ctrlKey || e.which == 91 || e.which == 224;`

haferje:
Added 'meta' key capture to row multi-select logic, for Cmd key on Mac OS.
Closes #2114

### DIFF
--- a/src/js/modules/select_row.js
+++ b/src/js/modules/select_row.js
@@ -54,7 +54,7 @@ SelectRow.prototype.initializeRow = function(row){
 						var rows = self.table.rowManager.getDisplayRows().slice(0);
 						var toggledRows = rows.splice(fromRowIdx, toRowIdx - fromRowIdx + 1);
 
-						if(e.ctrlKey){
+						if(e.ctrlKey || e.metaKey){
 							toggledRows.forEach(function(toggledRow){
 								if(toggledRow !== self.lastClickedRow){
 									self.toggleRow(toggledRow)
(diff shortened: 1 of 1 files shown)


olifolkerd:
looks simple enough, good work!


Thanks :)

PR #2226, opened 2019-07-22
fix "realtive" typo in progress formatter

tentus:
This should solve the Firefox-specific bug illustrated in https://jsfiddle.net/baf14q28/1/ where the progress bar does not render correctly until the browser repaints.


olifolkerd:
Good Catch,

I will include this in the next patch release

Cheers

Oli :)

Issue #2229, opened 2019-07-23
Provide a visual selection feedback


angeliski:
**Is your feature request related to a problem? Please describe.**
The current select row module is nice, but I would like to have a visual feedback for the select function, maybe provided by a first column.
I think that is good to provide a better UX, the [Usability Heuristics](https://www.nngroup.com/articles/ten-usability-heuristics/) have a rule called **Visibility of system status**. When I first see the table I don't know if the table is selectable or not.

**Describe the solution you'd like**
The Tabulator provide a column feedback for that. Included a Select All Header 

Ex:
![image](https://user-images.githubusercontent.com/1574240/61672822-8581fe80-acc3-11e9-9f29-77ae31eb66ec.png)

**Describe alternatives you've considered**
I can build that behavior external of Tabulator core, but is very complex control that without using Tabulator core (using just callbacks and properties). And I belive is better to Tabulator have that feature.

**Additional context**
Maybe the issue #2208 is related

olifolkerd:
Hey @angeliski 

This is already on the roadmap for the 4.5 release.

Cheers

Oli :)

angeliski:
Nice @olifolkerd I can help? Maybe you can draft your design code ideia and I can help to put that in 4.4 release, would be nice :)

olifolkerd:
Hey @angeliski 

Thanks for the offer, that would be great!

So this will involve tinkering around in Tabulators internal module structure, so what i would like to do is work through it with you piece by piece so i can introduce you to how it all works internally.

### Making a Pull Request
Here is a top level set of rules for pull requests:

-Your changes should consider all users, not just your specific usage case, I will not accept changes that reduce functionality in one area just to fix a different usage case

-Internal functionality changes should call internal functions, the external API functions are for end users

-Please try and match Tabulators current coding style

-Changes must be made to files in the src directory, the dist files are auto generated and changes there will be automatically overwritten on the next commit

-querySelectors are banned, tabulator works by storing references to elements and passing them arround, querySelectors are very inefficient and will result in automatic rejection of the pull request.


### The Tasks
So now thats out of the way, how to build the row selection features.

so there are three bits of functionality you will need to add:
- Some updates to the select_row module to accommodate the new functionality
- A formatter to show the tickbox
- A formatter to handle the header tickbox

[...]

PR #2241, opened 2019-07-26
[WIP] - Draft to select row input feature 

angeliski:
Hey @olifolkerd 

That is the PR for #2229 
I made a more simple possible. Now we can talk about the solution. You can see in the bellow the visual result.
tabulator-checkbox

See you

### DIFF
--- a/src/js/modules/format.js
+++ b/src/js/modules/format.js
@@ -570,6 +570,32 @@ Format.prototype.formatters = {
 
 		return el;
 	},
+
+	//TODO copy from tickcross
+	rowSelection:function(cell, formatterParams, onRendered){
+		var row = cell.getRow(),
+		value = row.isSelected(),
+		element = cell.getElement(),
+		empty = formatterParams.allowEmpty,
+		truthy = formatterParams.allowTruthy,
+		tick = typeof formatterParams.tickElement !== "undefined" ? formatterParams.tickElement : '<svg enable-background="new 0 0 24 24" height="14" width="14" viewBox="0 0 24 24" xml:space="preserve" ><path fill="#2DC214" clip-rule="evenodd" d="M21.652,3.211c-0.293-0.295-0.77-0.295-1.061,0L9.41,14.34  c-0.293,0.297-0.771,0.297-1.062,0L3.449,9.351C3.304,9.203,3.114,9.13,2.923,9.129C2.73,9.128,2.534,9.201,2.387,9.351  l-2.165,1.946C0.078,11.445,0,11.63,0,11.823c0,0.194,0.078,0.397,0.223,0.544l4.94,5.184c0.292,0.296,0.771,0.776,1.062,1.07  l2.124,2.141c0.292,0.293,0.769,0.293,1.062,0l14.366-14.34c0.293-0.294,0.293-0.777,0-1.071L21.652,3.211z" fill-rule="evenodd"/></svg>',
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey, great first start.

So a bit of feedback.

 You will need to handle clicking on the input to toggle row selection
because a lot of users turn of selection by clicking on the row as a whole.
You will just need to stop the click propagating from the input to the row
to prevent it from immediately untoggling itself.

In the select row module where you are changing the checked state of the
element you will need to check it has been set, other wise it will break
row selection without the formatter.

Looking good so far.

Cheers

Oli

olifolkerd:
Hey, great first start!

So a bit of feedback.

You will need to handle clicking on the input to toggle row selection because a lot of users turn of selection by clicking on the row as a whole. You will just need to stop the click propagating from the input to the row to prevent it from immediately untoggling itself.

In the select row module where you are changing the checked state of the element you will need to check it has been set, other wise it will break row selection without the formatter.

in your formatter when you are calling ***registerRowSelectChecbox*** *(which there is a typo in :p, it should be check not chec)* you should wrap the line in the following if statement, which checks to see if the selectRow module is installed before running the function, and warns the user if it isnt. this is needed because there are slimmed down ways to use Tabulator without all the modules installed, so all calls to module functions are guarded this way
```js
if(this.modExists("selectRow", true)){
    ///call function on the module
}
```
Looking good so far.

Cheers

Oli :)

angeliski:
Fixed @olifolkerd 

Let me know any problem and thanks for your feedback

olifolkerd:
Hey @angeliski 

Looking good so far, 

I like the aria tagging in the formatter, great idea, but it isnt actually needed if you are using an input of type checkbox, it is only needed to tell a screen render that if you are using non standard tags. but this might come in useful for your svg formatter idea, which we can come onto shortly

So onto the fun bit, the title formatter!

so the ***titleFormatter*** param takes the same formatter as the ***formatter*** param. so we are going to use our same formatter to do two different jobs depending on where it is used.

title formatters work by passing in a pusdo cell component that mimics a normal cell component to provide access to the header element, but importantly it is lacking a ***getRow*** function. so in your formatter, if you use an in statement on the `typeof cell.getRow == 'function'` you can then put the cell formatting code in that and the header formatting code in the else.

so in this case we want to create another function in the selection module, lets call it ***registerHeaderSelectCheckbox***, and pass in the checkbox element. we then want to store this on a variable set in the module, lets call it ***headerCheckboxElement***.

Then in the ***_rowSelectionChanged*** function if the ***headerCheckboxElement*** is set we want to check that the number of rows in the ***selectedRows*** array is equal to the number of rows on the `this.table.rowManager.rows` array, if it is equal the check box should be checked, if it is empty the tickbox should be unticked, and if it is greater than zero but less than the number of rows then it should be set to an indeterminate state with `headerCheckboxElement indeterminate = true`.

You will also need to decide what happens when a user clicks on a checkbox that is in an indeterminate state.

Let me know if that all makes sense.

Cheers

Oli :)

angeliski:
Nice @olifolkerd 
I need spend some time in other urgents issues, but I will retake in monday, ok?

See you

olifolkerd:
All good, I'm looking to release in a couple of weekends time, so as long as we can get it finished by next Friday it will make the 4.4 release

angeliski:
Hey @olifolkerd I make the header works

Let me know any problem

See you

olifolkerd:
Hey @angeliski 

Great work, i think we are almost there, i look forward to being able to include it in this sundays release.

So a couple of tweaks remain:

- in the formatter, you probably want to change the event listener from ***click*** to ***change*** to handle scenarios where the user has tabbed to the tickbox and changed it with an enter or space key

- in your _rowSelectionChanged logic if you change the order of the if statements you can avoid setting the indeterminate property twice. make your if else the if and remove the first if statement, then take the logic from the first if statement and put it inside the else to set the indeterminate true or false.

Then i think we should be all there! great work!

Once those changes are in i will merge it into the 4.4 branch. Once that is released we can work through adding the SVG option if you like.

Cheers

Oli :)

angeliski:
Hey @olifolkerd you  can move the release to monday? I stay in a trip away from my computer and will return soon to fix that comments

olifolkerd:
Hey @angeliski 

No chance im afraid, i have a busy week ahead so sunday afternoon is the latest i can push it.

I am happy to merge things in as they stand and make the last couple of updates myself, then leave it up to you to put in the next set of updates.

Cheers

Oli :)

angeliski:
OK @olifolkerd 
We keep improve that in other releases.
Thank you for help

PR #2257, opened 2019-07-31
Fixed Console typos

dota2Pro:
Fixed Console typos

### DIFF
--- a/src/js/column.js
+++ b/src/js/column.js
@@ -140,7 +140,7 @@ var Column = function(def, parent){
 	this.tooltip = false; //hold column tooltip
 	this.hozAlign = ""; //horizontal text alignment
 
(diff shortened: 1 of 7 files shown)


olifolkerd:
Hey @dota2Pro 

Thanks for your feedback i generally agree with your updates but i have a couple of ammends before i can accept the PR:

- in clipboard.js you have fixed a typo and introduced a new one with the double n in warning "WARNNING"
- please do not remove the break statements from any case statements, i include them in all case statements as a matter of good practice to clearly indicate that the statement is not setup for fall through usage


If you can make those changes i will be happy to accept the PR

Cheers

Oli :)

dota2Pro:
@olifolkerd done :)

Issue #2258, opened 2019-08-01
ReferenceError in RowManager.prototype.redraw when global `self` is not defined


ttomii:
ReferenceError is occured in RowManager.prototype.redraw when global variable `self` is not defined.

https://github.com/olifolkerd/tabulator/blob/90146917a84e017402b1eb016a51497d1f986a72/src/js/row_manager.js#L1652-L1663

**Tabulator Info**
4.3.0

**Working Example**

This code is executing on Node.js.

```js
const jsdom = require("jsdom");
const Tabulator = require('tabulator-tables');

const dom = new jsdom.JSDOM(`<html>
<body>
<div id="example-table"></div>
</body>
</html>`);

const window = global.window = dom.window;
global.Event = window.Event;
global.document = window.document;
global.navigator = window.navigator;

const table = new Tabulator("#example-table", {
    height: 205,
    data: [
        {id: 1, name: "Oli Bob", age: "12", col: "red", dob: ""},
    ],
    layout: "fitColumns",
    columns: [
        {title: "Name", field: "name", width: 150},
        {title: "Age", field: "age", align: "left", formatter: "progress"},
        {title: "Favourite Color", field: "col"},
        {title: "Date Of Birth", field: "dob", sorter: "date", align: "center"},
    ]
});

table.redraw(false);
// ReferenceError: self is not defined
```

aklaver:
Can you set up a jsFiddle/CodePen example?

If you do:
```
table.redraw(true);
```
does it work?

ttomii:
> Can you set up a jsFiddle/CodePen example?

It does not reproduce because `window.self` is existing in browser.
In `RowManager.prototype.redraw` , `self` refers `window.self`.
I think it is necessary to replace `self` to `this` or define `self = this` .

> If you do:
> 
> ```
> table.redraw(true);
> ```
> 
> does it work?

Yes, it works.

aklaver:
The example is still going to run in a browser so I am not sure why window.self would not exist.
I do not see this behavior in a non-Node.js case. My suspicion is that this is an out of sequence problem where the redraw() is happening before the table is completely built.

ttomii:
In a browser, `self.renderMode` is `window.renderMode`, and `self.table.options.groupBy` is `window.table.options.groupBy` . I don't think it is correct.

Another example, TypeError occurred when `window.renderMode = "classic"` .
https://jsfiddle.net/un5dxvtz/2/

aklaver:
[...]

PR #2278, opened 2019-08-07
Issue  #2258. ReferenceError in RowManager.prototype.redraw when global `self` is not defined.

aklaver:
Issue #2258 
Add self = this in redraw() when in 'classic' renderMode.

### DIFF
--- a/src/js/row_manager.js
+++ b/src/js/row_manager.js
@@ -1648,11 +1648,10 @@ RowManager.prototype.redraw = function (force){
 	this.table.tableWidth = this.table.element.clientWidth;
 
 	if(!force){
-                self = this;
-		if(self.renderMode == "classic"){
+		if(this.renderMode == "classic"){
 
-			if(self.table.options.groupBy){
-				self.refreshActiveData("group", false, false);
+			if(this.table.options.groupBy){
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hi @aklaver 

Thanks for the input, i like the idea for the test, but could i ask that you tweak your implementation a bit:

- could you move your test function onto the prototype to be consistent with other function calls
- could you set the property "_tabulator_test" instead of just "test" as it will reduce the chance of accidentally overwriting something outside of the scope of Tabulator.
- when calling functions on the window object please explicitly use `window.` as it makes it clear what the scope should be

If you can make those changes i would be happy to accept the PR.

Cheers

Oli :)

aklaver:
I'm guessing you are referring to my other PR(https://github.com/olifolkerd/tabulator/pull/2280). I will look at the above and make the changes.

olifolkerd:
My Bad, i had too many tabs open lol

olifolkerd:
There is no need to set the self variable in this function, just rename any uses of self in the function to ***this***

It will have been caused by a copy and paste error from inside a callback written before i started introducing arrow functions into the code base

Cheers

Oli

aklaver:
Got it.

Issue #2271, opened 2019-08-06
"persistenceMode:true" blocking error "TypeError: localStorage is null"


G-WAN:
**Describe the bug**
Tabulator.js does not check the case of a null localStorage (which happens if localStorage is disabled by the Internet browser options - here Firefox, 68.0.1, latest version).

**Tabulator Info**
4.3 (latest)
- copy of your construct object (works with any table...):

 tablexxx = new Tabulator("#ex-table",{
 height:"100%",
 virtualDom:true,
 resizableColumns:true,
layout:"fitDataFill",
 selectable:true,
 pagination:"local", 
 paginationSize:14,
 paginationButtonCount:1,
 persistenceID:"tablex",
//persistenceMode:true,    // "cookie" if "local" (storage) unavailable
 persistenceMode:"cookie", // true fails: "TypeError: localStorage is null"
 persistentLayout:true,
 });

aklaver:
Hmm:
```
window.localStorage
null
mode
"cookie"
m = mode = mode!== true ?  mode : (typeof window.localStorage !== 'undefined' ? "local" : "cookie");
"cookie"
```

aklaver:
Aah I see the problem. The code will set the storage to 'local' if persistenceMode is set to true or 'local' even if localStorage is not available. Later when it persists data it will fail with 'TypeError: localStorage is null'.  In FireFox when localStorage is disabled it is set to null. In Chrome localStorage is not actually disabled, it just cannot be written to by outside sites. I don't know what the other browsers do.  The storage test needs to change to deal with this. Not sure how at the moment.

G-WAN:
Maybe something [like](https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser):

```

/**
 * Gets the browser name or returns an empty string if unknown. 
 * This function also caches the result to provide for any 
 * future calls this function has.
 *
 * @returns {string}
 */
var browser = function() {
    // Return cached result if avalible, else get result then cache it.
    if (browser.prototype._cachedResult)
        return browser.prototype._cachedResult;

    // Opera 8.0+
    var isOpera = (!!window.opr && !!opr.addons) || !!window.opera || navigator.userAgent.indexOf(' OPR/') >= 0;

    // Firefox 1.0+
    var isFirefox = typeof InstallTrigger !== 'undefined';

    // Safari 3.0+ "[object HTMLElementConstructor]" 
[...]

PR #2280, opened 2019-08-07
Issue #2271. persistenceMode:true. 

aklaver:
Modify persistence.js to check for presence of localStorage. Previous code had this:
this.mode = mode !== true ?  mode : (typeof window.localStorage !== 'undefined' ? "local" : "cookie");
This did not do correct thing in FireFox when localStorage is disabled
as localStorage is set to null. window.localStorage !== 'undefined'
would resolve to true and mode 'local' would be selected even though
it was not available. Put in a generic test for localStorage
availability and use result to determine whether to use 'local' or
'cookie'.

### DIFF
--- a/src/js/modules/persistence.js
+++ b/src/js/modules/persistence.js
@@ -5,21 +5,25 @@ var Persistence = function(table){
 	this.persistProps = ["field", "width", "visible"];
 };
 
-//setup parameters
-Persistence.prototype.initialize = function(mode, id){
-	//determine persistent layout storage type
-	function lsTest(){
-            var  _tabulator_test =  "test";
+// Test for whether localStorage is available for use.
+Persistence.prototype.localStorageTest = function() {
+    function lsTest(){
+            var  testKey =  "_tabulator_test";
             try {
-                    window.localStorage.setItem( _tabulator_test, _tabulator_test);
-                    window.localStorage.removeItem( _tabulator_test);
+                    window.localStorage.setItem( testKey, testKey);
+                    window.localStorage.removeItem( testKey );
                     return true;
             } catch(e) {
                     return false;
-        }
-        }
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hi @aklaver

Thanks for the input, i like the idea for the test, but could i ask that you tweak your implementation a bit:

could you move your test function onto the prototype to be consistent with other function calls
could you set the property "_tabulator_test" instead of just "test" as it will reduce the chance of accidentally overwriting something outside of the scope of Tabulator.
when calling functions on the window object please explicitly use window. as it makes it clear what the scope should be
If you can make those changes i would be happy to accept the PR.

Cheers

Oli :)

aklaver:
Started on this and am not sure what you mean by 'onto the prototype'. The code is in Persistence.prototype.initialize which I thought covered that. 
Also not sure what the window object is in this context?

olifolkerd:
Hey @aklaver 

I think you may have misunderstood me there, i meant change the name of the key you are using to store the data, you renamed the variable but kept the key as "test", because local storage is used globally by any code on the site, you dont want to be using a generic word like "test" because if the developer is using that for their own purposes Tabulator would interfere with it which is why i wanted you to use "_tabulator_test" instead. please change the variable back and change the string to "_tabulator_test"

You have also not put the function on the prototype as asked, you should define the fuction on the prototype like this:

```js
Persistence.prototype.localStorageTest = function() {
 //the current code from your lsTest function
}
```

then call it in your function as:
```js
this.mode = mode !== true ?  mode : (this.localStorageTest() ? "local" : "cookie");
```

I hope that makes sense.

Cheers

Oli :)

aklaver:
Argh a stupid on my part vis-a-vis the key. 
Got it on the prototype code. Will need to study Javascript prototype more!

PR #2365, opened 2019-09-16
_clearSelection performance issue

richard-slabbert94:
Clearing text selection was slowing down the ordinary click row selection event on a large dataset in click selectableRangeMode. The default window text selection need only be cleared during a shift select.


olifolkerd:
Hey @richard-slabbert94 

Thanks for the suggestion, unfortunately i cannot accept this pull request in its current state because you have made it on the dist files. The dist files are programatically generated from the source files, so will be overwritten the moment the next change is made to the library.

Please make your changes to the files in the src directory.

Cheers

Oli :)

olifolkerd:
Hey @richard-slabbert94 

Thanks for the update but you are still making changes to the dist file as well as the src file. there are many dist files generated for different for different use cases. please only make a PR on src files, i will correctly generate the dist files for all merges when i pull them in and do a release version.

Cheers

Oli :)

olifolkerd:
This will be included in the 4.5 release in a couple of weeks

PR #2394, opened 2019-09-29
Add support for sorting editor items on value or label.

wdehoog:


olifolkerd:
Hey @wdehoog,

Thanks for the PR.  I like the idea, but there are a couple of refinements that will be needed before i can approve it.

First point is that you shouldn't be able to sort by both values and labels, there is only one visible list that is displayed to the user, it should therefore be one property ***listSortBy***  that can take a value of "value" or "label" or ***false*** for no sort, which should be its default value.

The second point comes down to considering usage cases outside of your own. Tabulator is used by a great many users each with different needs, each of its functions needs to be open to use in any number of situations. To this end please don't assume that the value in the list is a string *(you are using localeCompare in your sorting)* Tabulator can handle any sort of value.

Tabulators Sort module comes with a a load of built in sorters that should be usable for this, you should have a ***listSorter*** property that you can pass a string representing one of the inbuilt sorters ***(checkout /src/modules/sort.js)*** or a sorter function with a matching list of arguments. You will also need a ***listSorterParams*** property for passing additional params into these sorters. any sorter arguments that aren't relevant to this use can be passed an empty object

Let me know if you have any questions.

Cheers

Oli :)

wdehoog:
@olifolkerd Thanks for your comments. The reason I used both sortValuesList as well as sortLabelsList is that getUniqueColumnValues already uses sortValuesList.
Sorting  by both values and labels is prevented by this:

```
			if (editorParams.sortValuesList)
        			... 
			else if (editorParams.sortLabelsList)
				...
```

But your proposal is much better. I'll see how far I can come.

olifolkerd:
@wdehoog 

Good point about there being an existing ***sortValuesList*** property

I have updated my comment, with your new approach you should replace the the ***sortValuesList*** with a ***listSortDir*** property that takes a value of "asc" or "desc"

Cheers

Oli :)

wdehoog:
@olifolkerd Hopefully I understood what you wished for and how Tabulator is supposed to work. The way the 'dir' parameter is ignored for sorting in the sorters is a bit confusing.

olifolkerd:
Hey @wdehoog 

Thanks for making the updates, i will review them when i get a chance in the next few days.

Cheers

Oli :)

Issue #1936, opened 2019-03-29
Row groupings don't update automatically


takuy:
**Describe the bug**
Not sure if this intentional or not, but when changing a value in a row (where that value is the groupBy value) - the groups don't update unless you manually call setGroupBy(). unfortunately, this reloads the whole table and makes you lose your position. 

I am aware that the drag and drop behavior works fine, but my table has too many groups for that to be effective. 

**Tabulator Info**
4.2 - table config included in codepen below

**Working Example**
https://codepen.io/tako1337/pen/qveoYY

**To Reproduce**

**Expected behavior**
Groups update when row values change.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Chrome 73

olifolkerd:
Hey @takuy 

That will be happening because you have not got a height set on the table, and so the virtual DOM is not being used. without the virtual DOM tabulator has no way of regrouping rows without redrawing the whole table.

If you set a height it should work.

Cheers

Oli :)

takuy:
Hi @olifolkerd, Thanks for the response - I am setting a height on the table in my production code. I forgot to do it on the test codepen - so that has been updated, and I am getting the same results. The groups only update if I call setGroupBy() - and then, it still redraws the whole table.

olifolkerd:
Hey @takuy 

Thanks for the update to the code pen, that certainly looks buggy to me, i will look at including a fix in the next patch release.

Cheers

Oli :)

takuy:
Thanks @olifolkerd! Your work on Tabulator is much appreciated.

takuy:
Btw @olifolkerd can you un-close this? Just so I can get a notification if there's an update. Closed issues don't seem to send notifications. :)

olifolkerd:
@takuy if you update to 4.2.5 do you still have the issue?

takuy:
@olifolkerd Yes - I updated the codepen with 4.2.5 and it's still happening.

olifolkerd:
Hey @takuy 

ok i will investigate further in the next patch release.

Cheers

Oli :)

olifolkerd:
Hey @takuy 

Ahh sorry i follow you now, i did put some fixes into row movement but not for this specific circumstance.

It is actually intended behaviour for it to behave this way at the moment because it could be jaring for the user if they are working there way along a row and it suddenly moves to a new group.

But i do take your point on board, i will look at setting this as an option in the 4.3 release in may.

Thanks for the feeback

[...]

PR #2395, opened 2019-09-30
Adding automatic group updates

takuy:
This is an idea to resolve issue #1936. I am not sure if this is the best place to put this, but it does seem like a catch-all area for both user-based updates and code-based updates. 

Let me know what you think or if you would like any explanation. 

Thanks!

Edit: I guess I should explain:

I took into consideration that you could have multiple levels of groups and that not every update on a value that update the groups. This change will first check the possible new "path" to the bottom-most group, and then builds in-reverse the current path by checking the group parents. It compares the 2 paths and only redraws the table if the paths are not the same. 

I'm not sure if this the best way to redraw the table... so please let me know!

### DIFF
--- a/src/js/cell.js
+++ b/src/js/cell.js
@@ -515,6 +515,10 @@ Cell.prototype.setValue = function(value, mutate){
 			this.column.cellEvents.cellEdited.call(this.table, component);
 		}
 
+		if(this.table.options.groupUpdateOnCellEdit && this.table.options.groupBy && this.table.modExists("groupRows")) {
+			this.table.modules.groupRows.reassignRowToGroup(this.row);
+		}
+		
 		this.cellRendered();
 
 		this.table.options.cellEdited.call(this.table, component);
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -190,6 +190,7 @@ Tabulator.prototype.defaultOptions = {
 	groupBy:false, //enable table grouping and set field to group by
 	groupStartOpen:true, //starting state of group
 	groupValues:false,
+	groupUpdateOnCellEdit:false,
 
 	groupHeader:false, //header generation function
 
--- a/src/js/modules/group_rows.js
+++ b/src/js/modules/group_rows.js
@@ -54,6 +54,9 @@ GroupComponent.prototype.getTable = function(){
 	return this._group.groupManager.table;
 };
 
+GroupComponent.prototype.getPath = function() {
+	return this._group.getPath();
+}
 //////////////////////////////////////////////////
 //////////////// Group Functions /////////////////
 //////////////////////////////////////////////////
@@ -648,6 +651,12 @@ Group.prototype.generateGroupHeaderContents = function(){
 	this.element.insertBefore(this.arrowElement, this.element.firstChild);
 };
 
+Group.prototype.getPath = function(path = []) {
+	path.unshift(this.key);
+	if(this.parent) this.parent.getPath(path);
+
+	return path;
+}
 ////////////// Standard Row Functions //////////////
 
 Group.prototype.getElement = function(){
@@ -1040,7 +1049,31 @@ GroupRows.prototype.assignRowToGroup = function(row, oldGroups){
 	return !newGroupNeeded;
 };
 
+GroupRows.prototype.reassignRowToGroup = function(row){
+	var oldRowGroup = row.getGroup(),
+		oldGroupPath = oldRowGroup.getPath(),
+		newGroupPath = this.getExpectedPath(row),
+		samePath = true;
+	// figure out if new group path is the same as old group path
 
+	var samePath = (oldGroupPath.length == newGroupPath.length) && oldGroupPath.every(function(element, index) {
+		return element === newGroupPath[index]; 
+	});
+	// refresh if they new path and old path aren't the same (aka the row's groupings have changed)
+	if(!samePath) {
+		oldRowGroup.removeRow(row);
+		this.assignRowToGroup(row, self.groups);
+		this.updateGroupRows(true);
+	}
+};
+
+GroupRows.prototype.getExpectedPath = function(row) {
+	var groupPath = [], rowData = row.getData();
+	this.groupIDLookups.forEach(function(groupId) {
+		groupPath.push(groupId.func(rowData));
+	});
+	return groupPath;
+}
 
 GroupRows.prototype.updateGroupRows = function(force){
 	var self = this,
--- a/src/js/modules/group_rows.js
+++ b/src/js/modules/group_rows.js
@@ -54,9 +54,6 @@ GroupComponent.prototype.getTable = function(){
 	return this._group.groupManager.table;
 };
 
-GroupComponent.prototype.getPath = function() {
-	return this._group.getPath();
-}
 //////////////////////////////////////////////////
 //////////////// Group Functions /////////////////
 //////////////////////////////////////////////////
@@ -653,8 +650,9 @@ Group.prototype.generateGroupHeaderContents = function(){
 
 Group.prototype.getPath = function(path = []) {
 	path.unshift(this.key);
-	if(this.parent) this.parent.getPath(path);
-
+	if(this.parent) {
+		this.parent.getPath(path);
+	}
 	return path;
 }
 ////////////// Standard Row Functions //////////////
@@ -1055,7 +1053,6 @@ GroupRows.prototype.reassignRowToGroup = function(row){
 		newGroupPath = this.getExpectedPath(row),
 		samePath = true;
 	// figure out if new group path is the same as old group path
-
 	var samePath = (oldGroupPath.length == newGroupPath.length) && oldGroupPath.every(function(element, index) {
 		return element === newGroupPath[index]; 
 	});
@@ -1063,7 +1060,7 @@ GroupRows.prototype.reassignRowToGroup = function(row){
 	if(!samePath) {
 		oldRowGroup.removeRow(row);
 		this.assignRowToGroup(row, self.groups);
-		this.updateGroupRows(true);
+		this.table.rowManager.refreshActiveData("group", false, true);
 	}
 };
 
--- a/src/js/row.js
+++ b/src/js/row.js
@@ -571,6 +571,10 @@ Row.prototype.updateData = function(updatedData){
 				}
 			});
 		}
+		
+		if(this.table.options.groupUpdateOnCellEdit && this.table.options.groupBy && this.table.modExists("groupRows")) {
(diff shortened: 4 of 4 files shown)


olifolkerd:
Hey @takuy 

Thanks for the pull request, it is great to see people interested in contributing to Tabulator

i like the general idea, but there are a few points that need to be addressed before i can accept the pull request.

The fist point is that in the initial issue you raised i mentioned that this would need to be an optional behaviour it could not be the default behaviour. could you please add an option of ***groupUpdateOnCellEdit*** that defaults to false, this can be added in the ***defaultOptions*** object in core.js

You don't need 99% of the code you have added there, the group module already has an ***assignRowToGroup*** function that deals with assigning the row to the correct group so you dont have to, you should use the existing functionality rather than trying to recreate some of the modules functionality outside of the module. *(as a general rule for future PR's please dont try and replicate module functionality outside of the module)*

I would suggest that you create a ***reassignRowToGroupFunction*** that is based on the ***assignRowToGroup*** function, it should lookup the groupID as per ***assignRowToGroup*** and check that the group has changed. if it has add it to the new group and then trigger the redraw.

You also need to consider how the ***setValueProcessData*** function is called. it is used when users edit a cell, but also when uses call the ***row.update*** function. and in this case your function would be called several times for the same data which would be inefficient. i would recommend therefor that you move calling your new ***reassignRowToGroupFunction*** function to two different places, which cover the two different useage cases, the first is in the ***setValue*** function on the cell, inside the *if(changed)* statement before the callbacks. The second would be in the ***update*** function on the row before the *if(visible)* statement

Let me know if you have any questions.

Cheers

Oli :)

takuy:
Thanks @olifolkerd. I'll start making the requested changes. I did see `assignRowToGroup `when I was looking through before, but noticed there were 2 identical functions defined as `GroupRows.prototype.assignRowToGroup`.

olifolkerd:
Yup, I just pushed a fix for that to the master branch as I noticed the
duplication too when I was looking into this PR

takuy:
@olifolkerd I made the request changes, BUT, I could not figure out what method there already was to figure out if the group has changed. Part of the issue is that you can have nested groups, that share the same names at the lowest level 

ie, you can have nested groups like "America -> New Jersey -> Manager" and "America -> Pennsylvania -> Manager". These are two different "Manager" groups with the same name. Most of the functionality contained in "GroupRows" seems to deal with groups at level 0, so I wasn't able to adapt much code from there. 

So, I re-write part of my "group path" idea into the new function to check the full path of the group. I think it's pretty efficient, but let me know if you think otherwise. 

I am testing locally in a CodePen with this setup, which has 3 tiers of groups along with a "complex" group: https://codepen.io/tako1337/pen/xxKNMZG - editing the number column shouldn't change the group, but changes first/last name will.

takuy:
Hey @olifolkerd Just wanted to know if you can look at this when you get a chance. I'd really like to get this feature in as it would be very beneficial for row groupings. Let me know if there's any other changes you think I should make.

olifolkerd:
I really appreciate your enthusiasm on this one. But I do ask people to respect that I have a life outside Tabulator so can't always get back to people straight away.

I will have a look at it this weekend when I get a chance.

As it is a new feature it would not be merged in straight away. I only add new features in minor version releases to ensure that the code people pull down from the repo matches the documentation. In this case the 4.5 release in a few weeks time.

So there is plenty of time to work towards a solution.

Cheers

Oli :)

AndrewHutcheson:
Was this merged with 4.5 or is it for the next version?

I set the height manually but changing a field that is a grouping field causes the window to jump up whether I use redraw or update just that row.

olifolkerd:
It is still open, so it hasn't been merged in yet

takuy:
@olifolkerd Hope all is well. Please let me know if you're still interested in merging this at some point.

It would still really be useful in my own cases. I'd be glad to make any changes that you see fit.

olifolkerd:
Hey @takuy 

Sorry for the delay in getting back to you, my father died at the end of last year which made it hard to function the first few months of this year *(sorry for the sharp reply in January)*, then the pandemic stuck, so I have been spread a bit thin.

There are a few issues i would like to go over with the current approach in the group_rows.js file:
- on line 1048 you are using `this.table.modules.groupRows.groupIDLookups`, you could just use `this.groupIDLookups`
- on line 1049 you are making repeated calls to `row.getData()` it would be more efficient to load this into a variable outside of the for loop and then reference the variable inside the loop. You could in fact abstract this whole array generation to its own function that returns the array as i can see it being useful in future.
- on line 1053 you declare a variable, please make sure that all variables are declared at the top of functions like yu have with oldRowGroup and groupPath
- in your for loop  on line 1054 you use a `var`, please use a `let`declaration for block level variables
- where you determine if they are on the same path on line 1054  i would prefer that you create a `getPath` function on the group class that then iterates back over its parents calling the same function to return an array of path levels *(essentially the function should return an array of the parents group paths and push its own on the end, so the function start by calling the function on its partent, this propagates all the way to the top level group which returns an array containing its path value, which the child pushes its value  on and returns and so on until the whole array is returned)*, that way it is up to the groups to determine their current path. You could then directly compare the two arrays to see if they are the same
- at the moment you are redrawing the whole table when moving a single row, which is quite inefficient each group has a ***removeRow*** and ***addRow*** function that should be used for moving a row between groups, which is how it is handled when manually dragging and dropping rows as it is more efficient. In this case i would suggest using the ***removeRow*** function on the current group and then using the ***assignRowToGroup*** function on the GroupRows class that will find the group that a row should be in and assign it there. 

I hope that is useful.

Im looking to get the 4.7 release out next weekend, do you cany including it in that?

Cheers

Oli :)

takuy:
Sorry about the messy PR history above, tried cleaning up the commit history since I had to rebase to the latest master (and then kept finding things that should have been fixed).

> Sorry for the delay in getting back to you, my father died at the end of last year which made it hard to function the first few months of this year _(sorry for the sharp reply in January)_, then the pandemic stuck, so I have been spread a bit thin.

No problem at all! Very understandable considering the circumstances. 

> * on line 1048 you are using `this.table.modules.groupRows.groupIDLookups`, you could just use `this.groupIDLookups`
> * on line 1053 you declare a variable, please make sure that all variables are declared at the top of functions like yu have with oldRowGroup and groupPath
> * in your for loop  on line 1054 you use a `var`, please use a `let`declaration for block level variables

All fixed!

> * on line 1049 you are making repeated calls to `row.getData()` it would be more efficient to load this into a variable outside of the for loop and then reference the variable inside the loop. You could in fact abstract this whole array generation to its own function that returns the array as i can see it being useful in future.

I've moved the array generation to a new `getExpectedPath` function on the `GroupRows` object, which takes a row as an argument. Not sure if this is the best place for it, but since it independent of `Group`s themselves, I feel like this is a pretty okay spot. 

> * where you determine if they are on the same path on line 1054  i would prefer that you create a `getPath` function on the group class that then iterates back over its parents calling the same function to return an array of path levels _(essentially the function should return an array of the parents group paths and push its own on the end, so the function start by calling the function on its partent, this propagates all the way to the top level group which returns an array containing its path value, which the child pushes its value  on and returns and so on until the whole array is returned)_, that way it is up to the groups to determine their current path. You could then directly compare the two arrays to see if they are the same

Fixed! This was my original approach when I submitted the PR, I've re-implemented & I've abstracted it out to a `getPath` function for `Group`s and can also be called from a `GroupComponent`. 

> * at the moment you are redrawing the whole table when moving a single row, which is quite inefficient each group has a _**removeRow**_ and _**addRow**_ function that should be used for moving a row between groups, which is how it is handled when manually dragging and dropping rows as it is more efficient. In this case i would suggest using the _**removeRow**_ function on the current group and then using the _**assignRowToGroup**_ function on the GroupRows class that will find the group that a row should be in and assign it there.

I've implemented these changes but found that `assignRowToGroup` does not draw the row once it has been assigned to a new group. I'm currently calling `updateGroupRows`, as consistent with other  GroupRow functions. Any suggestions for this would be appreciated. 

> Im looking to get the 4.7 release out next weekend, would you like your PR including it in that?

That would be great! As long as it meets all of your standards.

olifolkerd:
Hey @takuy 

Thanks for making those changes, on the whole that is looking great.

Can i ask why you have added the ***getPath*** function to the GroupComponent as well as the Group? as it is for internal use only it should not be available on the component object.

With the ***getPath*** function, can you avoid using single line if statements, Tabulators coding style always uses brackets on if statements to keep things clear.

For now lets stick with using the ***refreshActiveData*** function to refresh the rows as you had before, it is a bit inefficient but at this point i would rather have the functionality in than not at all:
```js
this.table.rowManager.refreshActiveData("group", true, true);
```

Cheers

Oli :)

8483:
I am really interested in this change, but I don't quite get how it works, and I couldn't find the documentation.  

How do we use this?
```js
this.table.rowManager.refreshActiveData("group", true, true);
```

Example code:
```js
cellClick: (e, cell) => {
    let data = cell._cell.row.data;
    let row = cell._cell.row;
    let sku = data.sku;
    showEdit(sku, updateRowCallback);

    function updateRowCallback() {
        fetch(`/api/categorization/${sku}`)
            .then((res) => res.json())
            .then((data) => {
                table.updateRow(sku, data);
            })
        .catch((err) => showError(err));
    }
},
```

olifolkerd:
Hey @8483 

You should ***NEVER*** try and access any properties on the ***rowManager*** object, that is for internal table use ONLY, any attempt to fiddle with that from outside the table could cause significant damage to the table.

There wont be any documentation on this feature yet as the pull request is not yet merged in and is therefore not available for usage When it has been merged in and released the documentation will be updated to reflect its availability.

The point of this pull request is to automatically change the data of the row data when the group is moved without you needing to add any additional code on your end.

Please restrict future comments on pull requests to input reguarding the submitted code.

Cheers

Oli :)

8483:
I apologize.  

I guess I should submit a separate feature request, as your comment made me realize it's a different use case.

olifolkerd:
Hey @8483 

I have updated the description, u used the word "moved" when i meant "updated"

The feature should be coming in the next release, but it is not ready yet

Cheers

Oli :)

8483:
Awesome!

Sorry for posting the feature request. I'll refrain from further comments and wait for the next release.

olifolkerd:
Hey @takuy 

Is this ready for release?

Cheers

Oli :)

takuy:
I think so. My last commit should have addressed the last mentioned issues.

olifolkerd:
@takuy 

I will get this merged in for the 4.8 release next weekend.

Cheers

Oli :)

PR #2406, opened 2019-10-04
Performance optimization for accessor.transformRow

ryanlin1986:


olifolkerd:
Hey @ryanlin1986 

That pull request wont work, you have moved the if statement out of the for loop where the column variable is defined.

The whole point of that for loop is to iterate over columns, therefore the column variable is different in each itteration.

Oli :)

PR #2413, opened 2019-10-07
Bulk select active rows only

richard-slabbert94:
Hey Oli,

Here I have add a new title format param (activeOnly) to the rowselection format which only bulk-selects currently active rows.  Maybe you could find it useful?

Richard


richard-slabbert94:
Found a bug in my code.  I will create another pull request once it is sorted.

Issue #2054, opened 2019-05-11
CI enviroment to tests


angeliski:
The main proposal for that issue is talk about a enviroment to run tests.
Please share your opnion about the topics:
1 - CI enviroment. Ex: Travis, CircleCI
2 - Testing Framework Ex: Jest, Mocha
3 - Critical features to test. Where we need start?

olifolkerd:
Hey @angeliski 

Thanks for raising this, do you have any personal recomendations from your experience?

Cheers

Oli :)

angeliski:
Yes, for CI I prefer Travis CI, but I don't have a lot experience in CircleCI (just run one project in that).
A prefer Jest because is really simple run and configure.
You is the best man to say where we can start, but I can think in the core e column_manager.
What do you think?

olifolkerd:
Hey @angeliski 

Im happy for you to run with whichever tool you think best,

The column manager and row managers seem like sensible places to start

Cheers

Oli :)

olifolkerd:
Hey @angeliski 

I am going to create a new issue to track this now, closer to the 5.0 release notice, The codebase has been refactored to make unit testing easier so i would really like to encourage the community to implement this now.

Cheers

Oli :)

PR #2420, opened 2019-10-10
Draft unit test

angeliski:
Hey @olifolkerd 
That is a first draft to #2054

Can you provide feedbacks?
You can see the travis build here: https://travis-ci.org/angeliski/tabulator/builds/596233421

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+anguage: node_js
+node_js:
+  - v8.10.0
+cache:
+  npm: true
+  directories:
+    - ~/.cache
+script:
+  - npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,9 @@ cache:
   npm: true
   directories:
     - ~/.cache
+before_script:
+  - npm install -g gulp-cli    
 script:
+  - npm install
+  - gulp
   - npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-anguage: node_js
+language: node_js
 node_js:
   - v8.10.0
 cache:
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,182 +1,25 @@
-// For a detailed explanation regarding each configuration property, visit:
-// https://jestjs.io/docs/en/configuration.html
-
 module.exports = {
-  // All imported modules in your tests should be mocked automatically
-  // automock: false,
-
-  // Stop running tests after `n` failures
-  // bail: 0,
-
-  // Respect "browser" field in package.json when resolving modules
-  // browser: false,
-
-  // The directory where Jest should store its cached dependency information
-  // cacheDirectory: "/tmp/jest_rs",
-
-  // Automatically clear mock calls and instances between every test
   clearMocks: true,
-
-  // Indicates whether the coverage information should be collected while executing the test
-  // collectCoverage: false,
-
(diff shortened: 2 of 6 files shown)


olifolkerd:
Hey @angeliski 

Very Exciting!

Looking food so far, it seems to have some basic tests in place, so next up i guess we need to define a series of test cases to be included?

Do you have an recommendations?

Cheers

Oli :)

angeliski:
Hey @olifolkerd 
I put only a simple test case, just to run the pipeline.
The better way is make the test run a file per module, we can think about the test cases to run in priority and develop the others on demand.

angeliski:
@olifolkerd A question, nowadays you use the gulp to inject the files, like ```/*=include column_manager.js */```

You see any problem if we translate that to CommonJS(using module.exports) and using something like webpack/browserify to make the build?

My question is just because the current files don't have the module.exports, so we can't require that in the test file.

That will fail:
```
const ColumnManager = require('../src/js/column_manager');
```
What do you think?

olifolkerd:
It is done that way because the there are different combinations of files pulled in for the different modules.

I'm not sure why you would need to change it. The modules can't be run independently they have to be included with the table so surely the tests should run on the compiled dist files.

angeliski:
Hey @olifolkerd Sorry the late response, I was working in another thing

I was understanded the modules can't be run independently, but when you think about unit tests, the expect is module's test  in a isolated enviroment. We can add a integration layer to  catch any error on use the modules inside the tabulator, but is nice test the module in a unit test first. Does it make sense?

I proposed that change because we need a way to test each module, like module.exports or maybe export default.

olifolkerd:
Hey @angeliski 

Thanks for the additional information.

Im not against moving to this, but there is a lot that needs to be through through first. there are many versions of dist JS files build from the src files. not just the minified files. For example there is the option to use the core.js module and then install each module individually on your site.

In that instance the require syntax would be invalid as it could not be handled in browser an on IE. so if this is the way we need to move then we need to consider the whole packaging process further.

more information on the packaging options can be found [Here](http://tabulator.info/docs/4.4/install)

Cheers

Oli :)

angeliski:
Hey @olifolkerd 
Make sense, if we choose to move to CommonJS we need to build another output, using something like  webpack or browserify.

Can I work in a draft for that?

rotht2:
Hello, we are also interested in unit tests with Tabulator. Are there any updates on this? I recently spiked getting Jest working and I ran into the same issues that @angeliski had.

olifolkerd:
@angeliski Sorry for the delay in getting back to you, my father passed away at the end of November so i haven't been terribly focused on things the last month.

I can't see why you would need to move away from gulp to support this. it is a task automation tool, it will do what you ask it too.

in version 4.6 we are moving to gulp4 which provides even better support.

I would be reluctant to move to another automation tool unless you can provide some very clear advantages as to why this would be needed.

Cheers

Oli :)

jjech:
@olifolkerd really sorry to hear about your father's passing.

One benefit to moving away from gulp is this would eliminate the dependency on the deprecated gulp-cssnano package, which has broken the build process as explained in issue #2581.

But we can use npm shrinkwrap as shown in PR #2582 or the manual workaround to continue building with gulp at the moment, so it doesn't really seem to be much of a problem.

olifolkerd:
Hey,

Are you still interested in making some unit tests now that version 5.0 is most ready?

Oli

PR #2427, opened 2019-10-14
Fixed missing highlight color (for selected rows) in the new materialize design

bertoverflow:
Selecting a row would not change the color. I added the missing snippets from the materialize-css so it works now correctly.


bertoverflow:
closed, because I will open a new PR without changes to dist

PR #2444, opened 2019-10-21
Fixed missing highlight color (for selected rows) in the new materialize design

bertoverflow:
Selecting a row would not change the color. I added the missing snippets from the materialize-css so it works correctly.

### DIFF
--- a/src/scss/materialize/variables.scss
+++ b/src/scss/materialize/variables.scss
@@ -46,6 +46,23 @@ $materialize-red: (
   "darken-4":   #8b1014,
 );
 
+$colors: (
+        "materialize-red": $materialize-red,
+) !default;
+
+// usage: color("name_of_color", "type_of_color")
+// to avoid to repeating map-get($colors, ...)
+@function color($color, $type) {
+  @if map-has-key($colors, $color) {
+    $curr_color: map-get($colors, $color);
+    @if map-has-key($curr_color, $type) {
+      @return map-get($curr_color, $type);
+    }
+  }
+  @warn "Unknown `#{$color}` - `#{$type}` in $colors.";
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey,

Thanks for the PR, 

could you go into a bit more detail on the effect of these changes as you don't appear to have altered anything actually used in the tabulator_materialize.scss

This theme should work independently of having any materialize stylesheets in place

Cheers

Oli :)

bertoverflow:
Hi Oli,

surely I can go into a bit more detail. To see the actual changes of my commit you have to build the project so you get a updated `dist/css/materialize/tabulator_materialize.css`.

In your current 4.5-branch this file contains a lot of invalid css-color definitions like this:
`color("materialize-red", "lighten-2");`

For example the css-selector for a selected tabulator-row:
```
.tabulator-row.tabulator-selected {
  background-color: color("materialize-red", "lighten-2");
}
```
This invalid color-definition was the cause for the missing color-change for selected rows (which is how I noticed the problem in the first place).

These css-color-definitions stem from the scss-variable `$primary-color` in `src/scss/materialize/variables.scss`:
`$primary-color: color("materialize-red", "lighten-2") !default;`

I looked in the original materialize-css-file (https://github.com/Dogfalo/materialize/blob/master/sass/components/_color.scss) and noticed they defined a `color`-function (which in turn works with a color-array) and I copied the corresponding parts to make the scss-file work.

Of course there are other possibilities to tackle this problem, I just thought you might want to adhere to the way materialize is doing this.

olifolkerd:
Thanks for the explanation, that was just what i needed.

This will be included in the 4.5 release.

Cheers

Oli :)

PR #2445, opened 2019-10-21
Improved sorting performance when sorting on multiple columns

bertoverflow:
The current implementation for sorting on multiple columns seems to be unnecessary slow. It will call the sort-Function on the data-array once for each sorter (i.e. item in the sortList).
However, only one (!) call is actually needed: the last one.

There are two reasons for this:
* a) the last call to sort will "overwrite" the previous sorting anyways
* b) the comparator used in the actual sorting-algorithm falls back to the previous columns in case the current comparison result was 0 (=equal) -> this already fully handles the multi-sort

### DIFF
--- a/src/js/modules/sort.js
+++ b/src/js/modules/sort.js
@@ -228,6 +228,9 @@ Sort.prototype.sort = function(data){
 
 	if(!self.table.options.ajaxSorting){
 
+		var lastPossibleIndex = -1;
+
+		// iterate over the sortList to find the last possible sorting column (i.e. column that is actually valid for sorting)
 		sortList.forEach(function(item, i){
 
 			if(item.column && item.column.modules.sort){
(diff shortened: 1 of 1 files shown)


olifolkerd:
Nope this is definitly still needed, in order for the equals case to work the column must have been previously sorted by it therefor the previous sorter must have been called first.

bertoverflow:
Hmmm... have you tested my proposed change? Because I tested it locally with my tabulator-table and I got the same sorting result as with the original code? But of course there might be something I am missing here.

To demonstrate my point, I made a minimal example where I am also only calling sort once on the array and it is still correctly sorted: https://jsfiddle.net/58L1eav2/5/

olifolkerd:
Hey @bertoverflow 

On further investigation i do see your point, that that it is essentially resorting the already sorted data in the event of a match.

You are indeed correct about not needing to be calling all the sorts in their full order. 

I have a few other updates to the sort module this weekend, i will include a tweak based on your initial sorter improvement.

Cheers

Oli :)

PR #2456, opened 2019-10-24
adding in the ability to set initial layouts

clurect:
I'd like to see if you want this new feature.

I've made it so that I can pass in a layout as an option. Previously, I was setting the layout and sort after the table was created which was 2-3 times slower than creating the table without sorters or saved layout. The layout needs to be set before the sort. It's much faster and my users will probably enjoy that.

@olifolkerd if you want me to try to add in the persistence like you have for sort and filters I would be happy to add more commits to this PR for it. I thought it would be nice to run this feature idea past you before I just go crazy and add stuff.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -60,6 +60,7 @@ Tabulator.prototype.defaultOptions = {
 	initialSort:false, //initial sorting criteria
 	initialFilter:false, //initial filtering criteria
 	initialHeaderFilter:false, //initial header filtering criteria
+	initialLayout:false, //initial layout criteria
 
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @clurect 

The persistence module has had a complete overhaul on the 4.5 branch that will be released in a few weeks. So im afraid im not accepting any PRs on it.

Also if all you are looking to do is to persist the column order, this happens already in the existing system.

Cheers

Oli :)

clurect:
Ok, thanks for letting me know. I couldn't find a way to set the layout before creation if I saved the layout in my application DB for instance.

Thanks

PR #2504, opened 2019-11-27
Fixing a TypeError that was thrown inside a Promise and was uncaught

arimeq:
Hi,

I am using Tabulator in an internal application developed for large, international bank. We're using it with LitElement/WebComponents and most of the time it works just fine, awesome job ;) Today I noticed error messages in developers console, saying that `Uncaught (in promise)`. So I started hunting. It appears that pagination tries to set a Page when the table is not yet ready. The root cause of the issue was line 487 of `src/js/modules/page.js`:

```javascript
left = this.table.rowManager.scrollLeft;
```

In the code above, `this.table` was `undefined`, which caused an uncaught `TypeError`. I added a check at the beginning of the function and I explicitely throw an `Error`. Why? Because this is inside a `Promise`, so it will be caught in the caller. I think this solution is clean and it resolves my issue.

BTW problem was not visible otherwise; I only noticed it because of a message in Chrome Developer Tools :)

### DIFF
--- a/src/js/modules/page.js
+++ b/src/js/modules/page.js
@@ -474,7 +474,10 @@ Page.prototype.getRows = function(data){
 Page.prototype.trigger = function(){
 	var left;
 
-	return new Promise((resolve, reject)=>{
+	return new Promise((resolve, reject) => {
+		if (!this.table) {
+			throw new Error("we're too early, table not constructed yet!");
+		}
 
 		switch(this.mode){
 			case "local":


olifolkerd:
Hey,

Thanks for the info.

Could you create a js fiddle that demonstrates the original issue,

I don't quite follow when the problem is showing up.

Cheers

Oli

arimeq:
Hey,

Thanks for your comment, it made me dig deeper and it turns out it was my fault. I'm closing this PR :) Have a nice day! :)

cube-access-repo:
HI arimeq, im having the same exact issue here.... could u tell me what was the issue you discovered?

arimeq:
> HI arimeq, im having the same exact issue here.... could u tell me what was the issue you discovered?

I had a race condition in my code. Simply saying, somehow I tried to change page of Tabulator before it was even constructed. This was because I made my own pager for Tabulator and there was some error inside.

cube-access-repo:
Yes. same here. found the bug... it was another functiont thats was called on pagination hiding somewhere.... thanks for your response!

Issue #2524, opened 2019-12-10
Tabulator modifies original data array with undefined values and empty objects


britvik:
**Describe the bug**
Tabulator modifies original data array if a row does not have field/object that is referenced via 'field' in column definition. It creates undefined fields and objects with undefined fields. See codepen example.

**Tabulator Info**
- 4.5.1

**Working Example**
https://codepen.io/britvia/pen/ExaKYqG

**Expected behavior**
I would expect tabulator to NOT modify my data unless I use mutators.

lorpmtp:
I agree,

When column setting property 'field' is set and the intial data property is null, the data returned by events return a wrong formatted object (instead of being set to null, it returns an object with the property.

Column Definition : 
`  
{
        field: 'field1.subField1',
        title: 'field1',
        headerFilter: true
}
`

Initial Object
`
{
        field0: 'something',
        field1: null,
        field2: 'something else'
}
`

Retreived Object
`
{
        field0: 'something',
        field1: { 
                       subField: null
                   },
        field2: 'something else'
}
`

olifolkerd:
I will look into this further over the next couple of weeks, this is definitly a bug.

@lorpmtp proposed fix does fix this issue but unfortunatly it breaks other parts of the system.

please folllow the PR for more info.

Cheers

Oli :)

olifolkerd:
Hey @britvik 

I have pushed a fix for this to the 4.6 branch, which will be released later this weekend.

Cheers

Oli :)

britvik:
Heya @olifolkerd thanks for the new release :).

I tested this issue in new version at https://codepen.io/britvia/pen/ExaKYqG and it seems the nested objects don't get created anymore. But the fields still do.

Original data inserted into table:
`{}`
Result data in v4.5.1:
`{ name: undefined, brother: { name: undefined }, favourite: { movie: { name: undefined } } }`
Result data in v4.6.1:
`{ name: undefined }`
Expected data:
`{}`

PR #2526, opened 2019-12-10
Fix issue #2524 + fix typo setNestedData function

lorpmtp:


olifolkerd:
Hey @lorpmtp 

Thanks for your pull request.

Please can you include a description with your pull request that explains what you are trying to achieve and the method you have taken. please also bear in mind that you need to take into account all uses of a function when making an update.

There is an issue with your current approach, if someone calls the ***setValue*** function on a cell component and for a cell whos current value is undefined then the current version of this function  works exactly as it needs to because it would need to create the series of nested objects to set the value on the cell. your proposed update would break this.

There are a few structural issues with the commit that need to be resolved before i can accept this:
- you have included updates to the css files and map files in your commit that are totally unrelated to your changes. please only include the files you have actually update
- the comparison in the if should be if the type of the variable is undefined not if the variable is null
- the closing bracket on line 773 is not correctly indented 

If you can fix the structural issues in the commit and address my concerns about the setValue issue then i would be happy to proceed with this.

Cheers

Oli :)

olifolkerd:
A simple way of achieving what you are after is to check if the value variable comming into the function is undefined and then handling it differently

lorpmtp:
Hi Oli !

Thx for your return on this.

I effectively need to check whether the type is undefined or null depending the case.

Cheers

olifolkerd:
As I mentioned in my above reply, your current approach is flawed and breaks other parts of the table.

Pull requests must account for all uses of the table and not break the table for others.

I will be unable to accept this PR unless you can fully address the issues outlined in my initial response.

Cheers

Oli

olifolkerd:
Hey @lorpmtp 

Are you interested in resolving the issues or shall i close this PR?

Cheers

Oli :)

lorpmtp:
I plan to fix it this week. Happy new year !

Issue #2530, opened 2019-12-11
'getColumnLayout()' throws a 'Cannot read property 'forEach' of undefined' error when persistence is not configured.


dotariel:
**Describe the bug**
`getColumnLayout()` throws a `Cannot read property 'forEach' of undefined` error when persistence is not configured.

**Tabulator Info**
4.5.2

**Working Example**
Run the below JSFiddle with the developer console window open:
https://jsfiddle.net/5uev8abr/10/

**To Reproduce**
1. Create a Tabulator instance without specifying the `persistence` option
1. Call `getColumnLayout()` on the Tabulator instance

**Expected behavior**
The columnLayout should be returned.

**Desktop (please complete the following information):**
 - OS: macOS 10.15.1
 - Browser: Chrome
 - Version: 78.0.3904.108 (64-bit)

dotariel:
The issue is in `Persistence.prototype.parseColumns` and also in `Persistence.prototype.mergeDefinition`, where they are both performing the following checks to set the column keys:

```
if(self.config.columns === true){
    keys =  Object.keys(colDef);
    keys.push("width");
}else{
    keys = self.config.columns;
}
```

The first check ignores the scenario where there is no persistence configuration. Perhaps a better check would be:

```
if(self.config.columns === true || self.config.columns == undefined){
    keys =  Object.keys(colDef);
    keys.push("width");
}else{
    keys = self.config.columns;
}
```

olifolkerd:
Hey @dotariel 

good spot, that is definitely a bug, i will review your PR over the festive break.

Cheers

Oli :)

dotariel:
Thanks, @olifolkerd.

Have a wonderful break.

olifolkerd:
PR has been merged into the master branch, i will include it in the 4.5.3 patch release.

olifolkerd:
4.5.3 is released now :)

dotariel:
Thanks!

PR #2537, opened 2019-12-14
Check that Persistence.config.columns is not 'undefined'

dotariel:
Fixes issue #2530


olifolkerd:
Looks good, i will merge that in now and include it in the 4.5.3 patch release

PR #2551, opened 2019-12-25
Update build to gulp 4

Jojoshua:
I was going to see if I could help with a bug but started having issues with building. This is to update to gulp v4. My editor also did some autocleanup of ' to " but I thought that was OK since the rest of the code uses ". 

I don't work with gulp a lot but this seems to work as intended.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,14 +2,9 @@ var gulp = require("gulp"),
   sass = require("gulp-sass"),
   autoprefixer = require("gulp-autoprefixer"),
   cssnano = require("gulp-cssnano"),
-  jshint = require("gulp-jshint"),
   uglify = require("gulp-uglify"),
-  imagemin = require("gulp-imagemin"),
   rename = require("gulp-rename"),
   concat = require("gulp-concat"),
-  notify = require("gulp-notify"),
-  cache = require("gulp-cache"),
-  livereload = require("gulp-livereload"),
   del = require("del");
 (include = require("gulp-include")), (sourcemaps = require("gulp-sourcemaps")), (babel = require("gulp-babel")), (plumber = require("gulp-plumber")), (gutil = require("gulp-util")), (insert = require("gulp-insert")), (fs = require("fs"));
 
@@ -213,8 +208,8 @@ gulp.task(
 
 gulp.task("watch", function() {
   // Watch .scss files
-  gulp.watch("src/scss/**/*.scss", ["styles"]);
+  gulp.watch("src/scss/**/*.scss", gulp.series("styles"));
 
   // Watch .js files
-  gulp.watch("src/js/**/*.js", ["scripts"]);
+  gulp.watch("src/js/**/*.js", gulp.series("tabulator", "core", "modules", "jquery"));
 });
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,201 +1,220 @@
-var gulp = require('gulp'),
-sass = require('gulp-sass'),
-autoprefixer = require('gulp-autoprefixer'),
-cssnano = require('gulp-cssnano'),
-jshint = require('gulp-jshint'),
-uglify = require('gulp-uglify'),
-imagemin = require('gulp-imagemin'),
-rename = require('gulp-rename'),
-concat = require('gulp-concat'),
-notify = require('gulp-notify'),
-cache = require('gulp-cache'),
-livereload = require('gulp-livereload'),
-del = require('del');
-include = require('gulp-include'),
-sourcemaps = require('gulp-sourcemaps'),
-babel = require('gulp-babel'),
-plumber = require('gulp-plumber'),
-gutil = require('gulp-util'),
-insert = require('gulp-insert'),
-fs = require('fs');
+var gulp = require("gulp"),
+  sass = require("gulp-sass"),
+  autoprefixer = require("gulp-autoprefixer"),
+  cssnano = require("gulp-cssnano"),
+  jshint = require("gulp-jshint"),
+  uglify = require("gulp-uglify"),
+  imagemin = require("gulp-imagemin"),
+  rename = require("gulp-rename"),
+  concat = require("gulp-concat"),
+  notify = require("gulp-notify"),
+  cache = require("gulp-cache"),
+  livereload = require("gulp-livereload"),
+  del = require("del");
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @Jojoshua 

Thanks for that, it is always good to keep the build tools up to date.

You seem to have made some very significant changes to the gulp file there.

Would you mind summarising what changes you have made, as there is very little of the original file left there.

Cheers

Oli :)

gulpfile.js
```diff
@@ -1,201 +1,196 @@
...
     gulp.start('modules');
     gulp.start('jquery');
-    });
+});
 
-gulp.task('clean', function() {
+gulp.task('clean', function () {
     return del(['dist/css', 'dist/js']);
-    });
+});
```
Jojoshua: Reworked the default and watch task to convert to using gulp.series

gulpfile.js
```diff
@@ -1,201 +1,196 @@
...
+        .pipe(insert.prepend(version))
+        .pipe(gulp.dest('dist/js'))
+        .on('end', function () { gutil.log('Core complete'); })
+});
 
 
 
 //make jquery wrapper
-gulp.task('modules', function(){
+gulp.task('modules', function (done) {
```
Jojoshua: had to pass through the done function

gulpfile.js
```diff
@@ -1,201 +1,196 @@
...
+                .pipe(rename({ suffix: '.min' }))
+                .pipe(uglify())
+                .pipe(insert.prepend(version))
+                .pipe(gulp.dest('dist/js/modules/'))
         }
-        });
-
     });
 
+    gutil.log("Modules complete");
```
Jojoshua: Added notification when modules was complete. Called done function to let gulp know when the async function was done

gulpfile.js
```diff
@@ -1,201 +1,196 @@
 var gulp = require('gulp'),
-sass = require('gulp-sass'),
-autoprefixer = require('gulp-autoprefixer'),
-cssnano = require('gulp-cssnano'),
-jshint = require('gulp-jshint'),
```
Jojoshua: Removed reference to unused plugins (jshint, imagemin, notify, cache, livereload)

Jojoshua:
Hi @olifolkerd , I went back and updated from the latest version and removed the prettier formatting to make it clearer what changed. I also added comments to highlight where changes were made. Overall, the changes were 

- Removed unused plugins
- Convert gulp.task(deprecated) to use gulp.series

olifolkerd:
Hey @Jojoshua 

Thanks for those tweaks, that is much easier to see what has been changed!

I will have a play with it over the weekend, and if all goes well, i will merge it in for the 4.6 release.

Cheers

Oli :)

olifolkerd:
Sadly @Jojoshua  after merging this in i have had to revert the commit.

when i try and run gulp after making the proposed updates i get the following error:
```
gulp.js:129
gulpInst.start.apply(gulpInst, toRun);
    at C:\Program Files\nodejs\node_modules\gulp\bin\gulp.js:129:20
    at process._tickCallback (internal/process/next_tick.js:61:11)
    at Function.Module.runMain (internal/modules/cjs/loader.js:832:11)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
```

Any thoughts?

Cheers

Oli :)

Jojoshua:
Hmm sounds like similar problem as here https://github.com/gulpjs/gulp-cli/issues/192

Try instructions here to see if it cleans up the old version https://gulpjs.com/docs/en/getting-started/quick-start

olifolkerd:
@Jojoshua i saw that article, my concern is that this upgrade now requires users to install gulp cli globaly before being able to develop with Tabulator. Which i would prefer not to be the case if avoidable. 

Ideally i want it to be a case of running npm install and you are good to go.

Cheers

Oli :)

Jojoshua:
Understood, your call. The general idea for this was to keep the build tools current. It may be different environment for each contributor anyway. In my case by not having v4 caused other problems trying to get started.

olifolkerd:
My point was more that we could install that locally as part of the npm install, thus not needing users to global alter their computer setup to develop tabulator

Jojoshua:
Ah yes agreed so would we need to add the cli to package.json?

olifolkerd:
yarp :)

Jojoshua:
Want me to include that or you got it?

olifolkerd:
hmmm, just tried, no luck :(

Jojoshua:
Maybe gulp is meant to be installed globally?

nrayburn-tech:
I have submitted a PR against the remote branch, https://github.com/Jojoshua/tabulator/pull/2.  It seems to work.

Gulp does not have to be installed globally, nor does gulp-cli have to be installed locally.

Running `npm run-script build` will build as if you called `gulp` from the command line.  Doing this, I have a few differences in the dist files, but I think it is just the newlines because I am on a mac.

olifolkerd:
Hey @nrayburn-tech 

Thanks for that, we would need to modes of operation a run and a watch that updates live like `gulp watch` does.

I think for sake of keeping this in line with other workers we should have an `npm run watch` command and an `npm run build` or `npm run dev` command

Cheers

Oli

nrayburn-tech:
I can submit a new PR if you would prefer, so that this is cleaned up some.  I just didn't know if it should be part of this one to keep it all together, or a separate one.

Between Jojoshua's changes and my changes, the gulpfile has been updated to use gulp4.  I think that if the PR against the remote branch is merged in, then this PR will be updated and require a merge.  ...That is getting a bit confusing at this point.  But, here is the final file from by branch, [gulpfile.js](https://github.com/nrayburn-tech/tabulator/blob/master/gulpfile.js).  (I would recommend you run the diff between the two files in an external tool, I used vscode.)

Changes between current master and final version in my branch.
- Replace gulp.task with functions.  The functions are then exported to make them public for calling from other files or the command line.
- Remove gulp.start and replace it with function calls.
- Remove jshint, imagemin, notify, cache, livereload.

After updating the gulpfile, the npm scripts need to be updated.
```
  "scripts": {
    "build": "gulp",
    "watch": "gulp watch"
  },
```
The scripts can be run with npm run-script build, or watch.

olifolkerd:
Hey @nrayburn-tech 

Would you mind creating a new PR for this forked from the current master branch just to try and separate things a bit. 

While this PR was merged in, i ended up reverting the code as it caused a lot of issues. so a nice clean start would be much easier.

Please make sure the PR includes all updates needed to make a functional change to the system, if the npm scripts need updating, that should be included in the PR

Cheers

Oli :)

Issue #2564, opened 2020-01-06
printAsHtml does not apply the custom group header function (groupHeader)


akohring:
**Describe the bug**
Custom group header formatting specified by a groupHeader function is not applied during printing.

**Tabulator Info**
Tabulator v4.5.3

**Working Example**
https://jsfiddle.net/x1f75t83/1/

**To Reproduce**
A step by step guide to recreate the issue in your JS Fiddle or Codepen:
1. Run the above jsfiddle
2. Click on the print button

**Expected behavior**
The custom header formatting should be in the print window.

**Desktop (please complete the following information):**
- OS: Xubuntu
- Browser Google Chrome
- Version 79.0.3945.88

olifolkerd:
Hey @akohring 

Thanks for getting in touch,

I wouldnt consider this a bug, more of a missing feature. I will reveiw your pull request and get back to you shortly.

Cheers

Oli :)

akohring:
Thanks for the feedback Oli.  I have pushed the suggested fix.  Is it possible to have this fix included in the 4.6 release?

olifolkerd:
Hey @akohring 

Yeah i would certainly be happy to include it in the 4.6 release, when it is ready to go.

Cheers

Oli :)

olifolkerd:
Hey @akohring 

As part of the 4.6 build i had a rethink about how to make the export module more customisable and reusable in a broader sense across Tabulator.

As a result of this rethink u have restructure the module to allow specific callbacks for each export task, print, clipboard and html *(download coming soon)*

I have just pushed a commit to the 4.7 branch to add the new ***groupHeaderPrint*** option along with ***groupHeaderClipboard*** and ***groupHeaderHtmlOutput*** to allow greater customisation of each output.

The 4.7 branch will be released at some point in may.

Cheers

Oli :)

PR #2565, opened 2020-01-06
#2564 apply custom group header function on print

akohring:
This PR will apply the custom group header function on print and fixes the following issue: https://github.com/olifolkerd/tabulator/issues/2564

### DIFF
--- a/src/js/modules/download.js
+++ b/src/js/modules/download.js
@@ -226,7 +226,9 @@ Download.prototype.processGroupData = function(group, visRows){
 
 	var groupData = {
 		type:"group",
-		key:group.key
+		key:group.key,
+		rows: visRows,
+		generator: group.generator
 	};
 
 	if(subGroups.length){
@@ -843,7 +845,26 @@ Download.prototype.downloaders = {
 			function parseGroup(group, calcObj){
 				var groupData = [];
 
-				groupData.push(group.key);
+				var groupWrapper = {
+					_group: group,
+					getKey:function(){
+						return this._group.key;
+					},
+					getField:function(){
+						return this._group.field;
+					},
+					getElement:function(){
+						return this._group.element;
+					},
+					getRows:function(){
+						return this._group.getRows(true);
+					},
+					getVisibility:function(){
+						return this._group.visible;
+					}
+				};
+
+				groupData.push(group.generator(group.key, group.rows.length, group.rows, groupWrapper));
 
 				groupRowIndexs.push(rows.length);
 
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @akohring 

Thanks for submitting a PR for this feature. Could i ask that in future you provide a bit of description on your PR and a link back to the issue that it relates to as it makes it easier to track things.

You are almost there with the Pull Request but there is one thing that needs to be changed before i can accept it.

When Tabulator uses a component out of context it passes a sudo component into the function rather than the actual component. Where as in your code you are actually generating a group component from the group.

If you have a look on line 272 of the `html_table_export.js` you will see what i am talking about, there is a ***cellWrapper*** object rather than a ***CellComponent*** that is passed into the formatter function. It mimics only the bare minimum functions of the component that would be needed for a layout task

There are several reasons for this:

- First of all the groupHeader function may actually be manipulation settings within the group, which is fine if it is called when generating the actual group header, but called for a second time when generating the print out may corrupt the table.
- The render pipeline for the HTML generation is not the same as for the virtual DOM so not all the functions that can be called on a group are relevant.
- When the function is called the developer needs to know whether it is being run to generate the table or in another usage case *(eg, clipboard, download, print, etc...)* In these cases they can use `instaceof` on the component, if it isnt a type of ***GroupComponent*** then they know this isnt the usual use of the function and can take alternative action as needed.


I understand that for a simple change of text of the group header this seems like overkill, but a lot of people have quite involved headers with buttons and in some cases other js widgets, in which case the need to know the difference, because in some cases (clipboard for instance) the html is never added to the DOM so there is no chance to call any JS on it.

Cheers

Oli :)

olifolkerd:
Hey @akohring 

Thanks for that, there are a few functions it may be useful to map in the wrapper. the getVisibility, getRows, getElement, getField, and getKey functions are often used.

Cheers

Oli :)

akohring:
Thanks @olifolkerd 

The above mentioned functions have been added to the wrapper.

Adam

olifolkerd:
Hey @akohring 

Following discussion with a number of users, as part of the 4.6 build i had a rethink about how to make the export module more customisable and reusable in a broader sense across Tabulator.

As a result of this rethink I have restructured the module to allow specific callbacks for each export task, print, clipboard and html *(download coming soon)*

I have just pushed a commit to the 4.7 branch to add the new ***groupHeaderPrint*** option along with ***groupHeaderClipboard*** and ***groupHeaderHtmlOutput*** to allow greater customisation of each output.

The 4.7 branch will be released at some point in may.

As a result of this, i am afraid this PR is no longer needed. Thank you for your effort, as a result of feedback from users the export module is getting some substantial upgrades over the next few releases, so watch this space for even more features

Cheers

Oli :)

Issue #2584, opened 2020-01-14
Download to XLSX does not export the bottom column calculation


akohring:
**Describe the bug**
Download to XLSX does not export the bottom column calculation.

**Tabulator Info**
v4.5.3

**Working Example**
https://jsfiddle.net/c8j5eL0t/

**To Reproduce**
1. Open the jsfiddle example
2. Notice the last row has an Age column calculation with value of 55
3. Click the XLSX button above the table and open the downloaded file
4. Notice the bottom calculation row is not exported

**Expected behavior**
The XLSX export data contains all of the data visible in the table

**Desktop (please complete the following information):**
- OS: Xubuntu
- Browser Google Chrome
- Version 79.0.3945.88

olifolkerd:
Hey @akohring 

The download module has been rebuilt for the 4.7 release to be based on the output of the export module. which will resolve this issue.

Version 4.7 will be released in the next couple of weeks.

Cheers

Oli :)

PR #2587, opened 2020-01-14
#2584 Add top and bottom row calcs during xlsx export when row groups are applied

akohring:
This is a fix for pull request https://github.com/olifolkerd/tabulator/issues/2584

### DIFF
--- a/src/js/modules/calculation_colums.js
+++ b/src/js/modules/calculation_colums.js
@@ -334,7 +334,10 @@ ColumnCalcs.prototype.redraw = function(){
 //return the calculated
 ColumnCalcs.prototype.getResults = function(){
 	var self = this,
-	results = {},
+	results = {
+		top: this.topRow ? this.topRow.getData() : {},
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @akohring 

Thanks for the PR

Can i ask that you look again at your proposed solution. there should be no need to touch anything in the column calculations library to get your proposed code to work.

Making your proposed changes risks breaking the correct function of that module just to add the group calcs to the download, which would be an unacceptable compromise.

Cheers

Oli :)

akohring:
Hey @olifolkerd 

Thanks for reviewing.  I had to move the "top" and "bottom" results properties out of the else condition in calculation_column so that they would always be available, regardless if the group by was applied.  How do you propose the download module should handle retrieving the top and bottom calc properties if they are not added by the calculation_column module?

Adam

olifolkerd:
Hey @akohring 

The download module is getting an overhaul in the 5.7 release to switch it to using the export module like other table export functions, therefore this wont be needed from that point.

Cheers

Oli :)

PR #2619, opened 2020-02-11
Updating download.js

venison:
Autotable has deprecated the use of addPageContent in favor of didDrawPage.  This change update's download.js to use that updated method.

### DIFF
--- a/src/js/modules/download.js
+++ b/src/js/modules/download.js
@@ -645,7 +645,7 @@ Download.prototype.downloaders = {
 		}
 
 		if(title){
-			autoTableParams.addPageContent = function(data) {
+			autoTableParams.didDrawPage = function(data) {
 				doc.text(title, 40, 30);
 			};
 		}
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @venison 

Thanks for your PR, The move to updated functionality makes sense but im afraid there are a few changes that need to be made before the PR can be accepted.

The download module has been totally rebuilt in the 4.7 branch, if you want to make the changes to the ***addPageContent*** function on that branch and resubmit i will be happy to accept them, But it will be released next weekend so i would need them in by then (sorry for the added pressure)

I also notice you have made a few unnecessary changes to the source that are outside the scope of the PR, such as changing var definitions to lets, 

Please don't make changes to the coding style of Tabulator just for the hell of it. Tabulator uses var list declarations at function level and let declarations at functional level. Please dont submit PR's that alter any of the structure of the existing code unless your are fixing a bug, just because you disagree with the structure is not a valid reason for a PR, and will not be accepted.

There also seem to be changes to the triggerDownload function that are unrelated to your comments. If you are attempting to make a different change there, please create a second PR so that changes that need updates dont delay your other functionality.

Cheers

Oli :)

PR #2627, opened 2020-02-19
set this.config.rowGroups to false if there are no groups

### LINKED ISSUE #2626: Grouping headers are printed out after grouping has been removed from the table from xlsx download
**Describe the bug**
With a set of ungrouped data, apply a grouping and export that table to xlsx.
Remove the grouping, and export the table to xlsx
Open the latter xlsx file and note that there are groupings in the file, yet they are not present in the table.

The underlying problem is that there is state that is set to true the first time you download to xlsx with groups. This state is never reset so when you clear the groups the state still says that there are groups.

**Tabulator Info**
4.5.3

**Working Example**
https://codepen.io/tylerroth/pen/wvazKBP

**To Reproduce**
1. Go to https://codepen.io/tylerroth/pen/wvazKBP
2. Click on the xlsx to get an example of an expected download
3. Click on the "group" button
4. Click on the xlsx button to download the grouped spreadsheet
5. Click on the "Ungroup" button
6. Click on the xlsx button
7. Open the downloaded xlsx document
8. Note that there are grouping headers present even though the table is not grouped

**Expected behavior**
the first xlsx file matches the thirds xlsx file

**Screenshots**

 - OS: macOS 10.15.3
 - All

rotht2:
this fixes #2626

### DIFF
--- a/src/js/modules/download.js
+++ b/src/js/modules/download.js
@@ -57,9 +57,7 @@ Download.prototype.processConfig = function(){
 		}
 	}
 
-	if (config.rowGroups && this.table.options.groupBy && this.table.modExists("groupRows")){
-		this.config.rowGroups = true;
-	}
+	this.config.rowGroups = config.rowGroups && this.table.options.groupBy && this.table.modExists("groupRows");
 
 	if (config.columnGroups && this.table.columnManager.columns.length != this.table.columnManager.columnsByIndex.length){
 		this.config.columnGroups = true;
--- a/src/js/modules/download.js
+++ b/src/js/modules/download.js
@@ -816,7 +816,12 @@ Download.prototype.downloaders = {
 
 				fields.forEach(function(field){
 					var value = self.getFieldValue(field, row);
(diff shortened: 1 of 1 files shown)


rotht2:
Adding another commit to master that fixes #2648

olifolkerd:
Hi @rotht2 

Thanks for your PR.

I dont see how this fixes issue #2626, you have refactored the if statement but not actually changed any of the logic so it does exactly the same thing as before. Please could you clarify what this is supposed to do.

Could i ask that you don't submit multiple fixes in one PR, it make it hard to track what has been updated and also means that if one of the fixes work and another dosn't then they are both held up waiting for approval while one is fixed.

As for your second fix for the cell formulas, i agree that this fixes your usage case, but what if a user actually has an array of values and simply wants them included in their spreadsheet, this fix then breaks the table for them. How would you suggest we proceed for this usage case?

Cheers

Oli :)

rotht2:
Closing this PR as the commits have been separated into their own PRs

PR #2664, opened 2020-03-18
Fix Shift+Tab navigation for Number and Range editors

haferje:
The 'number' and 'range' cell editors do not follow the convention when using Shift+Tab to move to the previous cell.  Instead, focus remains in the cell.  Removing the onChange keyhandler for the Shift (keycode 9) key allows it to work as expected.


haferje:
I'll fix the issues. Sorry, not familiar with GitHub Desktop.

PR #2666, opened 2020-03-18
Fix 'range' editor preventing Shift+Tab

haferje:
Using the key combination Shift+Tab in the 'range' editor keeps the focus in the editor instead of moving to the previous editable cell.

I commented out the code, as you did in the case of the 'number' editor, which originally had the same issue.

### DIFF
--- a/src/js/modules/edit.js
+++ b/src/js/modules/edit.js
@@ -617,7 +617,7 @@ Edit.prototype.editors = {
 		input.addEventListener("keydown", function(e){
 			switch(e.keyCode){
 				case 13:
-				case 9:
+				// case 9:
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @haferje 

Sorry for the delay in getting back to you on this one.

Do you have a JS Fiddle that demonstrates the original issue?

Cheers

Oli :)

haferje:
[Here](https://jsfiddle.net/haferje/unav3d5e/9/) is an example using the latest version.  As you can see, you can edit the first cell and Tab through the Example column to the last one, but you can't Shift+Tab through the Example column to the first, because you are prevented.

Issue #2732, opened 2020-04-25
column component get headerfiltervalue does not return a value


vec2k4:
**Describe the bug**
Calling `getHeaderFilterValue()` on a column component doesn not return a value (always `undefined`).

Using the corresponding function directly on the table the value is returned as expected.

**Tabulator Info**
- 4.6.2

**Working Example**
https://jsfiddle.net/vec2k4/ce9tbmxu/

**To Reproduce**
1. Go to the fiddle
2. Press button 'Set header filter by table'
3. A header filter is set programatically in column 'Name'
4. Press button 'Get header filter by table'
5. An alert shows up displaying the current header filter value (this is fine)
6. Press button 'Get header filter by column'
7. An alert shows up displaying `undefined`

**Expected behavior**
Using the `getHeaderFilterValue()` on a column component should return the same value as using  `getHeaderFilterValue(column)` on a table.

**Desktop (please complete the following information):**
- OS: Windows 10 Pro 1909 (Build 18363.778)
- Browser: Edge
- Version 81.0.416.62 (64-Bit)

**Additional context**
Adding a `return` in [column.js](https://github.com/olifolkerd/tabulator/blob/4531ef88896760e051e847dd5981fe18e0445d88/src/js/column.js#L109) should fix the issue.

olifolkerd:
Hey @vec2k4 

Thanks for the headsup,

I will ensure that your PR is merged into the patch release this weekend.

Cheers

Oli :)

PR #2733, opened 2020-04-25
fixed returning the header filter value

vec2k4:
This solves #2732


olifolkerd:
Hey @vec2k4 

Thanks for the fix, i will include this in the patch release later this weekend.

Cheers

Oli :)

olifolkerd:
This will be included in Sundays patch release.

Thanks for the PR

Oli :)

PR #2756, opened 2020-05-03
Select editor: ability to define classes for items

### LINKED ISSUE #2752: css add to select editor
Would be great to have css classes added to option construct for built-in select editor. I am mapping to hierarchical financial categories where debits are red...groups are bold...an optional class list associated with each option would allow css driven solution.

Low priority but helpful for my use-case.

Thanks

takuy:
add ability to add classes to option elements generated by Select editor (fix #2752)

### DIFF
--- a/src/js/modules/edit.js
+++ b/src/js/modules/edit.js
@@ -808,9 +808,10 @@ Edit.prototype.editors = {
 
 			function processComplexListItem(item){
 				var item = {
-					label:editorParams.listItemFormatter ? editorParams.listItemFormatter(item.value, item.label) : item.label,
+					label:item.label,
 					value:item.value,
-					class:item.class,
+					itemParams:item.itemParams,
+					elementAttributes: item.elementAttributes,
 					element:false,
 				};
 
@@ -838,7 +839,8 @@ Edit.prototype.editors = {
 							item = {
 								label:value.label,
 								group:true,
-								class: value.class,
+								itemParams:value.itemParams,
+								elementAttributes:value.elementAttributes,
 								element:false,
 							};
 
@@ -854,7 +856,7 @@ Edit.prototype.editors = {
 					}else{
 
 						item = {
-							label:editorParams.listItemFormatter ? editorParams.listItemFormatter(value, value) : value,
+							label:value,
 							value:value,
 							element:false,
 						};
@@ -870,7 +872,7 @@ Edit.prototype.editors = {
 			}else{
 				for(var key in inputValues){
 					var item = {
-						label:editorParams.listItemFormatter ? editorParams.listItemFormatter(key, inputValues[key]) : inputValues[key],
+						label:inputValues[key],
 						value:key,
 						element:false,
 					};
@@ -894,17 +896,17 @@ Edit.prototype.editors = {
 			while(listEl.firstChild) listEl.removeChild(listEl.firstChild);
 
 			displayItems.forEach(function(item){
+
 				var el = item.element;
 
 				if(!el){
-
+					el = document.createElement("div");
+					item.label = editorParams.listItemFormatter ? editorParams.listItemFormatter(item.value, item.label, cell, el, item.itemParams) : item.label;
 					if(item.group){
-						el = document.createElement("div");
 						el.classList.add("tabulator-edit-select-list-group");
 						el.tabIndex = 0;
 						el.innerHTML = item.label === "" ? "&nbsp;" : item.label;
 					}else{
-						el = document.createElement("div");
 						el.classList.add("tabulator-edit-select-list-item");
 						el.tabIndex = 0;
 						el.innerHTML = item.label === "" ? "&nbsp;" : item.label;
@@ -918,14 +920,15 @@ Edit.prototype.editors = {
 							el.classList.add("active");
 						}
 					}
-				   	if(item.class) {
-						if(typeof item.class === "string") {
-							el.classList.add(item.class);
-						}
-						else if(Array.isArray(item.class)) {
-							item.class.map(function(item) {
-								return el.classList.add(item);
-							});
+					
+					if(item.elementAttributes && typeof item.elementAttributes == "object"){
+						for (let key in item.elementAttributes){
+							if(key.charAt(0) == "+"){
+								key = key.slice(1);
+								el.setAttribute(key, input.getAttribute(key) + item.elementAttributes["+" + key]);
+							}else{
+								el.setAttribute(key, item.elementAttributes[key]);
+							}
 						}
 					}
 					el.addEventListener("mousedown", function(){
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @takuy 

Thanks for the PR.

While i agree it addresses the needs of the issue it is related to i think with a couple of tweaks you could make it so that this offers far more functionality.

What i propose is not adding another property to handle class. but instead the following:

- Instead of processing the ***listItemFormatter*** when the item is parsed, keep it as a function at that point.
- the ***listItemFormatter*** should then be called when it is used in the ***fillList*** function, that way you could pass the containing list element as a third item into the formatter function and the user would not only be able to add classes to it, but directly set styling information etc.

How does that sound to you? would you be up for making those tweaks?

Cheers

Oli :)

takuy:
I do agree that using a function provides greater flexibility for element manipulation and would work even for styling non-"complex list items". 

However, since the `listItemFormatter` is set on a column definition, it limits how you get the data to the function. The proposed solution would force you to keep a separate mapping of classes with items, as the list items themselves are not passed directly to `fillList()`, but are re-built by `parseItems` (for grouping items) and `processComplexListItem` (for the actual items) so any params of the object aside from "label", "value", '"group", and "element" do not get passed along to `fillList()`.

I have another proposal that would still involve adding an additional parameter to the list items, but ultimately would be the more flexible solution:
a `itemParams` (or similar name) that could then be parsed by the end-user's code, as you propose, in `listItemFormatter`. 

That way, if the select values are provided by AJAX, this could be entirely generated on the server-side and simply provided to the Tabulator constructor. Or if the values list is generated based on other values - you can use their data properties to populate itemParams dynamically. 

I can work on this and update the PR over the weekend if you like this direction.

This would then look something like for the end-user:
```js
    listItemFormatter:function(value, title, itemParams, elem){
        if(itemParams && itemParams.style) elem.setAttribute("style", itemParams.style); 
        // or if you pass an object of style elements, use a for-in loop to get attributes

        if(itemParams && itemParams.class) {
            if(typeof itemParams.class === "string") {
                elem.classList.add(itemParams.class);
            }
            else if(Array.isArray(itemParams.class)) {
                itemParams.class.map(function(item) {
                    return elem.classList.add(item);
                });
            }
        }
    },
```

With the values being defined like:
```js
        editorParams: {
          values: [
            {
              //option group
              label: "Men",
              itemParams: {
                class: ["red", "blue"],
              }
              options: [
                //options in option group
                {
                  label: "Steve Boberson",
                  value: "steve",
                  itemParams: {
                    class: "red"
                  }
                },
                {
                  label: "Bob Jimmerson",
                  value: "bob"
                }
              ]
            },
            {
              //option group
              label: "Women",
              itemParams: {
                style: "background-color: #ddd;"
              },
              options: [
                //options in option group
                {
                  label: "Jenny Jillerson",
                  value: "jenny"
                },
                {
                  label: "Jill Betterson",
                  value: "jill"
                }
              ]
            },
            {
              //ungrouped option
              label: "Other",
              value: "other",
              itemParams: {
                class: "blue",
                style: "background-color: orange;"
              }
            }
          ]
        }
```

olifolkerd:
Hey @takuy 

Im afraid those options just add more and more params, where as the proposed update to the listItemFormatter would offer all this additional functionality for no additional arguments.

It being defined on the column in no way limits how you get data into the function, we could pass the cell component from the editor into the function thus providing access to all cell, row and column information. 

Additionally then passing in the element for the list item itself would then allow developers full access to do what ever they like to the element rather than constrain them to just params and classes.

I am not prepared to complicate editors by creating ever nesting parameters for elements inside the editor itself. i would far prefer to use a formatter to allow direct maniplation of the element, in a similar fashion to the way the actual formatters work.

I understand your point of view, but i would not be happy taking tabulator down this route as i feel it will overcomplicate an already complicated editor. and i can just see there becoming more and more of these options over time, when we could add a simple update to an existing function right from the start that would provide a solution to this and all future list manipulation needs in one go.

Cheers

Oli :)

kevans91:
It's unclear how this last proposal by @takuy adds that much additional complexity, to be honest. Tabulator would only need to learn about `itemParams`, which is:

1.) Mostly opaque to Tabulator,
2.) Provided by the data source, and
3.) Passed along with the element to `listItemFormatter` to allow the `listItemFormatter` to act on the `itemParams` (that Tabulator doesn't really know about, except that it needs to pass it along) and transform the element.

olifolkerd:
Hey @kevans91 

It is not that one property adds complexity, it is that is that it solves only that issue, when another need to maniuplate the element comes along in a different way it will need another property later down the line adding more complexity. by using the formatter option it allows a wider range of control by the developer without needing to add another option to the item list.

The formatter approach solves not only this issue but a range of potential future issues, and also brings the list item formatter more in line with other formatters used throughout the table.

Oli :)

kevans91:
> Hey @kevans91
> 
> It is not that one property adds complexity, it is that is that it solves only that issue, when another need to maniuplate the element comes along in a different way it will need another property later down the line adding more complexity. by using the formatter option it allows a wider range of control by the developer without needing to add another option to the item list.
> 

I think this is where I'm not quite understanding -- what other need might arise that isn't covered by this opaque `itemParams`? The data source can provide any arbitrary data in the `itemParams` and act on it in the `listItemFormatter`, so this should theoretically cover almost anything the Tabulator user may need to do without needing future expansion on Tabulator's part.

olifolkerd:
@kevans91

It only allows you to set properties directly on the element with no logic behind them, so you could make a property conditional on a cell or row value for example, you couldn't choose to set the background colour based on the value etc.

It wouldnt allow you to take the element and store the reference elsewhere (for example in an external array or function) for further processing. Which happens more often than you would think in some peoples tables.

When building out these functions i need to think through a wide range of usage cases, not just simple assignment.

Oli :)

olifolkerd:
Ahhh, 

Hangon a second, i may have misread @takuy second post there, my bad.

I see your point about not being able to pass in additional data with a list item and needing to store that information with each item in the list.

Totally with you now. In that case i think a hybrid solution is prefereable. i would be happy to accept the ***itemParams*** idea, but i would also like to see the ***listItemFormatter*** being called in the ***fillList*** function and being passed in the cell component and the element for the list item.

Does that work for you @takuy ?

Cheers

Oli :)

takuy:
Totally works for me! Will get some a new commit in soon. Thanks @olifolkerd and @kevans91.

olifolkerd:
@takuy 

Could you look at using a similar property name and functionality to the  ***elementAttributes*** property on some of the built in editors, just to try and keep things consistent.

Cheers

Oli :)

olifolkerd:
Hey @takuy 

Are you happy that you have made all the changes you need for this? if so i will schedule it to be merged when i merge the 4.7 branch into master in a week or so.

Cheers

Oli :)

takuy:
Yup! I did some tests to make sure nothing broke with the existing formatter behavior, and everything seemed good to me! As long as it meets your standards, it would be great to see this merged in. 

Thanks!

SlappySlap:
hello, nice work, does the display value (final on the grid) have the same attribute ?

Issue #2803, opened 2020-05-21
Enable headerContextMenu and headerMenu for column groups


thenesk:
I would like table users to be able to show and hide columns easily.  A context menu attached to the column headers seems like a standard way to do this.  Following the documentation I was able to do this for most columns, but not for column groups.

It seems like this is not currently supported and I see the same limitation described on stackoverflow, so I think other people would benefit from this feature.  https://stackoverflow.com/questions/61292404/is-there-a-way-to-add-a-headermenu-to-a-group-column

I tried to fix the issue myself by editing `tabulator.js` and making `Column.prototype._buildGroupHeader` look more like `Column.prototype._buildColumnHeader`, specifically by adding `table.modules.menu.initializeColumnHeader(self);`, and this allowed the context menu to trigger.  However, I am not sure if this is the right way to do it.

The context menu that I am adding currently has two items, 'hide column' and 'show all columns'.  With the above hack in _buildGroupHeader, hiding columns seems to work, however showing a column group that has been hidden does not work - the header appears but none of the cells appear.  I am not sure if this is a separate issue or if I broke things by adding the context menu to the group in the first place.

Overall I am very impressed with Tabulator.  It is responsive, looks clean, and has useful features.  Thanks!

thenesk:
Enabling the context menu by calling initializeColumnHeader seems to be working well so I created a pull request:  https://github.com/olifolkerd/tabulator/pull/2806

I resolved the issue with my show column callback by explicitly showing all child columns.

thenesk:
Oops, I see that there is already a pull request addressing the lack of context menus on column group headers: https://github.com/olifolkerd/tabulator/pull/2757

The implementations are a bit different.  I am new to this code base and I am not sure what the trade offs are here, but it looks like the original PR is already slated for inclusion.

olifolkerd:
Closing this issue as it was included in the last release

PR #2757, opened 2020-05-03
added group context menu functionality

vec2k4:
Implements feature request #2749

PR #2806, opened 2020-05-22
enable header menu for column groups

thenesk:
A small addition to _buildGroupHeader so that it calls menu.initializeColumnHeader (just like what _buildColumnHeader does for non-group column headers).  This lets column group headers have context menus just regular column headers.


thenesk:
This PR is in response to issue: https://github.com/olifolkerd/tabulator/issues/2803

I should have studied the existing PRs more carefully as there is already a PR for adding context menus for column groups: https://github.com/olifolkerd/tabulator/pull/2757

I am new to this code base and not sure what implementation is more appropriate.  For better or worse, this PR does not introduce new property names, it treats column group headers the same as column headers by checking for `headerMenu` and `headerContextMenu`.  That said, I see the existing PR is already slated for inclusion so I will not be offended if this PR is rejected.  :)

olifolkerd:
Hey @thenesk 

Unless i am mistaken your PR is trying to do something very different to #2757, that PR is to add context menus to row groups, where as you are trying to add it to column groups

Cheers

Oli :)

thenesk:
well in that case maybe my efforts are not wasted.  :)  Yes, I am interested in adding menus to column groups.

olifolkerd:
Hey @thenesk 

Have you checked that this works correctly in a nested group scenario where each subgroup and column has their own context menu and that the correct menu is shown for each header.

Cheers

Oli :)

thenesk:
I have not tried nested groups before, but I will try that

thenesk:
@olifolkerd, you were right to suggest testing nested column groups each with their own menus.  With my original fix, the top level column group context menu  overshadowed the child columns.  I made a tweak to `Menu.initializeColumnHeader` so that the contextmenu event listener is attached to only the `contentElement` in the case that there are child columns.  The code is ready for your review.  Thanks!

olifolkerd:
Hey @thenesk 

Thanks for looking into that.

One small tweak i would make, rather than checking to see if the group has children 'column.columns.length > 0', there is a boolean set on a column when it is initialised to determine if it is a group or not `column.isGroup`, That is what is used by most other parts of the system, so for consistence i would appreciate if you could take that apparoach.

Cheers

Oli :)

thenesk:
Ok, change made.  :)

olifolkerd:
Hey @thenesk 

Looking into your code a bit further i think the approach you have taken for the nesting of group header context menus may cause unexpected behaviour in certain scenarios.

Yes it certainly does allow for menus to exist at multiple levels on the header. but what if a person only wanted to define a header on the group and it be triggered with a context click anywhere in that header? with the current approach that wouldn't work.

Your initial approach failed because the event propagated, so while it may have initially loaded the correct menu in a nested group scenario it was immediately replaced with the menu of the parent.

We cant stop the event propagation because other events maybe bound to that listener higher up the chain, but inside the menu library we could look to block the immediate opening of a new menu if one has just opened

Cheers

Oli :)

thenesk:
I am not quite sure what the code would look like, but I think I understand the behavior you are describing.  You would like the group level context menu to show when clicking a child if the child does not have a context menu of it own, right?  The current design does not have this behavior as default, but it can support it since the context menu can be applied to the children as desired.  So if someone wants all children to have the same context menu as their parent/group, then they can just set the property on each of the children to the desired menu.

If we make the group context menu apply automatically to all children, then that could save a bit of code for users that want that behavior by default.  But it could complicate things if a user wants the context menu to only apply to the group level and not have a context menu on the children.

For my case it does not matter too much which way you want this to work, I am just trying to think through the use cases.

thenesk:
One possible implementation might be for a column to have a recursive function for getting the header context menu.  If there isn't a locally defined headerContextMenu, but there is a parent column, then call the parent column's getter function.  Is that the sort of thing you are looking for?

olifolkerd:
Hey @thenesk 

I think it would have to work the other way round, otherwise we would have to search for a header menu anytime the header is clicked even if it isnt set. we probably want to sore the menu for the header in the headers module object, if there is one for a parent group it is set to that, otherwise it it set to the current headers menu.

I hope that makes sense.

Cheers

Oli :)

thenesk:
I think I got it working the way you want using one of your earlier suggestions, blocking the creation of a menu if a menu is already open.  This seems to allow child menus to take precedence if defined.  Take a look and let me know.  Thanks,
Mark

olifolkerd:
Hey @thenesk 

Nice approach there.  Happy to merge this in, It will be available in the 4.7 release later today.

Cheers

Oli :)

Issue #2807, opened 2020-05-22
Allow Esc key press to close context menus


thenesk:
**Is your feature request related to a problem? Please describe.**
Context menus do not close when the user hits the escape key.  If the user does not want to select an option from the menu they must click outside the menu to get rid of the menu, and may inadvertently change something in the process (for example clicking on another column and changing the sort).

**Describe the solution you'd like**
I would like the escape key to close any open tabulator context menu, just as the escape key can be used to close browser context menus.

**Describe alternatives you've considered**
NA

**Additional context**
NA

olifolkerd:
This change has been merged in and will be included in the 4.7 release later today :)

PR #2809, opened 2020-05-22
allow escape key press to close context menus

thenesk:
This PR adds a keydown event listener so that the escape key can be used to close context menus.  This corresponds with the feature request issue: https://github.com/olifolkerd/tabulator/issues/2807

### DIFF
--- a/src/js/modules/menu.js
+++ b/src/js/modules/menu.js
@@ -2,6 +2,7 @@ var Menu = function(table){
 	this.table = table; //hold Tabulator object
 	this.menuEl = false;
 	this.blurEvent = this.hideMenu.bind(this);
+	this.escEvent = this.escMenu.bind(this);
 };
 
 Menu.prototype.initializeColumnHeader = function(column){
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @thenesk 

Thanks for the PR, 

Could you please use key codes when checking what key has been pressed to keep things consistent with the rest of the Tabulator codebase.

You also appear to have altered the whitespace on the last line of the file for some reason.

Cheers

Oli :)

thenesk:
The white space change at the end was not intentional, I think my editor added a new line.

I think key codes are deprecated, I based my change off of the advice here: https://stackoverflow.com/questions/3369593/how-to-detect-escape-key-press-with-pure-js-or-jquery

But I can still use key codes if you prefer.  Let me know either way.  :)

olifolkerd:
Hey @thenesk 

Please could you clear out the change to the last line

While they may have recently deprecated this feature, it is currently used in all key identification throughout Tabulator and i would prefer to keep things consistent to avoid confusion.  ***keyCode*** is still functional in all browsers so presents no immediate problem. I will likely switch to a different approach when i complete the code base refactor for the 5.0 release later this year, but until then i would like to keep the code consistent across the project.

Cheers

Oli :)

thenesk:
Ok, I removed that pesky new line and switched to using the keyCode.  :)

olifolkerd:
Hey @thenesk 

Thanks for making those updates, i will merge this in when i bring the 4.7 branch into master in a week or so.

Cheers

Oli :)

thenesk:
Great to hear, thanks Oli!

Issue #2810, opened 2020-05-22
allow table access from downloadDataFormatter


thenesk:
**Is your feature request related to a problem? Please describe.**
When exporting table contents to CSV/JSON/etc it may be helpful to access the table itself.  In my case, the download data format needs special handling for html columns, and needs access to the column definitions to know which columns have html formatting.  In this case, I am grabbing the alt text instead of including all of the cell html, which make much more sense for CSV and JSON exports.  There may be other use cases where table access allows the downloadDataFormatter to be smarter about formatting data for export.

**Describe the solution you'd like**
Including the table as a property of the data that is passed to downloadDataFormatter would do the trick.  Something as simple as: 
'data.table = self.table;'

**Describe alternatives you've considered**
Not using html formatting in cells or not allowing data export.

**Additional context**
NA

olifolkerd:
Hey @thenesk 

The download module is getting an overhaul in the next release to map its functionality through the export module so the current structure will be being replaced anyway and you will gain access to the component objects themselves.

Cheers

Oli :)

PR #2811, opened 2020-05-22
allow table access in downloadDataFormatter

thenesk:
This small addition allows downloadDataFormatters to access the table so they can make more informed formatting decisions.  It corresponds with feature request issue: https://github.com/olifolkerd/tabulator/issues/2810


olifolkerd:
Hey @thenesk 

Thanks for the PR

The download module is about to be completely rebuilt in the 4.7 release to use the export module like other export functions. as a result im afraid this PR will no longer be applicable

Cheers

Oli :)

PR #2814, opened 2020-05-24
Update To Gulp 4

nrayburn-tech:
This should contain all of the changes needed to move to Gulp 4.

Removed a few dependencies from `package.json` and `gulpfile.js` that didn't seem to be used. (Based on the original PR, and looking at the file.)
Converted the `gulp.task` into function calls.
Update gulp to most recent version in `package.json`.
Add `build` and `watch` script to `package.json`.

When running `npm run-script build`, the output should be the same as what it is when using the current build process.

All of the functions are exported, so that they can be called individually if necessary.  For example, you can run just a clean by calling `npx gulp clean`.  Or, you could add it to the `package.json`.  If the function is not exported, then it is not available to the `gulp` call.

I added a `clean` and `build` that is run before the `watch`.  I don't know if this is necessary.  On larger projects it is probably an issue, but I don't know if it is a problem here.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,14 +2,9 @@ var gulp = require('gulp'),
 sass = require('gulp-sass'),
 autoprefixer = require('gulp-autoprefixer'),
 cssnano = require('gulp-cssnano'),
-jshint = require('gulp-jshint'),
 uglify = require('gulp-uglify'),
-imagemin = require('gulp-imagemin'),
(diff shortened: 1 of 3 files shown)


olifolkerd:
Hey @nrayburn-tech 

Thanks for submitting that, on a quick look it all looks great.

I will have a proper play with it in the next couple of days and approve it for merge if all is well.

Cheers

Oli :)

PR #2946, opened 2020-08-19
fix: home/end key capture issue

### LINKED ISSUE #2945: bug: When editing a cell 'home' and 'end' scroll the table rather than moving the cursor
**Describe the bug**
When in an `input`, `home`, `end`, `page-up`, and `page-down` typically move the cursor around in the input. However, these keyboard shortcuts are used to scroll the table, so it cancels edit mode and jumps the user around in the table.

**Tabulator Info**
- 4.7.2

**Working Example**
http://tabulator.info/docs/4.7/edit
^ Edit the name field in the edit demo, then press home, end, page-up, or page-down

Note - the location field works as expected (select box)

**To Reproduce**
1. edit a text field with either an `input` or `textarea`
2. press 'home' or 'end' keys
3. editing gets cancelled, table scrolls

**Expected behavior**
cursor moves to front/end of the input.

rathboma:
fixes #2945.

If you are in edit mode the table will not scroll up or down if you press the keyboard shortcuts.

I think this is desirable behavior as currently it just cancels you out of editing. 

Questions: 
- Should this be behind a flag to turn it on?
- Should this be the existing `verticalNavigation` key maybe? It's not really the same issue

### DIFF
--- a/src/js/modules/edit.js
+++ b/src/js/modules/edit.js
@@ -535,7 +535,7 @@ Edit.prototype.editors = {
 	textarea:function(cell, onRendered, success, cancel, editorParams){
 		var self = this,
 		cellValue = cell.getValue(),
-		vertNav = editorParams.verticalNavigation || "hybrid",
+    vertNav = editorParams.verticalNavigation || "hybrid",
 		value = String(cellValue !== null && typeof cellValue !== "undefined"  ? cellValue : ""),
 		count = (value.match(/(?:\r\n|\r|\n)/g) || []).length + 1,
 		input = document.createElement("textarea"),
@@ -607,7 +607,17 @@ Edit.prototype.editors = {
         	switch(e.keyCode){
         		case 27:
         		cancel();
-        		break;
+            break;
+            
+            case 36: // home
+            e.stopImmediatePropagation();
+            e.stopPropagation();
+            break;
+
+            case 35: // end
+            e.stopImmediatePropagation();
+            e.stopPropagation();
+            break;
 
         		case 38: //up arrow
         		if(vertNav == "editor" || (vertNav == "hybrid" && input.selectionStart)){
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @rathboma 

Thanks for the PR

Im not sure this will work in the virtual DOM scenario,

By allowing the event to propagate down to the element without using the scroll functions you are allowing the table to scroll without populating the virtual DOM, which will result in scrolling to blank space eventually instead of more rows.

In each of the functions you only checking to see if the edit module exists `if (this.table.modExists("edit")) {`  not that someone is actually editing the table. Which means you have effectively bypassed all of these keybindings permanently for all the standard configurations of Tabulator. a module is only not included if someone uses a custom build of Tabulator on their site.

If you want to check if something is being edited then the correct statement would be:
```js
if (this.table.modExists("edit") && this.table.modules.edit.currentCell) {
```
Probably the only thing you need to do is prevent the ***focus*** call at the bottom of the function if a cell is being edited.

```js
if (!(this.table.modExists("edit") && this.table.modules.edit.currentCell) {
    this.table.element.focus();
}
```

Let me know if that makes sense

Cheers

Oli :)

rathboma:
Yeah I wasn't really sure how to do this change the right way tbh. Thanks for the direction.

Thinking on it more, we probably would want most shortcuts disabled during editing, like copy paste and table refresh for example could be really frustrating when editing an input.

Also - my change only works on keycodes, if someone has changed the navigation shortcuts it would still be broken.

Do you think this could work in the shortcuts module instead? That could check edit state I guess too?

 Maybe I could add a flag to enable this behavior.

What do you think?

olifolkerd:
Hey @rathboma 

Thinking this through, this should be a choice per editor depending on how they want to handle the events.

After a bit of experimenting it appears that input  elements care about home and end events but dont do anything with page up or down events.

To get something in for the 4.8 release i have just pushed an update that allows input and textarea based areas to handle the home and end events without passing the event to the keybinding module. (https://github.com/olifolkerd/tabulator/commit/25438737131c4a79dba5aef25984adf429e572d1)

I think that should do most of what you are after.

If you need any other tweaks. please let me know

Cheers

Oli :)

Issue #2935, opened 2020-08-17
Layout fitDataStretch with stretching of all columns instead of only last. Alternatively, fitColumns without shrinkage


imbasimba:
Dear Oli,

I would love a layout mode, which defaults the column size to fit its content, but if there is extra space available, the extra space is spread among all flexible columns.

Basically fitDataStretch, but spreading the available width among all columns instead of just the last.

Or fitColumns, where columns are never shrunk more than the size of its content.

Cheers,
Henrik Norman

olifolkerd:
Hey @imbasimba,

If you set appropriate minimum widths on the columns in ***fitColumns*** layout mode, it should do what you want, you can also control the ratio of shrinkage using the ***widthShrink*** option


See the [FitColumns Docs](http://tabulator.info/docs/4.7/layout#fittowidth) for more details

Cheers

Oli  :)

imbasimba:
Hi @olifolkerd,

Sure, I could manually determine the minimum width for each column. However, it is a hassle to determine the appropriate minimum width for all types of columns & the code is already in Column.prototype.fitToData.

What I want is to be able to set 'minWidth' to 'fitToData' instead of a pixel value, as the desired pixel value is not constant.

Cheers,
Henrik

PR #2953, opened 2020-08-24
FitDataStretch layout with spread of overflow width

imbasimba:
Hi @olifolkerd,

Regarding issue #2935, I found this implementation to be the best solution, and I would argue it is more useful than the original fitDataStretch in most cases.

The main differences are that the overflow width is spread among all flexible columns instead of just the last column & the width of the last column in never set to anything smaller than its content.

Cheers,
Henrik Norman

### DIFF
--- a/src/js/column_manager.js
+++ b/src/js/column_manager.js
@@ -652,7 +652,7 @@ ColumnManager.prototype.redraw = function(force){
 		this.table.rowManager.reinitialize();
 	}
 
-	if(["fitColumns", "fitDataStretch"].indexOf(this.table.modules.layout.getMode()) > -1){
+	if(["fitColumns", "fitDataStretch", "fitDataColumns"].indexOf(this.table.modules.layout.getMode()) > -1){
 		this.table.modules.layout.layout();
 	}else{
 		if(force){
--- a/src/js/modules/layout.js
+++ b/src/js/modules/layout.js
@@ -61,42 +61,44 @@ Layout.prototype.modes = {
 		}
 	},
 
-	//resize columns to fit data the contain and stretch last column to fill table
+	//resize columns to fit data the contain and stretch the flexible columns to fill table
 	"fitDataStretch": function(columns){
 		var colsWidth = 0,
 		tableWidth = this.table.rowManager.element.clientWidth,
 		gap = 0,
-		lastCol = false;
+		lastCol = false,
+		flexibleColumns = [];
 
-		columns.forEach((column, i) => {
-			if(!column.widthFixed){
-				column.reinitializeWidth();
-			}
+		columns.forEach(function (column) {
+			column.reinitializeWidth();
 
-			if(this.table.options.responsiveLayout ? column.modules.responsive.visible : column.visible){
+			if (this.table.options.responsiveLayout ? column.modules.responsive.visible : column.visible) {
+				if (!column.widthFixed) {
+					flexibleColumns.push(column);
+				}
 				lastCol = column;
 			}
 
-			if(column.visible){
+			if (column.visible) {
 				colsWidth += column.getWidth();
 			}
 		});
 
-		if(lastCol){
+		if (lastCol) {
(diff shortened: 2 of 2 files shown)


olifolkerd:
Hey @imbasimba 

Thanks for the PR.

Im always happy to add new layout modes, but could i ask that you submit this as a new layout mode with a new name rather than changing an existing mode. all the modes that are already in the table are there because they have been specifically requested by the userbase and i would not be looking to change their function.

The point of the ***fitData*** series of layout modes is that the columns fit the data they contain, if you are stretching all the columns it no longer fits this moniker and should be called something different.

What visual difference is there between your appraoch and the existing [fitColumns layout mode](http://tabulator.info/examples/4.7#fittowidth) which stretches columns to fit the table and allows control over which column stretches by how much?

Cheers

Oli :)

imbasimba:
Alright, I guess my approach belongs more to the fitColumns series.

The problem I have run into with fitColumns is that it also shrinks the columns below the size of the data (causing truncation -> i.e column name becomes colum... ). 

In theory, setting the minWidth of each column to the size of each column data would (I think) accomplish the same thing as my pull request. But that entails actually calculating the needed pixel size for each column and setting all the minWidth properties as soon as anything changes. I believe a new layout mode to be easier.

I do not think I could accomplish what I want with the widthShrink option.

Let me know if I actually can accomplish my desired layout with the options already available or if I have not articulated the desired result good enough.

Cheers,
Henrik

olifolkerd:
Hey @imbasimba 

Thank you for clarifying, that makes sense.

I would be happy to accept this as a ***fitDataColumns*** layout mode, if you could update the PR, and leave the original fitDataStretch as it was.

Cheers

Oli :)

imbasimba:
Done! :)

olifolkerd:
Hey @imbasimba 

Thanks for that, i will approve this for merge.

It will get merged in for the 4.8 release in a couple of weeks.

Cheers

Oli :)

olifolkerd:
Hey @imbasimba 

I have just been testing the functionality of this after merging and im afraid i dont see the difference between this and ***fitColumns*** layout mode when using the ***widthShrink*** property on columns to prevent them shrinking.

I have reverted the PR for the 4.8 release.

If you could get back to me with where you think the difference is i will consider reinstating.

Cheers

Oli :)

imbasimba:
Hi @olifolkerd,
I have not been able to use widthShrink to accomplish the layout I would like. [See this Fiddle](https://jsfiddle.net/j021bg3t/)

The root issue is that in order for widthShrink:0 to work, you have to set some specific width on the column in the columnDefinition. With very dynamic data, that would be very impractical.

Let me know if I'm wrong, and you can accomplish what I want with just fitColumns.

I could also just happily keep this layout mode on my local fork, if it is not useful to others.

Cheers,
Henrik

mdmangus:
@imbasimba and @olifolkerd,

We are trying out Tabulator 5.4 and so far really like it.  We experiencing the same need described @imbasimba wanting to dynamically fitDataStretch to evenly stretch column instead of just last column when there is not enough columns to scroll.   I guess the mode could be called: "fitDataStretchColumnsEvenly".  

Here "fitColumns" mode  evenly stretches each column to fit when not enough columns:
Screenshot 2023-01-04 at 4 26 40 PM

Layout "fitColumns" mode does not work when the table requires scrolling for the headers are now starting to be cut off losing part of column name, sort icon, custom filter icon, etc.
Screenshot 2023-01-04 at 5 39 07 PM

But if it would instead do "fitDataFill" it then works when columns are expected to scroll with whatever dynamic columns have been added:
Screenshot 2023-01-04 at 5 42 13 PM

Would be nice if we could get a "fitDataStretch" variant layout that when needing to stretch would just stretch columns evenly instead of just stretching last column.  I'm not needing to set widths, shrink, grow that fitColumns provides.  I'm wanting fitData with a differently stretch when needed.  Hope images and this helps describe need.

Is it possible for Tabulator incorporate this additional "fitData" layout mode?

Issue #2594, opened 2020-01-17
Persistence="local" not storing group propery in localstorage?


gitjocke:
Hi! 
First of all, thanks for a awesome tabulator! 

i´m trying to use the persistence setting (with grouping) in tabulator 4.5 but getting a error. Found a fix in 4.5.1 and the error disapeard when using persistenceMode:"local"..
I have the following setup: 
persistence:true,
 persistenceID:"tabperstest2",
persistenceMode:"local",

Looking in localstorage (in chrome ) i see this; 
tabulator-tabperstest2-sort
tabulator-tabperstest2-filter
tabulator-tabperstest2-columns
but no group or pagination setting? 

sort and filters are persistent but i wan´t the grouping to be persistent. 

Any ideas`?

olifolkerd:
Hey @gitjocke 

If you think you have found an error please use the bug reporting issue template, and have a read of the [Bug Reporting Guide](http://tabulator.info/community#bug) they will prompt you to provide all the info we need to help you out.

It is really important that you create a JS  fiddle or code pen that shows your table in action so we can understand the full setup.

Cheers

Oli :)

PR #2963, opened 2020-08-31
Declare var in getGroupConfig

LilDevil:
I ran into the same problem that was reported in #2594. After stepping though in the debugger, I found a simple error I have also made dozens of times.


olifolkerd:
Hey @LilDevil 

Thanks for the PR, looks effective, i will approve this for merge when i put together the 4.8 release next weekend.

Cheers

Oli :)

Issue #1563, opened 2018-11-06
4.1: Always get `Uncaught (in promise) undefined` in ajaxProgressiveLoad scroll mode


puffsnow:
**Describe the bug**
Hi Oil, I'm appreciate with this amazing library! Thanks for a lot of help!
Here I'm trying to upgrade my tabulator from 3.5 to 4.1, but I get some problems.
My tabulator use ajax scroll mode to make paging, and when I scroll to bottom ( also last page ), I always get error about "Uncaught (in promise) undefined". The "Page.prototype.nextPage" will always call reject() but I'm not sure how to handle it. Is it a bug or I should do something in my creator?

**Tabulator Info**
- Version: 4.1
`self.table = new Tabulator(selector, {
        index: id,
        height: "600px",
        layout: "fitDataFill",
        ajaxURL: ajaxUrl,
        ajaxParams: ajaxParams,
        ajaxProgressiveLoad: "scroll",
        paginationSize: 20,
        placeholder: "",
        columns: columns,
        columnVertAlign: "bottom",
    });`

**To Reproduce**
When the table scroll to bottom and there is already last page, although it will not call ajax again, but the error message will always occurs.

**Desktop (please complete the following information):**
Browser: chrome

**Additional context**

Rodbjartson:
` ajaxURL: ajaxUrl`
Should this not be your own url?

olifolkerd:
I agree with @Rodbjartson 

Both your ***ajaxURL*** and ***ajaxParams*** are set to variables, without knowing what these are set to it is difficult to diagnose your issue.

Cheers

Oli :)

puffsnow:
Hi @olifolkerd,
Sorry, because it's a business project so I hide them as parameters. I made another test case and it still has the same problem. Here is my code and please noted me if there is any unclear.
tabulator creator (in coffeescript) :
```
$(document).ready ->
  table = new Tabulator("#ttt", {
    index: "id",
    height: "200px",
    layout: "fitDataFill",
    ajaxURL: "/home/test_a",
    ajaxProgressiveLoad: "scroll",
    paginationSize: 10,
    placeholder: "",
    columns: [
        {title:"id", field:"id", sorter:"string"},
        {title:"point", field:"point", sorter:"number"},
    ],
    columnVertAlign: "bottom",
  });
```
response data:
```
[...]

PR #2966, opened 2020-09-01
Add call rejection handling for nextPage to all clicks

colindh:
I ran into an issue related to https://github.com/olifolkerd/tabulator/issues/1563 but for other page changes. I think catching call rejection for any pagination is a good idea because the data behind the call could be changing real time.


olifolkerd:
Hey @colindh 

Thanks for the PR, i will include it in the 4.8 release next weekend.

Cheers

Oli :)

PR #3054, opened 2020-10-23
Fix localisation not working in collapsed column titles

Aresius423:
Hi,

here's a PR in lieu of a bug report.
Localised header names revert back to the default title when collapsed. See: https://jsfiddle.net/ckqsL9je/

If there's a better way of getting the translated title, please feel free to edit this.
item.title is kept for backwards compatibility and fallback purposes.

### DIFF
--- a/src/js/modules/responsive_layout.js
+++ b/src/js/modules/responsive_layout.js
@@ -249,11 +249,13 @@ ResponsiveLayout.prototype.generateCollapsedRowData = function(row){
 				};
 
 				output.push({
+					field: column.field,
 					title: column.definition.title,
 					value: column.modules.format.formatter.call(self.table.modules.format, mockCellComponent, column.modules.format.params)
 				});
 			}else{
 				output.push({
+					field: column.field,
 					title: column.definition.title,
 					value: value
 				});
@@ -267,6 +269,7 @@ ResponsiveLayout.prototype.generateCollapsedRowData = function(row){
 ResponsiveLayout.prototype.formatCollapsedData = function(data){
 	var list = document.createElement("table"),
 	listContents = "";
+	let columns = this.table.modules.localize.lang.columns;
 
 	data.forEach(function(item){
 		var div = document.createElement("div");
@@ -276,7 +279,7 @@ ResponsiveLayout.prototype.formatCollapsedData = function(data){
 			item.value = div.innerHTML;
 		}
 
-		listContents += "<tr><td><strong>" + item.title + "</strong></td><td>" + item.value + "</td></tr>";
+		listContents += "<tr><td><strong>" + (columns[item.field] || item.title) + "</strong></td><td>" + item.value + "</td></tr>";
 	});
 
 	list.innerHTML = listContents;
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @Aresius423 

Thanks for the PR.

While i agree there is definitely an issue there, im afraid your approach does not make correct use of the localisation system.

When working with modules you should let the module do the work, not try and access its internal variables. This is especially true in the localisation system which can handle multiple localisations at once.

The simple solution would be to use the ***getText*** function on the localisation module that will retrieve the text string for a for a given path:
```js
this.table.modules.localize.getText("columns", "item.field");
```

However that wouldn't allow the value to be switched out it the user changes the localisation language in use on the table while that is being displayed.

To make this work properly you would need to use the ***bind*** function on the module which is called whenever the localisation changes and updates all the necessary values across the table.

an example of this can be found in column.js where the column headers themselves are localised:
```js
table.modules.localize.bind("columns|" + def.field, function(text){
    titleElement.value = text || (def.title || "&nbsp;");
});
```
But this would require a complete rebuild of the ***formatCollapsedData*** function as it will need to manage the list contents as actual DOM nodes not just a simple text string.

If you feel comfortable doing this please resubmit the PR, if not feel free to submit a bug issue and i can add it to the roadmap for a patch release.

Cheers

Oli :)

Aresius423:
That's fair - I've updated the PR accordingly, let me know if you're happy with it.

olifolkerd:
Hey @Aresius423 

I would say you are 99% of the way there, there are just a couple of coding style tweaks to bring it inline with the rest of Tabulator and i think we are sorted.

On line 288 you are defining a variable inside an if, please dont do this, it is bad practice and the variable definition will only end up being hoisted to the top of the function when run anyway. I can see you were trying to not generate the div when not needed, which is much better from an efficiency point of view, but the correct approach would be to define the empty variable at the top of the function, then initialize it to the div in the if statement if needed.

And just to be consistent with the coding style of the rest of tabulator on lines 276 - 278 you append the row etc to the list before filling them, would you mind moving them to the end of the function and doing actual appending after the contents has been populated. The reason for this is that in a lot of other areas of the table the containing element is already visible on the DOM and it is more efficient to build up the contents and append it in one go rather than appending it to a visible element. I understand that isn't the case in this instance but it will help readability of the code if we can follow the same structure here.

Im planning on releasing version 4.9 in a week, if you could get the changes in by then i would love to include this in that release.

Thanks for all your hard work.

Cheers

Oli :)

Aresius423:
No problem! I reformatted these bits.

olifolkerd:
Great Stuff, Thanks!

I will get this merged in when i build the 4.9 release next week.

Thanks for your contribution.

Cheers

Oli :)

birbilis:
Btw, do columns with empty values that get collapsed disappear by default? If not, is there an option for that?

olifolkerd:
No they do not, as it may be desirable to show this.

I'm not sure I would want to add that as a built in feature, but you could easily define your own responsive collapse formatter function to add any bespoke functionality

birbilis:
The idea is that when collapsing columns like that the screen estate is quite small so there is no reason wasting it to show empty items in the list of collapsed columns under each row. e.g. at http://minedu.gov.gr/Covid19-Schools.html I have a Remarks row that is empty for some schools. If collapsed I'd prefer it is hidden for those rows that don't have remarks, not just write "Remarks" with no value next to it for those rows

PR #3074, opened 2020-11-12
fix: 🐛 Use floating numbers for cell width

SheepFromHeaven:
There was an issue when using fitDataFill + first column frozen leading to having always a scrollbar, even though everything would fit on the screen.

By using the real width of the cell and not the rounded offsetWidth (see note here https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetWidth), the tableRows will not become one to two pixel longer than the whole table, and scrollbars will only appear once nessecary.

### DIFF
--- a/src/js/cell.js
+++ b/src/js/cell.js
@@ -663,7 +663,7 @@ Cell.prototype.clearWidth = function(){
 };
 
 Cell.prototype.getWidth = function(){
-	return this.width || this.element.offsetWidth;
+	return this.width || this.element.getBoundingClientRect().width;
 };
 
 Cell.prototype.setMinWidth = function(){
--- a/src/js/column.js
+++ b/src/js/column.js
@@ -1353,7 +1353,7 @@ Column.prototype.fitToData = function(){
 		});
 
 		if(maxWidth){
-			self.setWidthActual(maxWidth + 1);
+			self.setWidthActual(maxWidth);
 		}
 
 	}


olifolkerd:
Hey @SheepFromHeaven 

Thanks for the PR

You have definitely identified an issue there, but im afraid your approach to resolving the problem will cause more trouble for other parts of the table than it solves. You should not use floating point numbers for any width considerations, especially when building out the Virtual DOM as this will cause a great many issues in other parts of the system. the maths must be constrained to integers.

There are known issues with a number of browsers in how they inconsistently handle rounding of fractional widths of the same value when retrieving them using `offsetWidth` & `offsetHeight`. *(i have in fact just fixed such a zoom based issue in another part of the table)*.  The issue actually arises because the values return from these properties are inconsistently rounded causing offset calculation errors when using these values to set new widths *(ie they are off by one or two pixels as a result of browsers inconsistently choosing to round the numbers in different ways)*.

The principal of getting the value from the `getBoundingRectangle` function is correct  but you must then handle flooring them yourself to maintain consistency rather than letting the browser choose how to round.

I am also unsure as to why you are adding one in the `setWidthActual` function, if this is to try and account for the fractional nature of the returned number, then as i mention above, handling the rounding yourself will render this unnecessary.

I am aiming to release version 4.9 next weekend. if you can update the PR before then i would be happy to include this PR in the release.

Cheers

Oli :)

SheepFromHeaven:
Hey @olifolkerd thanks for the Feedback! For the last part: i actually removed the + 1 which, as you said, seems like a workaround. The problem is, that we cannot fix this problem with rounding in any direction, as floating pixel values are actually a thing.
By using floor, the text in the column will break the text and using ceil leads to scrolling. So I would actually need the floating pixel value there.

Any other ideas? :/

olifolkerd:
Hey @SheepFromHeaven 

Do you have an example JS Fiddle that demonstrates the issue. it will be easier to base this discussion around a physical example. 

The problem you will face is that the small change you have made is actually to one of the most widely used functions in the table and it will have far reaching consequences to have to go through and refactor all the other table modules to cope with the change.

Cheers

Oli :)

SheepFromHeaven:
@olifolkerd will setup an example :)

SheepFromHeaven:
hey @olifolkerd please have a look here: https://codepen.io/SheepFromHeaven/pen/QWEPrbw

If you now try to resize, the rows will a small bit outside the table on some screen sizes and scrolling will happen, even though it could break.

SheepFromHeaven:
@olifolkerd could you already have a look? 🙂

olifolkerd:
Hey @SheepFromHeaven 

I will have a look when i get a second, i have a lot of people competing for my attention so i can only focus on a few things at a time, i try and review PR's every couple of weeks and i have been primarily focused on getting the 4.9 release out the last couple of weeks.

Cheers

Oli :)

SheepFromHeaven:
No probs. Thx for the expectation management 🙂

PR #3097, opened 2020-11-27
Row selection fix for elements not yet initialized

imbasimba:
I believe this PR explains the issue better than raising an issue to explain the bug.

Anyway, see this [https://jsfiddle.net/rk9n6qc1/2/](https://jsfiddle.net/rk9n6qc1/2/)
1. Click on "select all" checkbox
2. Scroll to the last row
3. Click the checkbox on the last row
4. Error: Selection style not changed & no row selection event is fired

The cause seems to be rows that have not yet been fully created. In these cases, the row.getElement() eventually get into SelectRow.prototype.initializeRow where the row.modules.select values are overridden. 

I did the quick and easy thing of just moving the row.getElement call up a few rows. However, I suspect this issue may be solved cleaner by not overriding the modules.select inside initializeRow.

### DIFF
--- a/src/js/modules/select_row.js
+++ b/src/js/modules/select_row.js
@@ -218,6 +218,7 @@ SelectRow.prototype._selectRow = function(rowInfo, silent, force){
 
 	if(row){
 		if(this.selectedRows.indexOf(row) == -1){
+			row.getElement().classList.add("tabulator-selected");
 			if(!row.modules.select){
 				row.modules.select = {};
 			}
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @imbasimba 

Thanks for the PR, This issue has been caused by an update in version 4.9 that no longer creates rows when they are first loaded into the table *(which massively improves table performance)*, instead rows are only initialized when the row is first added to the DOM.

You are 100% correct that the rows were not yet initialized. there are a number of functions that when called on the row trigger initialization,and `getElement` is one of these, so in this case the solution works perfectly.

There is also the `create` function that can be called on a row, which initializes the row correctly and generates the elements necessary for other modules to use. which can also be used instead of moving the get element function call, but it is an either or scenario.

I will get this merged in and include it in this evenings patch release.

Cheers

Oli :)

PR #3203, opened 2021-02-09
fix issue with fitToData() setting a column width for group columns

### LINKED ISSUE #3202: Resizing columns using dynamically generated columngroups does not resize the column header
**Describe the bug**
I'm using a combination of static columns that are created prior data loading and dynamic columns that are created while the data loading is in progress. Some of the static columns form a column group and also the dynamic colums are grouped together forming another column group. 
So I create the dynamic columns and group using the `addColumn` function inside the `dataLoading` callback. Everything works as expected but when it comes to resizing the dynamic columns, I get weired results. The data part of the column is resized correctly, but the column header still remains at its old size. Please keep in mind that the fiddle is an extremly simplified version of my real `dataLoading` callback function. Column autogeneration is not an option for my use-case.

If I set the columns using `this.setColumns(this.getColumnDefinitions())` in the `dataLoading` callback after dynamically adding the columns as a workaround the error doesn't occur.

**Tabulator Info**
- Using Tabulator 4.9.3 and also tried current 5.0.0 branch

**Working Example**
- https://jsfiddle.net/vec2k4/abcypvmq/

**To Reproduce**
1. Go to the fiddle
2. Press "Load data" button
3. Resize one of the columns belonging to the "Dynamic column group" 
4. See the resizing error
5. Now resize one of the columns belonging to the "Static column group"
6. See the resize working
7. Press "Load data with workaround" button
8. See the dynamic and static columns resize correctly

**Expected behavior**
I would expect that resizing the columns, even if they are dynamically added, works as it should.

**Screenshots**
 - Error:
 Bug
 - With workaround:
Workaround

 - OS: Win 10 latest
 - Browser: Edge version 88.0.705.63 (Chromium)

 - Didn't test on mobile browsers

vec2k4:
This should fix #3202

Problem was that `Column.prototype.fitToData` sets a fixed column width for column group headers, which yields to HTML like this:

```html
<div class="tabulator-col tabulator-col-group" role="columngroup" aria-sort="none" 
     aria-title="Dynamic golumn group" title="" style="width: 202px; height: 49px;">
  ...
</div>
```

where it should be like this:

```html
<div class="tabulator-col tabulator-col-group" role="columngroup" aria-sort="none" 
     aria-title="Static column group" title="" style="height: 49px;">
  ...
</div>
```

The fix corrects this behavior and doesn't set the `style="width: ...px;"` for column group headers.

### DIFF
--- a/src/js/column.js
+++ b/src/js/column.js
@@ -1354,9 +1354,13 @@ Column.prototype.reinitializeWidth = function(force){
 	}
 };
 
-//set column width to maximum cell width
+//set column width to maximum cell width for non group columns
 Column.prototype.fitToData = function(){
 	var self = this;
+
+    if (self.isGroup){
+    	return;
+    }
 
 	if(!this.widthFixed){
 		this.element.style.width = "";


olifolkerd:
Hey

Sorry for the long delay in getting back to you, ive had a hellish month at work, but im back now.

Could you resumbit this against the 5.0 branch ive made significant changes to the file structure and i wont be able to easily merge in changes on master.

Cheers

Oli :)

vec2k4:
Hey Oli,

I will take a look at the 5.0 branch and come back to you with an updated pull request.

Best regards

Ralf

vec2k4:
See #3263 for updated pull request. Will close this one.

Issue #3249, opened 2021-03-25
Rows cannot be moved in a table with a scrollbar


piroor:
**Describe the bug**

Rows cannot be rearranged via drag and drop if a table rendered with the virtual DOM and there are large number rows and the table is scrollable.

**Tabulator Info**

- Tabulator 4.9.3

**Working Example**

https://jsfiddle.net/piro_or/w5cadgsz/5/

**To Reproduce**

1. Go to https://jsfiddle.net/piro_or/w5cadgsz/5/
2. Press the mouse button on the row "data-3".
3. Keep the button pressed and move your mouse up or down.
4. The row "data-3" suddenly fixed to the top of the table, and the row "data-0" become invisible. The row below the mouse cursor becomes blank, and the blank row moves following to the mouse cursor.
   * Google Chrome reports an error in the console:
     
     ```javascript
     tabulator.min.js:11 Uncaught TypeError: Cannot read property 'mousemove' of undefined
         at tabulator.min.js:11
         at Array.forEach (<anonymous>)
         at O._bindMouseMove (tabulator.min.js:11)
         at O.startMove (tabulator.min.js:11)
         at tabulator.min.js:11
     ```
   * Mozilla Firefox reports an error in the console:
     
     ```javascript
     Uncaught TypeError: e.modules.moveRow is undefined
         _bindMouseMove https://cdnjs.cloudflare.com/ajax/libs/tabulator/4.9.3/js/tabulator.min.js:11
         _bindMouseMove https://cdnjs.cloudflare.com/ajax/libs/tabulator/4.9.3/js/tabulator.min.js:11
         startMove https://cdnjs.cloudflare.com/ajax/libs/tabulator/4.9.3/js/tabulator.min.js:11
         checkTimeout https://cdnjs.cloudflare.com/ajax/libs/tabulator/4.9.3/js/tabulator.min.js:11
         setTimeout handler*O.prototype.initializeRow/< https://cdnjs.cloudflare.com/ajax/libs/tabulator/4.9.3/js/tabulator.min.js:11
         initializeRow https://cdnjs.cloudflare.com/ajax/libs/tabulator/4.9.3/js/tabulator.min.js:11
         generateElement https://cdnjs.cloudflare.com/ajax/libs/tabulator/4.9.3/js/tabulator.min.js:4
         create https://cdnjs.cloudflare.com/ajax/libs/tabulator/4.9.3/js/tabulator.min.js:4
[...]

PR #3250, opened 2021-03-25
Allow to move a row in a scrollable table

piroor:
This change looks to fix the problem #3249 for me (confirmed with Firefox). Is this approach valid?


piroor:
Oops, sorry the last line was modified by the GitHub online code editor unexpectedly .

PR #3255, opened 2021-03-26
Allow select editor to drop up

Sander-Chief:
Right now the select editor will only drop down, even at the bottom of the document, which is not always the desired behavior.
This change adds a new optional property **allowDropup** available in **editorParams** object.
If the property is set to **true**, the editor will drop up if it's height is bigger than available space at the bottom of document.

### DIFF
--- a/src/js/modules/edit.js
+++ b/src/js/modules/edit.js
@@ -1263,6 +1263,17 @@ Edit.prototype.editors = {
 				});
 
 				document.body.appendChild(listEl);
+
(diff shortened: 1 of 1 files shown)


olifolkerd:
The select editor is being merged with the auto complete editor and completely rebuilt in the upcoming 5.2 release and will be using the functionality of the menu module to ensure the correct positioning of elements on screen.

As a result I will close this PR.

PR #3296, opened 2021-05-04
Preserve vertical scroll position when adding a column to the table

yujiberra:
Hi there,

It's totally possible that there are situations I'm unaware of where this would cause issues, but in my testing I didn't encounter any problems with preserving (i.e. not resetting) vertical scroll position when adding a column to the table. Would be great if you could take a look!

Thanks,
Yuji


olifolkerd:
Hey @yujiberra 

please submit changes against the 5.0 branch.

The codebase has undergone a complete refactor for the 5.0 release and is no longer compatible with the 4.9 code.

Cheers

Oli :)

yujiberra:
Understood! Will migrate to 5.0 and give it a shot

PR #3298, opened 2021-05-05
Insert missing 'this' in VirtualDomHorizontal.js

yujiberra:
Just a missing 'this' before an object property


olifolkerd:
Hey @yujiberra 

`this.options` is actually a function, not a property, but i get where you are going with this, i will make the correct change this end.

Thanks for the headsup

Cheers

Oli :)

yujiberra:
Makes sense, thanks Oli!

PR #3302, opened 2021-05-08
Fix event bugs

yujiberra:
Fixing two issues with InteractionMonitor:
* track was causing an error since Events don't have a path property
* Group events weren't firing properly because bindComponents wouldn't look for a target (unless a row or cell event listener exists)


olifolkerd:
Hey @yujiberra 

Could you confirm what browser / OS / event types you are trying to trigger?

The interaction module works as expected when i use it and i see the path property on all my events.

Cheers

Oli :)

yujiberra:
Hi Oli,

Thanks for looking into this! So here's a StackBlitz based on the latest 5.0 branch (eb0992b) that gives me a "path is undefined" error when I activate a group header click event (which I set to toggle a group): https://stackblitz.com/edit/yuji-tabulator-path-undefined

Here's the same StackBlitz but with the changes from this PR applied, and the group header clicks are working: https://stackblitz.com/edit/yuji-tabulator-path-undefined-2 (Actually, the group visibility toggle only works the first time, but that's a different bug; a console.log is correctly occurring on each click.)

I guess if other events are working for you then this might not be the correct fix, but hopefully this points you in the right direction. Oh yeah, as for browsers I've been testing with Chrome and Firefox on Windows.

Best,
Yuji

yujiberra:
Just remembered that the commit I made to this pull request https://github.com/olifolkerd/tabulator/pull/3305 is the fix for the bug I mention above where group visibility toggle only works the first time, so I'll push that onto this branch shortly.

Here's a StackBlitz with that fix added, where toggling groups with group click events seems to be fully working: https://stackblitz.com/edit/yuji-tabulator-path-undefined-4

olifolkerd:
Hey @yujiberra 

The initial code snippet that errors in your example runs absolutely fine when it is running outside of stackblitz, when i run it on my local machine it is without issue.

I would prefer to understand why this is occurring in this instance as it would appear to be non standard behaviour.

Please don't merge pull requests, each PR should fix only one issue, that way if there are problems with the PR it dosne hot hold up other fixes.

Cheers

Oli :)

olifolkerd:
Hey @yujiberra 

Looking into this further, the `composedPath` method is only available in modern evergreen browsers. Because Tabulator currently supports IE 11, all codebase changes need to maintain that compatibility.

Could you update your PR so that it uses `.composedPath()' if it is present and falls back to  `.path` otherwise.

There is also no need to add the group check to row 163, it is deliberately not included, because all group elements are also row elements and to included it will result in a double event.

If you can fix those and remove the unrelated changes to the GroupRows module which should be in their own PR then i would be happy to approve this.

Cheers

Oli :)

yujiberra:
Hi Oli,

Thanks for investigating `composedPath`; I looked into `path` as well and it seems that `path` isn't available on IE11 so there might still be an IE11 compatibility issue here? In either case though, using `composedPath` with a fallback to `path` is still strictly safer than using one or the other so I've made that change.

> Please don't merge pull requests, each PR should fix only one issue, that way if there are problems with the PR it dosne hot hold up other fixes.

Yep, done, sorry about that. I encountered those issues working with the same test cases but in retrospect they are clearly separate issues.

> There is also no need to add the group check to row 163, it is deliberately not included, because all group elements are also row elements and to included it will result in a double event.

I just want to confirm this since I do still get an error trying to trigger a group click without that change. Here's a sample with only the switch to using `composedPath` (with `path` fallback): https://stackblitz.com/edit/yuji-tabulator-path-undefined-5 I get a "component.getComponent is not a function" error when clicking on a group (on Firefox and Chrome, both in StackBlitz and running on my local machine). (And adding the `listener.components.includes("group")` condition does fix the issue for me: https://stackblitz.com/edit/yuji-tabulator-path-undefined-6)

Anyway, I suppose that should arguably be a different PR too so I'll remove it as you originally suggested.

Thanks,
Yuji

PR #3305, opened 2021-05-09
Fix bug where GroupRows.displayIndex was not being set

yujiberra:
This fixes an issue where toggling group visibility would only work the first time after initializing a table, after which no groups could be toggled. Not 100% sure this is the right fix for this as I wasn't able to fully understand RowManager.setDisplayRows


yujiberra:
Just realized that this only happens when clicking on a group to toggle, and doesn't happen if you e.g. get the components with table.getGroups(), so I'm going to close this and add the fix to this other pull request https://github.com/olifolkerd/tabulator/pull/3302

Issue #3309, opened 2021-05-11
Access-Control-Allow-Origin header cause CORS issue


mbellagamba:
**Describe the bug**
Fetching data using an `ajaxURL` can be blocked by CORS policy because the request include an invalid `Access-Control-Allow-Origin` header. The browser console log this error:

>Access to fetch at 'https://swapi.dev/api/planets/' from origin 'https://2htzh.csb.app' has been blocked by CORS policy: Request header field access-control-allow-origin is not allowed by Access-Control-Allow-Headers in preflight response.

Request headers are automatically set by the library and there is no way to remove it. The `Access-Control-Allow-Origin` is a [**response** header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) and it should not be included in requests. The correct default is to set the [`Origin` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) instead.

I suggest to change the behavior in file _src/js/modules/ajax.js_, replacing lines 403-404 with:
```javascript
if(typeof config.headers["Origin"] === "undefined"){
  config.headers["Origin"] = window.location.origin;
```
I've just replaced `"Access-Control-Allow-Origin"` with `"Origin"`.

**Tabulator Info**
4.9.3 (but the bug is present also in version 5.0 in file [src/js/modules/Ajax/defaults/loaderPromise.js](https://github.com/olifolkerd/tabulator/blob/5.0/src/js/modules/Ajax/defaults/loaderPromise.js) at lines 50-51)

**Working Example**
I provided a [CodeSandbox](https://codesandbox.io/s/tabulator-access-control-allow-origin-issue-2htzh) showing the issue. The suggested update will solve the issue. I've tested locally on my machine.

**To Reproduce**
Just load the CodeSandbox to spot the issue. For a further understanding you can open Browser Dev Tools and che the error in console or in Network tab.

**Expected behavior**
I expect the table shows the data.

**Desktop (please complete the following information):**
 - OS: MacOS, Linux
 - Browser Chrome
 - Version 90.0.x

**Additional context**
Can I open a PR to solve this issue? I'd like to use this library in my project but this issue is blocking me.

olifolkerd:
Hey @mbellagamba 

That response sounds more like your sever is specifically configured to reject that header.

It has been used on literally thousands of websites without any issue so i would like to understand a bit more about with this is now a problem in your usage case before proceeding with this.

Cheers

Oli :)

[...]

PR #3323, opened 2021-05-26
fix origin header sent by default

mbellagamba:
Fixes the issue #3309.

The `Access-Control-Allow-Origin` is a response header and it should not be included in requests. The correct default is to set the `Origin` header instead. I've explained in detail the motivation for this PR in the issue #3309. 

Further reference:
- [`Access-Control-Allow-Origin` - MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin)
- [`Access-Control-Allow-Origin` - whatwg spec](https://fetch.spec.whatwg.org/#http-access-control-allow-origin)
- [`Origin` - whatwg spec](https://fetch.spec.whatwg.org/#origin-header)


olifolkerd:
Hey @mbellagamba 

Im afraid we are not currently accepting PR's on the master branch.

The library has been completely rebuilt for the 5.0 release and most of the original files are no longer there.

If you are happy to resubmit this om the 5.0 branch i would be happy to consider it for merge.

Cheers

Oli :)

mbellagamba:
Hi, thanks for reviewing it. I'll resubmit the PR on the 5.0 branch. Thanks for your work!

mbellagamba:
Hey @olifolkerd, I've updated the code and the Pull Request base branch. You should be able to accept the update. Thanks.

Issue #3173, opened 2021-01-20
when dataTreeElementColumn is set and the column is editable, a click in the cell with the tree element makes the element disappear


NnMllr:
**Describe the bug**
When the dataTreeElementColumn is set and this column has an editor, by clicking in one cell of this column the branchElement or the Collapse/Expand Element disappear which makes the row not collapsible/expandable anymore.

**Tabulator Info**
4.9.3

**Working Example**
http://jsfiddle.net/6edcjxpf/2

**To Reproduce**
A step by step guide to recreate the issue in your JS Fiddle or Codepen:
1. Go to any of the collapsable rows in the first table
2. Click on the cell of the row where the collapse/expand element is to edit it
3. Leave the cell
4. See error => the CollapseElement disappears

**Expected behavior**
The clicked cell should not loose the collapse/expand element as seen in the second table in the jsfiddle which has the dataTreeElementColumn not set.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 84 (Latest version)

**Additional context**
This isn't a problem when the dataTreeElementColumn option is not set.

olifolkerd:
Hey @NnMllr 

That is definitely a regression there, 

I will look at getting a fix for that out in the next release

Cheers

Oli :)

NnMllr:
Thanks, I really like using Tabulator :)

danielHin:
Hey Guys,

ran into the same problem and had a look into that issue. I added a small fix and tested successfully on my local machine  (using the examples from the jsfiddle in this issue)

Hope this helps :)

olifolkerd:
Hey @NnMllr 

I have pushed a fix for this to the master branch and will include it in the next patch release.

Cheers

Oli :)

PR #3325, opened 2021-05-28
Fix Expand/Collapse element disappears when dataTreeElementColumn is set and cell is editable

danielHin:
fixes the issue olifolkerd/tabulator#3173

When editing is stopped (canceled or successful) row.reinitialize was only called in case the index of the cell is 0. 
In case the dataTreeElementColumn is set to a specific field, the Expand/Collapse element wasn't re-added to the element after finishing editing. 

I've added a quick check if the field is set to be the dataTreeElementColumn when checkForRestyle is called. 

I'm not quite sure if I correctly accessed the options and the field property regarding your code style guide (if there is any function which should be used instead of directly accessing the property). 

Cheers
Daniel


olifolkerd:
Hey @mbellagamba

Im afraid we are not currently accepting PR's on the master branch.

The library has been completely rebuilt for the 5.0 release and most of the original files are no longer there.

If you are happy to resubmit this on the 5.0 branch i would be happy to consider it for merge.

Cheers

Oli :)

danielHin:
Hey @olifolkerd 

allright. Wasn't sure if there will be still bugfix releases for 4.

I will have a look into 5.0

- Daniel

danielHin:
@olifolkerd 
rebased and updated the PR.

I had to change two additional files to get tabulator running locally. You might already fixed that on other branches?

olifolkerd:
Hey @danielHin 

Thanks for the PR, this has already been fixed in a different way as part of the pre-release testing, but thanks for your input.

Version 5.0 will be released later today!

Cheers

Oli :)

Issue #3328, opened 2021-06-03
select headerFilter - freetext filtering change in behaviour between v4.7.0 and >=v4.7.1


minceheid:
**Describe the bug**
Hello - following an upgrade, I've discovered a change in behaviour of the select headerFilter between v4.7.0 and v4.7.1. From v4.7.1 onwards, you can no longer type freetext into the text input to filter the list. I think this is mentioned in the release notes: 

> typing in the select header filter in multiselect mode is now blocked

But I don't think I'm using "multiselect" mode, as I'm only expecting one item on the drop-down to be selected - checking the docs for "multiselect", if I'm interpreting it correctly, there appears to be no way of disabling "multiselect" mode (I was expecting to see a multiselect:false mode available):

> multiselect - allow the user to choose multiple values. With this option enabled the editor will accept and return an array of values. This option can take two different values:
> true - a boolean value of true will allow the user to select as many item from the list as they like
> number - if you set an integer for this option that will set the maximum number of items the user can select, if they try and select more than this, earlier selected items will be deselected

**Tabulator Info**
Seems to be reproducable at all versions >=v4.7.1

**Working Example**
v4.7.0 - the input box on the headerFilter is enabled, allowing freetext filtering on the list
https://jsfiddle.net/hindsight/jn2kdub4/

**To Reproduce**
v4.7.1 - the input box on the headerFilter is disabled, preventing freetext filtering on the list
https://jsfiddle.net/hindsight/z2wkt0p8/

**Expected behavior**
Freetext filtering (alongside single select on the drop-down is very useful to maintain).

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser: Chrome

olifolkerd:
Hey @minceheid there is no longer a freetext entry option for the select header filter, if you would like freetext entry then you should be using the autocomplete editor.

Cheers

Oli :)

PR #3333, opened 2021-06-09
Fix issue #3328 - select headerFilter - freetext filtering change in behaviour between v4.7.0 and >=v4.7.1

minceheid:
As per #3328 

This re-enables text entry in edit.js if multiselect is not in use, by adding a conditional.

### DIFF
--- a/src/js/modules/edit.js
+++ b/src/js/modules/edit.js
@@ -1398,7 +1398,7 @@ Edit.prototype.editors = {
 				break;
 
 				default:
-				if(self.currentCell === false){
+				if(self.currentCell === false && multiselect){
 					e.preventDefault();
 				}
 


olifolkerd:
Hey @minceheid 

Im afraid this would break the functionality of the multiselect usage of this editor. 

And im afraid  we are also not currently accepting PR'son the master branch, the 5.0 release, that is currently in progress has completely refactored the code base and most of the old files are no longer there.

If you would like to resubmit for the 5.0 branch and not break the multi select functionality then i would be happy to consider another PR.

Cheers

Oli :)

minceheid:
Ah, yes - that would indeed be a problem. I must try harder :-)

Thanks, will take a look in due course

PR #3340, opened 2021-06-23
3340 Preserve vertical scroll position when adding a row to the table

VSnemo:
Added two methods for setting and setting the scroll position. This will be useful when the scroll position is set to 0 and you need to get the last position.


olifolkerd:
Hey @VSnemo 

Im still not totally sold on this as a usage case, if you are just updating the table with new data you can simply use the `refreshData` function that will allow you to add rows to the table without updating the scroll position.

The `addRow` function is designed to add a row and then call focus to it.

Cheers

Oli :)

Issue #3498, opened 2021-12-06
RTL mode, table sort caret overlaps text in columns


carsonwood1:
**Describe the bug**
When the table is using RTL mode, the sort caret and column text overlap each other.

https://codepen.io/carsnwd95/pen/rNGxQLv

**Tabulator Info**
- 5.0.7

**Working Example**
https://codepen.io/carsnwd95/pen/rNGxQLv

**To Reproduce**
Load the codepen and observe.

**Expected behavior**
I would expect the caret to switch to the left side of the column or move out of the way in some manner.

**Screenshots**
![image](https://user-images.githubusercontent.com/95372797/144914778-30c9098e-658e-4901-bfdf-fd1954597d30.png)

**Desktop (please complete the following information):**
 - Browser: Chrome

carsonwood1:
I opened a PR with a fix I believe here #3499

rowenbuzz:
@carsonwood1 Thank you for reporting, I came here to say the same discrepancy after updating Tabulator to 5.0 this week.  But then I updated my CSS to fix it back to the previous state I was expecting to see.  

I changed the CSS to the following:

```
.tabulator .tabulator-header .tabulator-col .tabulator-col-content .tabulator-col-sorter {
    position: initial;
}
```

olifolkerd:
Hey All,

Ive opted for the fix posted by @rowenbuzz as i think it is the cleaner solution.

I have pushed the fix to the master branch and will include it in the next patch release.

Cheers

Oli :)

carsonwood1:
I think setting the position to initial isn't quite what we would expect.

Doing that causes the caret to flow below the header title

![image](https://user-images.githubusercontent.com/95372797/148447354-c3c7d8c0-7caa-4dce-a08d-4f2b59fe29e4.png)

https://codepen.io/carsnwd95/pen/JjrBXVV see it here, this pen uses the current 5.0.10 CSS

Where as changing that tabulator-arrow to tabulator-col-sorter in that PR I did more aligned to what we expected on RTL for sorting.

![image](https://user-images.githubusercontent.com/95372797/148447586-089aa4ff-9370-47b1-a939-31216715c455.png)

See example here with that fix https://codepen.io/carsnwd95/pen/rNGxQLv 

I mean...I'm a native English speaker only, don't speak in an RTL language ... so I could be wrong on what is expected on RTL here with the layout.

What are your @olifolkerd @rowenbuzz  thoughts ?

[...]

PR #3499, opened 2021-12-06
Fix so caret and column label do not overlap

carsonwood1:
This is a fix I believe for the issue where a column label text and the sort caret overlap in RTL mode. Change the CSS class the aligns the caret to be `.tabulator-col-sorter` and not `.tabulator-arrow` with RTL styling.

#3498

### DIFF
--- a/src/scss/tabulator.scss
+++ b/src/scss/tabulator.scss
@@ -1125,7 +1125,7 @@ $footerActiveColor:#d00 !default; //footer bottom active text color
 			}
 
 			.tabulator-col-content{
-				.tabulator-arrow{
+				.tabulator-col-sorter{
 					left:8px;
 					right:initial;
 				}


olifolkerd:
Hey @carsonwood1 

Do you have a link to a JSFiddle that demonstrates the current issue?

Cheers

Oli :)

olifolkerd:
How does this compare to the fix proposed by @rowenbuzz on the issue?

Cheers

Oli :)

carsonwood1:
@olifolkerd 

Looks like he changed the position to be initial what the code used to be it appears.

I just updated instead of position right on the `.tabulator-arrow` , add position right to the container of the arrow `.tabulator-col-sorter` . 

It was just what I noticed and fixed looking at the CSS locally, so I wouldn't be offended if y'all went with going back to position: initial instead 👍

olifolkerd:
Thanks for the response, i think i will opt for the "initial" approach as it is a little cleaner.

Cheers

Oli :)

Issue #3505, opened 2021-12-14
datetime parse alternative


AmauryVa:
Hi,

The default string of datetime formater transforms (inputFormat) is currently "yyyy-MM-dd HH:mm:ss". We can manually set string format.
But in luxon library, there are difficulties to parse datetime ISO 8601 by the function fromFormat('datetime string', 'datetime format').
Example of ISO 8601 : "2021-09-11T00:00:00.000Z"
I expose the current problem [here](https://stackoverflow.com/questions/70356141/tabulator-5-0-parsing-datetime-luxon-datetime-iso-8601)
I suggest instead to use the luxon function fromISO when no inputFormat is defined.
You can see the documentation [here](https://moment.github.io/luxon/#/parsing?id=iso-8601).
This function can parse all those kinds of date format :
```
2016
2016-05
201605
2016-05-25
20160525
2016-05-25T09
2016-05-25T09:24
2016-05-25T09:24:15
2016-05-25T09:24:15.123
2016-05-25T0924
2016-05-25T092415
2016-05-25T092415.123
2016-05-25T09:24:15,123
2016-W21-3
2016W213
2016-W21-3T09:24:15.123
2016W213T09:24:15.123
2016-200
2016200
2016-200T09:24:15.123
09:24
09:24:15
09:24:15.123
09:24:15,123
```
I think it's a better choice than default "yyyy-MM-dd HH:mm:ss"

I tried to code it in fork but my javascript skill doesn't allow me to make a clean script.

Thank you

P.S : The part of the datetime formatting documentation is not up to date.

olifolkerd:
Hey @AmauryVa

I would not be looking to change the default behaviour of the formatter, In its current form it works well for the majority of users and changing this default behaviour would be an unnecessary breaking change for a lot of the user base.

If you feel the current mode of operation does not allow for easy pasting of the ISO format then can i suggest that your PR adds that functionality as an option to the formatter.

Cheers

Oli :)

PR #3506, opened 2021-12-14
Modify datetime parsing, new function fromISO when inputFormat is und…

AmauryVa:
Hi,

#3505 
The default string of datetime formater transforms (inputFormat) is currently "yyyy-MM-dd HH:mm:ss". We can manually set string format.
But in luxon library, there are difficulties to parse datetime ISO 8601 by the function fromFormat('datetime string', 'datetime format').
Example of ISO 8601 : "2021-09-11T00:00:00.000Z"
I expose the current problem here
I suggest instead to use the luxon function fromISO when no inputFormat is defined.
You can see the documentation here.
This function can parse all those kinds of date format :

2016
2016-05
201605
2016-05-25
20160525
2016-05-25T09
2016-05-25T09:24
2016-05-25T09:24:15
2016-05-25T09:24:15.123
2016-05-25T0924
2016-05-25T092415
2016-05-25T092415.123
2016-05-25T09:24:15,123
2016-W21-3
2016W213
2016-W21-3T09:24:15.123
2016W213T09:24:15.123
2016-200
2016200
2016-200T09:24:15.123
09:24
09:24:15
09:24:15.123
09:24:15,123

I think it's a better choice than default "yyyy-MM-dd HH:mm:ss"

Best Regards

### DIFF
--- a/src/js/modules/Format/defaults/formatters/datetime.js
+++ b/src/js/modules/Format/defaults/formatters/datetime.js
@@ -1,12 +1,18 @@
 export default function(cell, formatterParams, onRendered){
 	var DT = window.DateTime || luxon.DateTime;
-	var inputFormat = formatterParams.inputFormat || "yyyy-MM-dd HH:mm:ss";
+	var inputFormat = formatterParams.inputFormat || undefined
 	var	outputFormat = formatterParams.outputFormat || "dd/MM/yyyy HH:mm:ss";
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @AmauryVa 

I would not be looking to change the default behaviour of the formatter, In its current form it works well for the majority of users and changing this default behaviour would be an unnecessary breaking change for a lot of the user base.

If you feel the current mode of operation does not allow for easy pasting of the ISO format then can i suggest that your PR adds that functionality as an option to the formatter.

Cheers

Oli :)

PR #3513, opened 2021-12-21
Callback the cellEdited function on the definition if it exists

xploSEoF:
As per the `cellEditing` and `cellEditCancelled` callbacks, the `cellEdited` callback should be fired before the corresponding event is dispatched.

### DIFF
--- a/src/js/core/cell/Cell.js
+++ b/src/js/core/cell/Cell.js
@@ -191,6 +191,10 @@ export default class Cell extends CoreFeature{
 
 			this.cellRendered();
 
+			if(this.column.definition.cellEdited){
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @xploSEoF 

Thanks for your suggested PR.

To be consistent with the behaviour of the `cellEdited` event which is dispatched from the table, this would need to be called in the `setValue` function of the `Cell` class.

If you can make those changes i would be happy to approve the PR.

Cheers

Oli :)

xploSEoF:
Hi @olifolkerd

This PR is already where you have asked for it.

Thanks
Liam

PR #3215, opened 2021-02-18
added maxInitialWidth

rathboma:
This is different from both `width` and `maxWidth` in that it simply limits the initial width of a column, it can still be resized larger afterwards, but initially it limits the width.

This is useful when you have a lot of columns of unknown sizes and you want to make things initially look pretty good

Still needs:
- Typings adding
- Website docs updating

I don't know how to do those

PR #3519, opened 2021-12-27
5.0 max initial width and edit module resolution fix

rathboma:
Max Initial Width is similar to Max Width except that it only applies on first render, so it allows the user to make columns larger (up to the actual maxWidth).

5.0 replacement for #3215

Also fixes a bug with edit module resolution

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*]
+indent_style = tab
+end_of_line = lf
+insert_final_newline = true
--- a/src/js/core/column/Column.js
+++ b/src/js/core/column/Column.js
@@ -42,6 +42,7 @@ class Column extends CoreFeature{
 		this.widthStyled = ""; //column width prestyled to improve render efficiency
 		this.maxWidth = null; //column maximum width
 		this.maxWidthStyled = ""; //column maximum prestyled to improve render efficiency
+		this.maxInitialWidth = null;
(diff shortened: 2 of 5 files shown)


rathboma:
@olifolkerd sorry this took so long. Let me know what you think

rathboma:
this is not done yet actually

rathboma:
should be good now (old `self` reference)

olifolkerd:
Hey @rathboma 

Thanks for the PR,

The idea looks good, but there are a couple of bits it would be good to resolve before it gets merged in *(it is possible that several of these have cropped up due to changes on master since your initial PR)*:

- You have included the tabulator.js `/dist` file in the PR, PR's should only include files in the `/src` folder. There are many different types of dist file generated so to save on them being overwritten with each PR, they are are only committed when the version number is incremented in a release.
- The indentation of several of the lines does not seem to match the surrounding code
- 
Cheers

Oli :)

rathboma:
@olifolkerd Fixed. This should be good to review now.

I also added a `.editorconfig` so that the IDEs of contributors will automatically use tabs :-)

olifolkerd:
great idea on the .editorconfig, i will make a note of that in the release notes.

Cheers

Oli :)

olifolkerd:
i will also be removing the columnMaxInitialWidth option, as now we have the `columnDefaults` option that should now be set there to be globale.

Cheers

Oli :)

PR #3236, opened 2021-03-08
Allow JSON downloader to use `titleDownload`

rathboma:
fixes #3235

PR #3520, opened 2021-12-28
Json downloader can use titleDownload

rathboma:
This allows you to override the key in the json objects.

Why is this useful?

In many cases the fields are abstract, and a user friendly key needs to be substituted.

For example in Beekeeper Studio two columns in a table can have the same title, so I need a way to make fields unique, but also provide a title for the download.

Replaces #3236

### DIFF
--- a/src/js/core/column/Column.js
+++ b/src/js/core/column/Column.js
@@ -30,6 +30,7 @@ class Column extends CoreFeature{
 		this.getFieldValue = "";
 		this.setFieldValue = "";
 
+    this.titleDownload = null;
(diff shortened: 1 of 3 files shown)


rathboma:
@olifolkerd sorry this one also took some time.

olifolkerd:
Hey @rathboma 

On the whole this looks good, but there seems to be an indentation issue with several of the new lines of code, they do not match the surrounding lines indentation.

If you can fix that i will happily merge this in.

Cheers

Oli :)

rathboma:
@olifolkerd this is fixed

PR #3549, opened 2022-01-21
Fix uncaught exception in progressive scroll

saleyn:
Fix leaking exception when trying to scroll past the last page in progressive scroll

### DIFF
--- a/src/js/modules/Page/Page.js
+++ b/src/js/modules/Page/Page.js
@@ -353,7 +353,7 @@ class Page extends Module{
 			});
 
 			this.nextBut.addEventListener("click", () => {
-				this.nextPage()
+				this.nextPage().catch(() => {})
 			});
 
 			this.lastBut.addEventListener("click", () => {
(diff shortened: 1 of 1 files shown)


rathboma:
Are there other reasons this could error? Is it worth adding a `console.warn`?

saleyn:
> Are there other reasons this could error? Is it worth adding a `console.warn`?

As I understand line 737, it would always happen when there's an attempt to scroll past the last page.  That is the behavior I've been getting, which let to this PR.  I am not sure if logging warning in that case would be any useful, why would it?

olifolkerd:
@saleyn surely rather than catching the error that occurs, the best approach would be to check the current page number before calling next and then not make the request if you are on the last page?

That way we are not obscuring any other potential errors (for example f the request failed

Cheers

Oli :)

olifolkerd:
This root cause of this issue was resolved in one of the 5.1 patch releases. So I'm gonna close this PR

PR #3557, opened 2022-01-25
cellEdited callback was not firing. It is now.

ivo-bloechliger:
The cellEdited callback was not firing. It is now.
I have to admit, that I don't understand the code in depth, but it works for me now (TM).

But in any case: Many thanks for providing such a useful and well documented library!


ivo-bloechliger:
To clarify: cellEdited was not firing when specifying it as follows:
`{title:"Name", field:"name", cellEdited:function(cell){
    //cell - cell component
    },
}`

olifolkerd:
thi is a duplicate of #3513 which has now been merged into master and will be included in sundays release

PR #3587, opened 2022-02-09
Allowing empty string as thousand separator in money formatter.

harbulot:
This is a feature request to be able to have an empty string as the thousand separator (`money` formatter).

At the moment, using `thousand: ""` in `formatterParams` is effectively ignored, since it falls back to the default separator (`,`).

This patch makes it possible not to use a thousand separator at all (i.e. use an empty string).

### DIFF
--- a/src/js/modules/Format/defaults/formatters/money.js
+++ b/src/js/modules/Format/defaults/formatters/money.js
@@ -25,10 +25,12 @@ export default function(cell, formatterParams, onRendered){
 	integer = number[0];
 	decimal = number.length > 1 ? decimalSym + number[1] : "";
 
-	rgx = /(\d+)(\d{3})/;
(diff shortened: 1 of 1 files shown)


harbulot:
Sure, here it is!

I guess I thought the advantage of `""` was that it didn't need to be documented as a special case (since it would do exactly what's intended with that string parameter value), but I also understand your reasoning behind using `false`.

Thank you!

olifolkerd:
Thanks for making those updates. This looks good. I will approve it for merge and include it in the 5.2 release which should be comming out in april

olifolkerd:
@harbulot  if you are on the tabulator discord server, drop me a private message and i will give you a contributor role so other can see you have contributed to Tabulator

PR #3594, opened 2022-02-14
Move rows on row drag

adarshk:
What this PR does:
When a row is dragged, scroll to top or bottom i.e. enables a more interactive drag. This addresses https://github.com/olifolkerd/tabulator/issues/3249#issuecomment-1036623968

Some issues that remain after applying this PR: 
1. the placeholder element is only available when dragged over certain number of elements beyond which it doesn't move anymore. My guess - is this due to `renderVerticalBuffer`? I am not sure.
2. The scroll only happens when when moving the mouse, not when _holding_ it at the bottom position.
3. Need to test touch events

I need help with the issues mentioned above
I am happy to help further but also would like feedback as I may not have all the context. Currently this is showing the potential changes that are required to enable a more interactive row drag


azmy60:
Closing this as I couldn't replicate on `v6.3.1`.

PR #3631, opened 2022-03-01
Fixed Footer Placement

larsvanherwijnen:
What does this PR:

After version 5.1 the footer didn't work anymore this was partly fixed in version 5.1.3 but you stil not able to place the footer in your own div, therefore I created this PR in order to fix this issue.

You can now do the following:
image

Before it would have looked like this:
image

You can see that on the second image that the pagination is within the tabulator (grey div). You might need the open the images in a second tab because it is not realy clear on the screenshots.

### DIFF
--- a/src/js/core/FooterManager.js
+++ b/src/js/core/FooterManager.js
@@ -44,7 +44,7 @@ export default class FooterManager extends CoreFeature{
 					this.containerElement.innerHTML = this.table.options.footerElement;
 				}else{
 					this.external = true;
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @larsvanherwijnen 

Good catch, i will merge this in now and include it in the weekends patch release.

Cheers

Oli :)

olifolkerd:
@larsvanherwijnen  if you are on the tabulator discord server, drop me a private message and i will give you a contributor role so other can see you have contributed to Tabulator

Issue #3671, opened 2022-03-21
Changing the filter or sorting, while pagination is enabled, does not revert the table back to the first page


KarmaProd:
**Describe the bug**
When using pagination, after changing a filter or sorting, the tabulator page is not restored to first one. Please note that documentation states otherwise, and this behaviour was working in a previous version.

**Tabulator Info**
5.1.7

**Working Example**
example of bug: http://tabulator.info/docs/5.1/page
example of working version: http://tabulator.info/docs/4.1/page

**To Reproduce**
A step by step guide to recreate the issue:
1. Go to http://tabulator.info/docs/5.1/page and move to the very first table (Overview)
2. Click on button "2" on the bottom right to move to page two.
3. Click on Rating in header to sort by rating
4. Page is not restored to the first one.

**Expected behavior**
Page should be restored to page 1, as the documentation clearly states shortly after in that same page:
`Note: Changing the filter, sorting or toggling a groups visibility, while pagination is enabled will revert the table back to the first page.`

**Desktop:**
 - OS: Windows 10
 - Browser: Google Chrome
 - Version: 99.0.4844.74

doublehrajput:
Compare #3676 these changes .. I am assuming that it might be also cause problem in other Modules as well

olifolkerd:
@doublehrajput Unfortunately those changes will result in a reset in horizontal scroll position, which is undesirable when not in paginated mode.

olifolkerd:
Hey @KarmaProd 

Thanks for point out the inconsistency here. As much as this is a regression against the previous version. i think it actually highlights a bigger problem with the current design paradigm.

At the moment scroll positions are preserved when a sort occurs (on local data) but not pagination. which is making the same action trigger different actions per module.

To make things more consistent i will make some improvements in the next update. all modules will maintain their current state on sort/filter and then i will add a `sortPositionReset` and `filterPositionReset` *(or similar)* setup option to allow control over whether this resets the table or not.

Cheers

Oli :)

KarmaProd:
Actually, I didn't mention something I noticed. In my stable installation I use tabulator 5.0.7, and this bug is NOT present. But in the http://tabulator.info/docs/5.0/page documentation page the bug IS present. @olifolkerd do you know why?

olifolkerd:
Because that uses local pagination not remote pagination, the issue isnt about pagination as such but the data source. a remote data source will result in data being reloaded a local source will result in it being refreshed

[...]

PR #3676, opened 2022-03-25
Issue #3671 -  Revert Back to first page on changing filter or sorting

doublehrajput:
As per previous release  `this.refreshData()` must pass without `true` param

### DIFF
--- a/src/js/modules/Edit/defaults/editors/select.js
+++ b/src/js/modules/Edit/defaults/editors/select.js
@@ -480,6 +480,7 @@ export default function(cell, onRendered, success, cancel, editorParams){
 		if(!input.value){
 			unsetItems();
 			chooseItems();
+			showList();
 		}
 	});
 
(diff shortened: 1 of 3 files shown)


olifolkerd:
Im afraid that while this PR addresses the issue you have mentioned it will break other functionality because it will reset the horizontal scroll position, which is undesirable.

I will include the correct fix in the next patch release.

Cheers

Oli :)

olifolkerd:
Im afraid that while this PR addresses the issue you have mentioned it will break other functionality because it will reset the horizontal scroll position,in non paginated tables which is undesirable.

In this case we are only trying to address that the page should be reset specifically on the action of sorting/filtering not that any of the rest of the render pipeline used by other modules should be changed.

In order to keep the modules isolated, you would need to dispatch an event that gives other modules the choice of whether to block the refresh in these scenarios, or to trigger them to make their own changes to their internal state before the refresh is completed.

Cheers

Oli :)

olifolkerd:
Thinking about this further i actually think there is a bigger issue to be addressed here as the behaviour between remote and local sorting/filtering is different.

I actually think this should be a new feature that we define an new prop that determines whether sorting/filtering results in redraw in position or a full reload.

This would give the users much more control over how this functions and ensure a more consistent behaviour

PR #3683, opened 2022-03-31
For formatter and sorter, accept datetimes that are already Luxon DateTime objects (vs. strings)

JiCiT:
Currently, the built-in formatter and sorter are "unaware" when datetime data is Luxon DataTime objects,  Instead, to use the built-in formatter and/or sorter, the data must 1) be converted to a string which is then 2) re-converted back into a DataTime object and, finally, 3) converted once again to a string.

This change allows use of DateTime objects directly bypassing steps 1 & 2.  It does introduce a low-cost check to determine if the data is a DataTime object.

### DIFF
--- a/src/js/modules/Format/defaults/formatters/datetime.js
+++ b/src/js/modules/Format/defaults/formatters/datetime.js
@@ -6,7 +6,15 @@ export default function(cell, formatterParams, onRendered){
 	var value = cell.getValue();
 
 	if(typeof DT != "undefined"){
-		var newDatetime = DT.isDateTime(value) ? value : ( inputFormat === "iso" ? DT.fromISO(String(value)) : DT.fromFormat(String(value), inputFormat) );
+		var newDatetime;
+
+		if(DT.isDateTime(value)){
+			 newDatetime = value;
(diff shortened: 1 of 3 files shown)


olifolkerd:
Hey @JiCiT 

That is a great idea!

Though as it has made that logic a bit more complex, would you mind breaking the if out onto multiple lines, jamming multiple ternary ifs onto one like hurts the code clarity a bit.

Cheers

Oli :)

JiCiT:
@olifolkerd I changed to if/else statements for clarity, I hope that's OK?

olifolkerd:
@JiCiT Looks perfect!, i will flag this as approved and include it in the 5.2 release later this month,

Cheers

Oli :)

olifolkerd:
Just a quick note following the merge, the sorter code had an error in it, it was using the `inputFormat` variable instead of the `format` variable.

No Biggy, i just released a patch for it, but worth mentioning for your next PR.

Cheers

Oli :)

olifolkerd:
@JiCiT  if you are on the tabulator discord server, drop me a private message and i will give you a contributor role so other can see you have contributed to Tabulator

PR #3729, opened 2022-04-28
Fix data validation logic in v5

### LINKED ISSUE #3206: cell.validate & cell.isValid to return validators failing
**Describe the solution you'd like**
When we **cell.validate()** the return is either _true or _false_ . In case of validation violation, we do not know what **validators** have failed (like we know upon cell edition using the callback), so it will very useful that **cell.validate()** when validation fails, to return the same as when editing.

In the same line of thought, when calling **cell.isValid()**, in case on **_not valid_**, should return the validators that didn't passed.

An option may pass to store the on the cell Component the failed validators information and implement a function **cell.getViolations()** which return:  { value: (cell value that failed validation), violations : [ (validator objects that failed) ] }

**Describe alternatives you've considered**
The alternative is "faking" cell edition to trigger the valididationFailed callback, but if we are validating an entire row or column or even table, this isn't a good solution :-(

Thanks a million 

P.S: if this can be part of a 4.9.x would be great :-)

cheers
João

mcaskill:
Before 5.0, the `Cell#validate` method would return a boolean (`true` if valid, `false` if invalid).

Since 5.0, cell, row, and column validation [has been moved](https://github.com/olifolkerd/tabulator/commit/f622cf1e16d3452486b4013a68aaf7e3890216a8) to the Validate module.

The `Validate#cellValidate` method returns `true` if the cell is valid and an array of failed validators if the cell is invalid.

Unfortunately, the `Validate#columnValidate` and `Validate#rowValidate` methods still expect `Validate#cellValidate` to return either a _truthy_ value if valid and a _falsy_ value if invalid.

This pull request corrects the inconsistent logic.

This resolves #3206 and potentially other issues.

### DIFF
--- a/src/js/modules/Validate/Validate.js
+++ b/src/js/modules/Validate/Validate.js
@@ -89,7 +89,7 @@ class Validate extends Module{
 		var invalid = [];
 		
 		column.cells.forEach((cell) => {
-			if(!this.cellValidate(cell)){
+			if(this.cellValidate(cell) !== true){
 				invalid.push(cell.getComponent());
 			}
 		});
@@ -105,7 +105,7 @@ class Validate extends Module{
 		var invalid = [];
 		
 		row.cells.forEach((cell) => {
-			if(!this.cellValidate(cell)){
+			if(this.cellValidate(cell) !== true){
 				invalid.push(cell.getComponent());
 			}
 		});


olifolkerd:
Good catch. I will include this in the next patch release!

olifolkerd:
This will be included in this afternoons patch release

olifolkerd:
@mcaskill if you are on the tabulator discord server, drop me a private message and i will give you a contributor role so other can see you have contributed to Tabulator

PR #3751, opened 2022-05-17
add missing js extension

lekoala:
otherwise the esm build won't work if used directly in the browser


olifolkerd:
Hey @lekoala 

Good spot, I will include this in the next patch release.

Cheers

Oli :)

PR #3752, opened 2022-05-17
update table size on collapse toggle

lekoala:
without this, if you don't use a fixed height, a scrollbar will appear when opening the responsive columns because in RowManager there is a height set this.element.style.height = (this.table.element.clientHeight - otherHeight) + "px";

alternative option is in RowManager to use a minHeight style instead, that allows the table to grow automatically and let the browser handle it. This PR is probably more aligned with the current practices of Tabulator that tends to compute things and not let the browser determine size itself.


olifolkerd:
Hey @lekoala 

Thanks for the feedback but this is definitely not the correct approach to resolving this issue. the adjustTableSize function should only be triggered by elements outside the table holder that then change the available size inside the row holder.

This is a more complex issue tied into numerous parts of the table and will be heavily dependent on your table setup as to how it presents and what the solution may be.

Please could you create an issue with a JS fiddle that demonstrates your setup and i can look into a more appropriate fix.

Cheers

Oli :)

lekoala:
@olifolkerd ok understood. My PR is indeed maybe not the right approach (although it does solve the issue). If I understand correctly, the table height is mostly necessary only when using virtual dom. But that seems really the primary use case (although, in my case, i'm perfectly happy to render all rows and let the browser deal with the actual size of the table). Even the "basic" renderer as a verticalMode "fill" which will set a height anyway. Is there a way to let the browser compute everything itself?

I've opened a distinct issue for this with the jsfiddle that demonstrate the issue
https://github.com/olifolkerd/tabulator/issues/3771

olifolkerd:
Hey @lekoala 

On approaching this a-fresh i have realised i completely missunderstood the issue you were trying to fix there, so my sincere apologies for the mix up.

In this instance your proposed fix should do the job nicely. i will reopen and merge

Cheers

Oli :)

Issue #3745, opened 2022-05-13
Header list filter with option multiselect triggers error


lekoala:
**Describe the bug**
When setting the option multiselect to true, there is an error 'List.js:73 Uncaught TypeError: this.initialValues.join is not a function'

**Tabulator Info**
5.2.4

**Working Example**
https://codepen.io/lekoalabe/pen/PoQPYVp

**To Reproduce**
go to the codepen and uncomment the line 14, on reload the table fails and throw the error.

**Expected behavior**
The list is displayed

**Additional context**
I tried various type of ways for values (just values, key/values, value/labels ...)
I also tried with an array as empty or default value, but that doesn't help

I think the issue is simply that the script expects the initialValue to be an array with multiselect true when clearly this is not always the case.

lekoala:
happy to help or propose a PR if needed :)

lekoala:
https://github.com/olifolkerd/tabulator/pull/3753

olifolkerd:
Hey @lekoala 

I have pushed a fix for this to the master branch and will include it in the next patch release.

Cheers

Oli :)

PR #3753, opened 2022-05-17
fix multiselect

lekoala:
getValue never returns an undefined value, therefore we check for an empty string instead. Since we need to join in all cases (not only for multiselect = true), if the initialValue is a string, we convert it to an array. As a safety measure, don't join if there is no initialValues (shouldn't be necessary but that never hurts)

### DIFF
--- a/src/js/modules/Edit/List.js
+++ b/src/js/modules/Edit/List.js
@@ -73,7 +73,7 @@ export default class Edit{
         this.initialValues = initialValue;
 
         if(this.isFilter){
-            this.input.value = this.initialValues ? this.initialValues.join(",") : [];
+            this.input.value = this.initialValues.join(",");
             this.headerFilterInitialListGen();
         }
     }
--- a/src/js/modules/Edit/List.js
+++ b/src/js/modules/Edit/List.js
@@ -62,19 +62,16 @@ export default class Edit{
     
     _initializeValue(){
         var initialValue = this.cell.getValue();
-
-        if(initialValue === "" && typeof this.params.defaultValue !== "undefined"){
+        
+        if(typeof initialValue === "undefined" && typeof this.params.defaultValue !== "undefined"){
             initialValue = this.params.defaultValue;
         }
-
-        if(typeof initialValue === "string"){
-            initialValue = [initialValue];
-        }
-        this.initialValues = initialValue;
-
+        
+        this.input.value = this.params.multiselect ? this.initialValues : this.initialValues.join(",");
+        
         if(this.isFilter){
             this.input.value = this.initialValues.join(",");
-            this.headerFilterInitialListGen();
+            this.headerFilterInitialListGen();            
         }
     }
     
--- a/src/js/modules/Edit/List.js
+++ b/src/js/modules/Edit/List.js
@@ -62,16 +62,19 @@ export default class Edit{
     
     _initializeValue(){
         var initialValue = this.cell.getValue();
-        
-        if(typeof initialValue === "undefined" && typeof this.params.defaultValue !== "undefined"){
+
+        if(initialValue === "" && typeof this.params.defaultValue !== "undefined"){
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @lekoala 

Could you explain what the purpose of this PR is. i am unsure from your description what the usage case is or what problem it is solving?

Cheers

Oli :)

lekoala:
it fixes this issue
https://github.com/olifolkerd/tabulator/issues/3745

as demonstrated here:
https://codepen.io/lekoalabe/pen/PoQPYVp
(uncomment L14)

tomalexander73:
Hello.  I'm having the same issue, but after debugging it locally, I noticed that at this line in the code:
https://github.com/olifolkerd/tabulator/blob/ef7485e98ce1d6184c69867b5e86db9dac909b86/src/js/modules/Edit/List.js#L73

The value of this.initialValues is "" - so running join() on that fails.

If I change to this:

`this.input.value = this.params.multiselect ? this.initialValues : this.initialValues.join(",");`

...it works as expected.  It looks up the values in my column and populates the select list.  Does this make sense?

olifolkerd:
@tomalexander73 Thanks for the feedback, i agree, initializing the value correctly in the first place would be a much simpler option that reconfiguring all the output logic.

@lekoala Are you happy to make the proposed changes to simplify your PR.

Cheers

Oli :)

lekoala:
@olifolkerd I think my approach is more robust overall, since the goal is to always make sure initialValues is an array. In the current state, you could still end up with an error because if multiselect = true and you passed a string as the default value (eg: "val1,val2"), it will break. Basically, there is no relation between initialValues and the state of multiselect (it's plural, and should always be an array that can be joined into a string for the editor).

my approach also remove conditions that are never met (getValue == undefined, that's never true) and therefore, defaultValue is never used.

so, no, i'm not going to change my PR :-) but obviously, that's just one line of code, so if you want to make a partial fix for now, you can always do that yourself.

tomalexander73:
@lekoala It was 2am when I was debugging so my thinking was a little cloudy.  I see your point, that if you want to pass a comma separated string as the initial values, then yes, my quick-fix wouldn't work.  I think I was assuming the multi-select option required an array.  If I get over my current deadline, I will try to help out more if I can - this is a fantastic library overall.

olifolkerd:
Hey @lekoala 

Appologies, i should have been clearer in my last response, i was typing on the phone while walking.

The `getValue` function absolutely can return an undefined value, if a cell does not have a value set on it then the function will return a type of undefined. that is specifically why the logic is there to handle that value.

In Tabulator, header filters use editors, but those are just standard editors that are mainly used on cells, but also repurposed to be used in a header filter environment. In the case of them being used in cells, your PR would break the existing functionality where the logic is specifically built to handle the scenario of a cell with an undefined value.

The `isFilter` prop on the list class can be used to determine if the editor is being used in the context of a header filter, so you should look leave the existing undefined logic in place for the cell and add your additional checks if it is a filter.  By convention editors are treated as cell editors first and foremost with any extra logic needed for filters being included after.

Cheers

Oli :)

olifolkerd:
@tomalexander73 Thanks for your kind words about Tabulator, it is always great to hear people are enjoying using the library. If you are interested in contributing more, please head over to the [Community Section](http://tabulator.info/community) which will point you in the direction of the areas we are most looking for help for at the moment.

Cheers

Oli :)

lekoala:
@olifolkerd wow I swear I thought I've seen somewhere that getValue returned "", but... it seems I'm wrong. Not sure where I've seen that...
so indeed, it's better to keep things as they are. I've simplified my PR to implement @tomalexander73  changes

olifolkerd:
Hey @lekoala 

On investigating the only change needed was to tweak the output being passed into the input in the isFilter conditional.

I have pushed a fix of this to master and will include it in the next patch release.

Cheers

Oli :)

PR #3759, opened 2022-05-18
fix null warnings

lekoala:
typeof never returns "null"
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null

while it's not a big issue (it's never going to match, and it's handled in the "object" case), it's better to keep the code clean
i caught this while trying to compile the source with esbuild

other option is to do something like var check = value === null ? "null" : typeof value;

### DIFF
--- a/src/js/core/cell/Cell.js
+++ b/src/js/core/cell/Cell.js
@@ -118,7 +118,6 @@ export default class Cell extends CoreFeature{
 			}
 			break;
 			case "undefined":
-			case "null":
 			this.element.innerHTML = "";
 			break;
 			default:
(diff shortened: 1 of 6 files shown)


olifolkerd:
Hey @lekoala 

I can see your point on the null there, but using a false comparison on the  value for a ternary if could cause issues with other values. please change this to an explicit check for null and i would be happy to accept the PR

Cheers

Oli :)

lekoala:
@olifolkerd done!

olifolkerd:
This will be included in the next patch release.

Cheers

Oli :)

PR #3761, opened 2022-05-19
add class for editable cell

lekoala:
a class for editable cells allow to drive some ui features (like custom background on hover, or custom cursor). otherwise it's not always easy to know in advance which cells are editable for a given table.


olifolkerd:
Hey @lekoala 

Thanks for the PR, It is a nice idea, but cells being editable is a bit more complex that this PR considers.

Not all cells with editors are editable, the 'editable' column definition option can be used by a callback to determine when a particular cell with an editor is editable. this is done in real time when the cell is clicked on.

How would you suggest handling this scenario?

There is already a 'cssClass' option for each column, could this not be used to add the classes you need to specific columns without any additional functionality?

Cheers

Oli :)

lekoala:
@olifolkerd yes indeed, i've discovered the "column.definition.editable" later on :-). In my case, it's quite simple, in the data of the row, i have a _editable field that is set to true or false, and therefore, it's quite predictable. It's true that with a fully dynamic function, there is no way to know in advance until you click if the cell is going to be editable or not, but i expect these cases are quite rare (and if you really need dynamically enabled cells, well then.. don't use the css class :) ).

the issue with using cssClass, is that you don't know in advance what's going to be in the dataset, so that's not an option. i mean, there a ways you could add this, but not as convenient.

ideally, i'd like to push this PR a bit more and set the editable class based on the actual check. so if it's a boolean, set if true and if it's a string (=> new feature), check the row data and look for a true boolean value.

what do you think ?

lekoala:
@olifolkerd you are quite right, i've improved the logic here https://github.com/olifolkerd/tabulator/pull/3761/commits/b3c0e740b9c11405b8da82aa3a9b30a0b613efbf
while at it, i've created a central allowEdit function that can be used elsewhere

lekoala:
done!

olifolkerd:
Fantastic stuff, i will schedule this for inclusion in the next minor release.

Cheers

Oli :)

lekoala:
@olifolkerd fyi, i just made a small update to this. actually, if the cell is changed, it might make the cell not editable anymore and therefore it's better to subscribe to the event and update the class accordingly.
i'm even wondering if it might be necessary to listen to row data changed, but in my case, what i did is simply forcing a change of value on that cell to trigger the update

you can see the changes here https://github.com/olifolkerd/tabulator/pull/3761/commits/001948a6f2bba6abee780e79074668a230a9d77e

olifolkerd:
Hey @lekoala,

Good catch there, That is a good addition to the PR.

Cheers

Oli :)

PR #3765, opened 2022-05-20
add extra tabulator-table-editing class

lekoala:
this allows to handle some complex scenarios where you would want for instance to:
- show editable cells on hover
- but remove the hover effect if you moved to another row through keyboard nav

Example css to leverage this:

```css
.tabulator:not(.tabulator-table-editing)
    .tabulator-row:hover
    .tabulator-cell-editable,
.tabulator.tabulator-table-editing
    .tabulator-row-editing
    .tabulator-cell-editable {
    background: var(--tabulator-highlight-color);
}
```

### DIFF
--- a/src/js/modules/Edit/Edit.js
+++ b/src/js/modules/Edit/Edit.js
@@ -411,6 +411,7 @@ class Edit extends Module{
 			while(cellEl.firstChild) cellEl.removeChild(cellEl.firstChild);
 
 			cell.row.getElement().classList.remove("tabulator-row-editing");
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @lekoala 

Nice idea, i will look to include this in the next minor release.

Cheers

Oli :)

olifolkerd:
Hey @lekoala 

I have merged this in to the 5.3 branch and will include it in the next release. To keep things consistent as the `tabulator-editing` class is applied to a cell that is being edited, i will also apply the same class to the table.

Cheers

Oli :)

PR #3766, opened 2022-05-20
properly deal with null and undefined

lekoala:
the current condition does not work as intended because of the || clause that use strict !== (null or undefined is always !== than input.value)

(cellValue === null || typeof cellValue === "undefined") && input.value !== "") || input.value !== cellValue

if cellValue is null, when entering the input then leaving it, it will mark the field as "edited" when in fact there was nothing done (there is no way to distinguish "null/undefined" from blank "" in the input field). Therefore, clicking the first time creates an "edited" callback, and next time, and "cancelled" callback.

this PR fixes this by treating null, undefined and "" as "" by default.

another enhancement would be to add a editorParam to allow "null" as a text value that should be converted to actual "null" value. This could be useful for header filters that should match actual "null" db values for example, but that's outside the scope of this PR.

### DIFF
--- a/src/js/modules/Edit/defaults/editors/input.js
+++ b/src/js/modules/Edit/defaults/editors/input.js
@@ -6,6 +6,10 @@ export default function(cell, onRendered, success, cancel, editorParams){
 	var cellValue = cell.getValue(),
 	input = document.createElement("input");
 
+	if((cellValue === null || typeof cellValue === "undefined")) {
+		cellValue = "";
+	}
+
 	input.setAttribute("type", editorParams.search ? "search" : "text");
 
 	input.style.padding = "4px";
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @lekoala 

These changes will break some of the existing functionality of the editor. By removing the conversion to a "" when the value is set, it will display a text equivalent in the output, which is an undesirable function for the item.

The check on a user changing a value to an empty string triggering an edit is intentional, as if a user wants to use validation on edit then it would need to consider the change from a null to an empty string. this was added in for a very specific reason.

If you would like different functionality then please free to add a param to handle it but i will not be looking to accept these changes in their current form.

Cheers

Oli :)

PR #3767, opened 2022-05-20
consistent value for elementField

lekoala:
since firstCol can exist but firstCol.field can be undefined, it's better to check for firstCol.field also and ensure elementField is either "false" or is a string (better than false, undefined or string)

in my opinion, elementField should rather start at "null" rather then false since it's a nullable string, and I'm happy to make the change if needed ;-)

### DIFF
--- a/src/js/modules/DataTree/DataTree.js
+++ b/src/js/modules/DataTree/DataTree.js
@@ -143,7 +143,7 @@ class DataTree extends Module{
 	initializeElementField(){
 		var firstCol = this.table.columnManager.getFirstVisibleColumn();
 
-		this.elementField = this.table.options.dataTreeElementColumn || (firstCol ? firstCol.field : false);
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @lekoala 

Does this fix an actual issue with Tabulator. I am not looking for opinion based PR's otherwise I would spend all my time dealing with opinion wars between developers. If it does, then as requested with the previous PR's please provide these practical details.

Please only submit PR's if you have found an actual issue using Tabulator and are presenting a fix for it. I am not looking for people to express their own coding preferences through PR's.

I need to be honest, at the moment it is starting to feel like you are pouring over the source and nitpicking in places. and while I am very appreciative of your feedback, and it is great that you are keen to contribute to Tabulator, I would much prefer you focused on dealing with actual issues affecting you. Please have a read of the [Community Guidelines](http://tabulator.info/community/pr) as they will point you in the right direction.

Cheers

Oli :)

lekoala:
@olifolkerd no worries, I will focus on actual issues

PR #3830, opened 2022-06-29
Update progress.js to introduce hozAlign: "right" as formatter param

dbauszus-glx:
hozAlign: "right" as option for the progress bar formatter params.


olifolkerd:
Hey @dbauszus-glx 

Thanks for the PR.

Given that there is already a `hozAlign` option for a column, and is purpose is to align the contents of the column, there does not seems to be a need for this as a feature.

I can see from your issue that there was a bug in the alignment of this formatter, but the correct approach to resolve this is to fix the underlying bug (which i have just done) rather than trying to patch over it with duplicated functionality that could confuse the user base, and which would not be compatible with other table settings.

Cheers

Oli :)

PR #3842, opened 2022-07-06
update min width if set on table element

lekoala:
when there is a min width set by renderEmptyScroll, that min width may not take into account the scrollbars. Therefore, it will cause a horizontal scrollbar when rendering actual data causing the table to overflow.
adjusting the min width avoid this effect

i didn't manage yet to make a codepen to show this, but these lines effectively fix the issue in my project and i don't think they should have any side effect (if no min width is set, it doesn nothing, so it applies only if renderEmptyScroll was called in the first place)

### DIFF
--- a/src/js/modules/Layout/defaults/modes/fitColumns.js
+++ b/src/js/modules/Layout/defaults/modes/fitColumns.js
@@ -101,6 +101,11 @@ export default function(columns){
 		totalWidth -= this.table.rowManager.element.offsetWidth - this.table.rowManager.element.clientWidth;
 	}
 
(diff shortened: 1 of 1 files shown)


lekoala:
Agreed, this may not be the best place to fix it, just wanted to share with you the solution I found
If you can find a better place to fix it, I'm happy to close this

You can see the bug here
https://github.com/olifolkerd/tabulator/issues/3844

olifolkerd:
Thanks for rasing that as an issue, i will close this and fix things through the issue.

Cheers

Oli :)

PR #3855, opened 2022-07-12
Allow save and loading of persistence for frozen columns

viicslen:
Save and load in the persistence data whether the column is frozen or not. This will allow for some cases where the developer will want to give the user the option of what column to freeze.

E.g.
image


olifolkerd:
Hey @viicslen 

You can already do this using the standard configuration options, there is no need to change the default behaviour:

```js
var table = new Tabulator("#example-table", {
    persistence:{
        columns: ["width", "visible", "frozen"], //persist changes to the width, visible and frozen properties
    }
});
```

You can include any column definition property you like in the array and that value will be persisted.

Full details can be found in the [Persistence Documentation](http://tabulator.info/docs/5.3/persist#columns)

Cheers

Oli :)

viicslen:
Ohh, sorry about that, I somehow missed that part. Thank you for your prompt response

Issue #3749, opened 2022-05-16
Editable collapsed data


lekoala:
*Is your feature request related to a problem? Please describe.**
I'm working on an editable table with a couple of editable fields. I'd like this to work with the collapsed responsive mode. But any column with an editor is not editable because of the way the collapsed table is generated.
Additionally, if data is updated through the editor (let's say column A updates a value in cell B) collapsed data is not refreshed.

**Describe the solution you'd like**
It would be great if there was a way to clone the cell editor in the collapsed view. Effectively, editing through the collapsed view or through the regular view should have the same effect.

**Describe alternatives you've considered**
I had a look at the responsiveLayoutCollapseFormatter but that seems too limited to do this properly since you don't get the actual cell.

If you are interested and have some ideas on how to achieve this i'd be happy to work on it

lekoala:
https://github.com/olifolkerd/tabulator/pull/3755 this is to fix the generated data not being updated after edit

olifolkerd:
do you have a JS fiddle that demonstrates this? :)

lekoala:
yes i had to work a bit to make it work ;-)
https://jsfiddle.net/5vnu6901/5/
it doesn't include the editor, but you will see the general idea there

you can check the changes in my patch-5 branch
https://github.com/lekoala/tabulator/tree/patch-5

lekoala:
updated the fiddle with editors and highlighted cells
https://jsfiddle.net/c8gp059z/
(sorry it's a public fiddle so i had to change the link)

lekoala:
@olifolkerd what do you think of the latest jsfiddle ? Does it seem like something you would consider to include? The general approach is still a bit rough, but I think I've demonstrated its benefit (using actual columns, no need for generating responsive data...) and it is the only (easy) way to make editors work for responsive columns
Or if you can find a way to abstract that logic so that it could be implemented in an external plugin that works for me too.

olifolkerd:
I havnt had a chance to look into this one yet, I am planning to have a Tabulator day next weekend and will catch up with it then.

Cheers

lekoala:
@olifolkerd with the work on 5.3 being done, what do you think about this ? it seems safe to introduce (since it's a new responsive mode without side effects) if you are open to this approach

olifolkerd:
Hey @lekoala 

I like the concept of editable cells in the responsive collapse view.  and the editors in your example seem to work well.

[...]

PR #3875, opened 2022-07-25
Editable responsive data

lekoala:
This is to make discussion around https://github.com/olifolkerd/tabulator/issues/3749 easier

It is not meant to be merged as is

### DIFF
--- a/src/js/modules/Format/defaults/formatters/responsiveCollapse.js
+++ b/src/js/modules/Format/defaults/formatters/responsiveCollapse.js
@@ -27,6 +27,7 @@ export default function(cell, formatterParams, onRendered){
 	el.addEventListener("click", function(e){
 		e.stopImmediatePropagation();
 		toggleList(!config.open);
+		cell.getRow()._row.dispatch("row-responsive-toggled", cell.getRow(), config.open);
 		cell.getTable().rowManager.adjustTableSize();
 	});
 


olifolkerd:
Hey @lekoala 

Thanks for submitting this as  PR, it has made it much clearer to see how you have things setup.

Before i discuss the PR, i just wanted to reframe this discussion as i feel the talk on the issue this came from was starting to become unconstructive. 

While i appreciate that you have spent time building this out for your usage case, when i accept a PR request, i must approach it from the perspective of the needs of entire user base, which means at times some of the changes i request will not directly apply to your usage case but will none the less be needed for this PR to be successfully included in the project.

As per my initial feedback the following changes would need to be made for this to work out correctly:

### Layout
At the moment you are using psuedo before elements to build your labels using fixed widths. This presents many challenges with making a layout that is responsive to a large number of different column title configurations and would need to be changed to use actual elements that users have the control to style and done in such a way that would maintain the correct label/cell spacing across multiple lines

Now i understand that in your initial usage case that using flex layout on the row was an effective solution, and it represented a simpler option, but i need to be clear on this, that converting the row to a flex layout is not a systemically viable solution. Many other users use row formatters and other table features that will not play well with this arrangement, that is the reason that the existing collapsed formatter works in its own element.

This is not a matter that is up for discussion, a number of table features and a lot of users custom styles are predicated on the row being display type block and this is not something that can be changed. I understand that this will mean a bit more work is needed to build out the display, but if it dosnt happen then this PR will not be accepted.

Looking at your code it would not take much effort to move it to a system were it gets the cell element and appends it to a container/table and then sticks it back in the row before it is made visible. Im happy to discuss with approaches for this that will work best with the rest of the table.

### Option Naming
It would be better if the name of the collapse option explained what it did differently from the main collapse option. I would propose a name of `collapseEditable` would be more intuitive than `collapseFlex`

### Editing Cell Labels
When editing the cells, the labels would need to remain in place rather than being hidden. Removing the label when editing creates a loss of context that can make it confusing if you come back to a form part way through editing.

### Editable Cell Styling
The cells that are editable should not be styled differently, Tabulator does not by default style editable cells differently, that should be left up to the user to implement if they want based on the tabulator-editable class on the cell

Cheers

Oli :)

lekoala:
Yes all that makes sense. I don't have the time right now to investigate how to do that but if I manage to find some spare time I'll give it a try :)

src/js/modules/ResponsiveLayout/ResponsiveLayout.js
```diff
@@ -109,19 +113,50 @@ class ResponsiveLayout extends Module{
...
+		}
+	}
+
 	initializeRow(row){
 		var el;
 
 		if(row.type !== "calc"){
-			el = document.createElement("div");
-			el.classList.add("tabulator-responsive-collapse");
+			if(this.table.options.responsiveLayout === "collapse"){
```
ortwic, line 149: How is this supposed to work? el in L.160 will fail as it will be undefinied.

src/js/modules/ResponsiveLayout/ResponsiveLayout.js
```diff
@@ -244,6 +289,9 @@ class ResponsiveLayout extends Module{
 
 		if(row.modules.responsiveLayout){
 			el = row.modules.responsiveLayout.element;
+			if(!el){
```
ortwic, line 292: I guess this problem results from L.150 by passing an undefined `el`.

lekoala:
@ortwic i do not intend to work on this any time soon so feel free to make your own proposal based on oli's feedback

lekoala:
https://github.com/ortwic/tabulator/pull/1

PR #3896, opened 2022-08-03
Fixing unresolved promise in updateData().

matpen:
I found that when calling `updateData()` and `updateOrAddData()` with an empty array, the promise is never resolved. This is because the current code tests for a falsey input, which does not cover for an empty array.

### DIFF
--- a/src/js/core/Tabulator.js
+++ b/src/js/core/Tabulator.js
@@ -417,7 +417,7 @@ class Tabulator {
 				data = JSON.parse(data);
 			}
 			
-			if(data?.length){
+			if(data && data.length > 0){
 				data.forEach((item) => {
 					var row = this.rowManager.findRow(item[this.options.index]);
 					
@@ -483,7 +483,7 @@ class Tabulator {
 				data = JSON.parse(data);
 			}
 			
-			if(data?.length){
+			if(data && data.length > 0){
 				data.forEach((item) => {
 					var row = this.rowManager.findRow(item[this.options.index]);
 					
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @matpen 

Thanks for your PR.

That definitely does resolve that issue, but could i ask you to change the style of your code to ensure compatibility in a wider variety of environments.

Could you replace the conditional object property you are using on the length with a check to see the object exists and that the property is set. Several of the environments that Tabulator is used in do not support the use of conditional object properties, which is why you will see they are not used anywhere else in the codebase.

If you can make that change I would be happy to accept this PR.

Cheers

Oli :)

matpen:
Thank you for your quick reply, Oli. If I understood you correctly, you prefer to avoid the use of the [Optional Chaining Operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), so I replaced that. I am happy to make more adjustments, so do not hesitate to ask.

olifolkerd:
@matpen if you are on the Tabulator discord chanel, drop me a DM and i will award you the contributor badge :)

matpen:
Ohhh, thank you... that is nice! However, I do not use discord :( 
Hopefully I can still make contributions without it?

olifolkerd:
Oh yes certainly please make as many PRs as you like.

The badge is an offering of recognition for those that help.

It is worth popping into the discord if you have a min. It is where a lot
of the community ask questions, share projects and discuss ideas

Cheers

Oli :)

PR #3935, opened 2022-09-09
SImplify 'date' filtering

FrantzUml:
To filter a date column when the data column is string, I have to :
1- add a Mutator: string to Luxon DateTime and format it with "yyyy-MM-dd"
2 - add a Formatter to show wanted outputFormat "dd/MM/yyyy"
3 - add an Editor to make the filter

Why the default date Editor compares in format "yyyy-MM-dd" ? and not on local system format ?

Thanks


olifolkerd:
Hey @FrantzUml 

You appear to have created a pull request, but have not made changes to the code.

If you want to submit a feature request you need to do this as an issue not a PR.

And just as a note you don't need to do most of the steps outlined above. You don't need to mutate the data at any point. And  a formatter is purely a visual thing so it is totally your choice to use it or not, it isn't required for filtering.

As for filtering with a date editor it isn't really advised. As is mentioned in the date editor docs, it uses the built in date input element which has very limited capabilities. It strongly recomends that you use a custom editor with a 3rd party date picker if you want to do anything complex.

So in reality the only thing you need to do is set up a custom header filter. The rest is entirely your choice.

Cheers

Oli:)

Issue #3963, opened 2022-09-30
GroupComponent.isVisible() always undefined


andy-5:
**Describe the bug**
`GroupComponent.isVisible()` always returns `undefined`. See console output of https://jsfiddle.net/ne3rv8k5/1/ .

The reason is that `isVisible()` [returns the field](https://github.com/olifolkerd/tabulator/blob/dbb5cb544b228e573ac50bbb3a94e12af84dea87/src/js/modules/GroupRows/GroupComponent.js#L43) `_visible` of the group, which is not defined - it should be `visible` (without underscore).

**Tabulator Info**
The bug is present in the latest git version.

**Working Example**
https://jsfiddle.net/ne3rv8k5/1/

**To Reproduce**
- Go to https://jsfiddle.net/ne3rv8k5/1/ and run it
- See console output with the results of `isVisible()`

**Expected behavior**
`isVisible()` should return either `true` or `false`, depending on the `visible` field of the group

olifolkerd:
Hey @andy-5 

As per the comment on the PR, this fix will be available on the 5.4 release in a couple of hours.

Cheer

Oli :)

PR #3967, opened 2022-10-03
fix isVisible() of GroupComponent

andy-5:
Fix for #3963.

I found calling `_group._visSet()` to be necessary, as `group.visible` may be a function initially (until `_visSet()` is called).

### DIFF
--- a/src/js/modules/GroupRows/GroupComponent.js
+++ b/src/js/modules/GroupRows/GroupComponent.js
@@ -40,7 +40,8 @@ class GroupComponent {
 	}
 
 	isVisible(){
-		return this._group._visible;
+		this._group._visSet();
+		return this._group.visible;
 	}
 
 	show(){


olifolkerd:
Hey @andy-5 

Thanks for the PR.

Good spot on the underscore there. Changing that fixes the issue, but im not sure why you have included the `visSet` function call there, as it should be set before that visibility check function is ever called.

As im just about to release 5.4 in the next hour or so i have included the underscore fix on there, and we can continue the discussion on the `visSet` function here.

Cheers

Oli :)

andy-5:
I had some code (that I've already refactored since) that called `isVisible()` from within the `dataGrouped` event. At that point the `visible` attribute was still set to a function. I guess this might have been an invalid use of the API.

(The use-case was to check if grouping resulted in only a single group, and to show that group if it was not already visible. But I've discarded that idea for now.)

I'm fine with the current fix, so I think this PR can be closed.

PR #3974, opened 2022-10-08
Bug Fix: List editor doesn't scroll to item the first time

jsnyder4:
**Demonstration of the bug**: https://jsfiddle.net/q65n8f7e/
Click any cell in the City column and type the letter D to scroll to Denver in the list. Nothing happens. Type the letter D again and it will scroll.

**Issue**: In the List.js constructor, `this.filterTerm` is never being initialized. Therefore, the first time a letter it typed this.filterTerm is being set to undefined + the letter (e.g. `undefinedd`) on line 436.

**Fix**: simply initialize this.filterTerm to empty string, which is what happens anyway after an 800 ms delay on line 447.


olifolkerd:
Hey @jsnyder4 

Good spot,

Thanks for the fix, i will include that in todays patch release.

Cheers

Oli :)

Issue #3962, opened 2022-09-30
Add option to reset the persisted pagination Initial Page when the dataset is too small when accessing remote data


antman3351:
Hi,

Not sure if this is a bug or a misuse of Tabulator.
 
With remote data & remote pagination.
When the current page is loaded from the persistence and the dataset is smaller ( has less pages ) making selecting the persisted page impossible, Tabulator shows an empty table.

This can happen when the rows of a table are deleted from the backend, when you reload the page the dataset is smaller. 

**Tabulator Info**
5.3.4

**Working Example**
https://jsfiddle.net/1c2tyhv9/1/

**To Reproduce**
1. Open the fiddle

**Expected behaviour**
If the page is not existent, the `paginationInitialPage` should be reset

Thanks!
Antonio

olifolkerd:
Hey @antman3351 

What would be your desired outcome in this scenario,

The server should return an empty array as there are no rows on that page, then when it responds with a max page of less thatn the current page Tabulator should make a request to the highest page instead?

Cheers

Oli :)

antman3351:
Hi Oli, 

Good question I hadn't given it much thought. As the value is invalid personally I'd start from page 1 but I can see how this could be annoying if it keeps throwing you back to the first page, so last page probably makes more sense.

Could it be a setting?

Thanks!

olifolkerd:
Hey @antman3351 

Yeah an option sounds like a good idea,

I will set this as a feature request, and add it to the todo list for a future minor release *(or feel free to submit a PR}*

Cheers

Oli :)

PR #3987, opened 2022-10-17
Add option to reset pagination when last page < current page

antman3351:
Hi Oli, 

I had a go at implementing the feature request  #3962. 

I added the option `paginationLastPageErrorResetTo` that expects a value of `first` or `last` with the default of `last`.

I'm not convinced by the return from my change (/src/js/modules/Page/Page.js@857).
Also would it be best to the the function finish loading ( dispatching the 'pageLoaded' event ) then reload the page ?

Let me know if I need to change anything :)

Cheers
Antonio


olifolkerd:
Hey,

Thanks for the pr. Sorry for the delayed reply, I have been focusing on the website the last couple of weeks

 I have a couple of points of feedback.

When you are changing the page no, I would recommend calling the modules internal setPage function instead of updating the variable yourself. That will trigger all the events and request that you need instead of replicating functionality.

You can also pass the word first or last directly to that function and it will take care of the rest.

The default behaviour should probably be to do nothing as it currently does, to allow for people to handle it differently. You could also allow the user to not just pass a string but to pass a callback that could then return the page incase they have a more complex usage case

As for the option name that is a bit on the verbose side. Maybe something like paginationOutOfRange or paginationMaxExceeded

Cheers

Oli :)

antman3351:
Hey Oli, 

I made the changes as you suggested.

Just two things:

Is it ok that the DataLoader issues a warning `Data Load Response Blocked - An active data load request was blocked by an attempt to change table data while the request was being made`. ? 

Also is the call to the user supplied function ok `paginationOutOfRange.call(this, this.page, this.max)` ?

cgountanis:
@antman3351 Sorry if this was covered but has this been implemented in the latest releases? We are running into same issue where last page is higher than dataset on revisits using built-in presentence enabled. Simple setting?

cgountanis (commented):
Would be cool if this was an option as of right now we get errors when dataset gets smaller and persistent page is > then available.

olifolkerd:
Hey @antman3351 

Cheers

Oli :)

PR #4035, opened 2022-11-24
Use consistent method of determining table width for different layouts

enkryptor:
The "fitColumns" layout uses getBoundingClientRect() which can lead to unstable results in some cases.

Both "fitDataStrech" and "fitDataFill" use this.table.rowManager.element.clientWidth so it's reasonable to use it in "fitColumns" for consistency.

### DIFF
--- a/src/js/modules/Layout/defaults/modes/fitColumns.js
+++ b/src/js/modules/Layout/defaults/modes/fitColumns.js
@@ -1,6 +1,6 @@
 //resize columns to fit
 export default function(columns, forced){
-	var totalWidth = this.table.rowManager.element.getBoundingClientRect().width; //table element width
+	var totalWidth = this.table.rowManager.element.clientWidth; //table element width
 	var fixedWidth = 0; //total width of columns with a defined width
 	var flexWidth = 0; //total width available to flexible columns
 	var flexGrowUnits = 0; //total number of widthGrow blocks across all columns


olifolkerd:
Can you provide more details about when this would be unstable. I have yet to come across a usage case where this is an issue.

Cheers

enkryptor:
@olifolkerd here's a screencast of an Angular application. Column widths are calculated differently on each Tabulator instantiation:

[Screencast_15.11.2022_14_15_30.webm](https://user-images.githubusercontent.com/668156/203995234-f7bbd825-a104-4162-9c1b-824afc9022da.webm)

enkryptor:
Anyways, the method used for determining a table size should be the same for all layout strategies (fitDataFill, fitColumns, etc).

olifolkerd:
Hey @enkryptor 

Sorry for the delay in getting back to you on this, had a bit of a family crisis over the xmas break and only just getting back to normal.

I have merged this in and will include it in the patch release later this week.

Cheers

Oli :)

olifolkerd:
Hey @enkryptor 

On further inspection the original method of calculating width was needed in this instance, removing it caused issues in variable height tables with optional scroll bars. I have therefore reverted this PR.

There was another issue affecting the fit columns layout but I will be patching that separately later today.

Cheers

Oli :)

enkryptor:
@olifolkerd is this problem relevant for other layouts like `fitDataStretch` which use `.clientWidth`?

olifolkerd:
@enkryptor nope not at all, it is due to the unique needs of the fit columns layout function

PR #4045, opened 2022-11-30
Memory leak fix

### LINKED ISSUE #4044: Memory leak in .destroy()
Hi Oli,

When calling `table.destroy()` memory isn't freed. 

The first issue is Print.js doesn't remove the event listeners attached to the window, so the garbage collector doesn't remove the table instance.

The second is `destroy()` doesn't empty the `activeRowsPipeline` array, so if a reference to the table is still being held after calling `destroy` memory is still being occupied.

Tabulator-memory-leak

**Tabulator Info**
5.4.2

**To Reproduce**
1. Instance the table
2. load a lot of data
3. call table.destroy()

antman3351:
Closes #4044 
Fix memory leak in Print not removing event listener on window and RowManager not emptying activeRowsPipeline


olifolkerd:
Hey @antman3351 

Thanks for the great PR.

I like your approach to moving the row manager destruction functionality into the destroy function, that is a much cleaner and more isolated approach than the current setup.

And the fix to the print module event handlers will do the job nicely. 

I will have a quick look through the other modules to make sure i haven't missed any others. I resolved this with most other events when i moved all interactions to the interaction monitor but i clearly missed the non  explicitly mouse/key events.

I will approve this now and include it in the next patch release which im hoping to do in the next week or so.

Cheers

Oli :)

PR #4071, opened 2022-12-21
check existence of cells before generating new

### LINKED ISSUE #4056: updateData function consumes all the available memory
**Describe the bug**
I have a table with roughly 500 rows, where only 50 are visible. Few cells in the row are static, and few of the periodic changes are based on the data from the server. I am reading numeric (float) values and some strings. I am using `updateData` function to update selected rows based on the current data. This function is called every 2000ms. The problem is that after a while, the page consumes all available memory and the browser shuts down the tab.

**Tabulator Info**
- 5.4

**Working Example**
https://jsfiddle.net/zag41ted/32/

**To Reproduce**
Just run the example and watch the Task manager. You can adjust the setInterval delay to speed up the memory consumption.

**Expected behavior**
Do not create new cell objects?

**Screenshots**
image

 - OS: Windows 10
 - Browser Opera
 - Version 93.0.4585.21

**Additional context**

jurafa:
If `updateData` is repeatedly called on not yet initialized rows new cells are always being generated causing memory leaks.
Fixes #4056.

### DIFF
--- a/src/js/core/row/Row.js
+++ b/src/js/core/row/Row.js
@@ -315,7 +315,7 @@ export default class Row extends CoreFeature{
 		
 		column = this.table.columnManager.findColumn(column);
 		
-		if(!this.initialized){
+		if(!this.initialized && this.cells.length === 0){
 			this.generateCells();
 		}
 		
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @jurafa 

Sorry for the delay in getting back to you on this, i had a bit of a family emergency over the xmas break and things are only just starting to return to normal.

Thankyou very much for your PR, i will merge this into the master branch now, and include it in the patch release later this week.

Cheers

Oli :)

jurafa:
I am glad I could help.
Have a nice day. :-)

Jura

Issue #4019, opened 2022-11-16
Date cell editing issue


adryx92:
Hello, 

**Describe the bug**
when editing a date cell input with the keyboard, Tabulator triggers the `cellEdited` event before the input is complete. This can lead to inconsistent or wrong dates. 

**Tabulator Info**
I'm using Tabulator 5.4.2.

**Working Example**
I created this pen to demonstrate the issue https://codepen.io/adryx_92/pen/zYadQLJ

**To Reproduce**
Try to edit the date cell using the keyboard: the `cellEdited` is triggered while the user is still editing. With text and number everything is fine.

**Expected behavior**
The `cellEdited` should be triggered after the edit is complete.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Tested on Firefox and Edge

Thank you

adryx92:
After some research, I found out that for `<input type='date'>` fields the `onchange` event is fired as soon as the browser detects a "valid" date in the field.

https://stackoverflow.com/a/60996719/3352304

 This weird behaviour is the root of the problem.

olifolkerd:
Closing this as it has been fixed in the #4112 merge that will be included in todays patch release.

Cheers

Oli :)

Issue #4102, opened 2023-01-26
Date editor is glitchy with manual entry and/or picker


MWWorks:
Great to see the date editor now built in (and happy that it uses the native date input type). But it seems a bit glitchy - see example using 5.4.3:

https://jsfiddle.net/js6hwanu/3/

- The native datepicker widget works as expected in Chrome; but in Firefox, clicking a date doesn't change the cell content
- In either browser, manual entry doesn't work - eg I'm UK locale and the picker shows dd/mm/yyyy - I can't type a full date, at some point during keying the focus drops and no change is saved to the cell
 
Looks like some event handling is amiss?

wdehoog:
I noticed it as well. It used to work so I guess it is due to an update of Firefox.
Seems that now one needs to listen to "input" events.

MWWorks:
As a followup, some experimenting and hunting around has shed a little light on this. Roughly:

- When using the picker, FF and Chrome fire a change event with the new value
- But FF also precedes this with a blur event with the old value - hence the tabulator editor 'succeeds' with that
- When manually entering dates, both browsers fire 0 or more change events for each component of the date - this also causes the editor to succeed before input is complete

Although not perfect, this is my current setup which covers most use patterns; it starts with the change event (that works for for using the picker), then when keying starts, removes it in favour of blue (after keying):

    Tabulator.extendModule("edit", "editors", {

        cdate:function(cell, onRendered, success, cancel, editorParams){
            let input = document.createElement("input");

            input.type = "date";
            input.style.padding = "4px";
            input.style.width = "100%";
            input.style.boxSizing = "border-box";
            input.value = cell.getValue();;

            onRendered(function(){
                input.focus();
                input.style.css = "100%";
            });

            let accept = function(e){success(input.value)};

            input.addEventListener("change", accept);
            input.addEventListener("keydown", function(e){
                input.removeEventListener("change", accept)
                input.addEventListener("blur", accept)
            });

            return input;
        }
    });

mastrdev:
@MWWorks Interestingly, I couldn't replicate this bug with Firefox ESR (v102); I had to install Firefox v109.

MWWorks:
> @MWWorks Your suggested code was still buggy since built-in date pickers update the input when selecting month/year, firing a change event prematurely.

Oh, good catch - I knew it would fail if starting keying then used the picker - but I hadn't tried the month picker. Just tried yours, seems working well, thanks for going the extra mile :)

[...]

PR #4107, opened 2023-01-30
Fix DT editors

mastrdev:
With date, datetime, and time inputs, the "change" event fires with nearly every key press (whenever there's a valid entry), so the editors are calling _onChange(e)_ before the date can be typed fully. I removed the "change" event listeners so the changes are only processed upon _blur_ or _enter key_. This fixes Issues #4019 and #4102.

[JSFiddle after fix](https://jsfiddle.net/tzvmoeq8/)

### DIFF
--- a/src/js/modules/Edit/defaults/editors/date.js
+++ b/src/js/modules/Edit/defaults/editors/date.js
@@ -83,8 +83,7 @@ export default function(cell, onRendered, success, cancel, editorParams){
 		}
 	}
 	
-	//submit new value on blur or change
-	input.addEventListener("change", onChange);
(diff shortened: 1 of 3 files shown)


MWWorks:
Note this works for picker+keyboard in Chrome, but doesn't work for the picker in Firefox - which fires both blur and change after using the picker - and when the blur fires, the field value is still the old one.

I have a suggested alternative in #4102 - although it may need refinement to work in all cases.

mastrdev:
I admit that I didn't test with FireFox. Will resubmit once that's working. Thanks, @MWWorks !

PR #4111, opened 2023-02-01
(Ignore)

mastrdev:
With date, datetime, and time inputs, the "change" event fires with nearly every key press (whenever there's a valid entry), so the editors are calling _onChange(e)_ before the desired value can be typed fully. I removed the "change" event listeners so the changes are only processed upon "blur" or enter key. I also added logic to ignore the premature "blur" event when opening FireFox's built-in date/time picker. Tested this solution in Edge, Chrome, FireFox, and Safari. This fixes Issues #4019 and #4102.

[JSFiddle after fix](https://jsfiddle.net/m0sv5dxk/)


mastrdev:
The DeepScan flags are outside the changes I made. I saw that unused _e_ but left it alone.

mastrdev:
Sorry, screwed up my branch again. Will resubmit.

PR #4112, opened 2023-02-01
Fix date/time editors

mastrdev:
With date, datetime, and time inputs, the "change" event fires with nearly every key press (whenever there's a valid entry), so the editors are calling _onChange(e)_ before the desired value can be typed fully. I removed the "change" event listeners so the changes are only processed upon "blur" or enter key. I also added logic to ignore the premature "blur" event when opening FireFox's built-in date/time picker. Tested this solution in Edge, Chrome, FireFox, and Safari. This fixes Issues #4019 and #4102.

[JSFiddle after fix](https://jsfiddle.net/m0sv5dxk/)

### DIFF
--- a/src/js/modules/Edit/defaults/editors/date.js
+++ b/src/js/modules/Edit/defaults/editors/date.js
@@ -66,7 +66,7 @@ export default function(cell, onRendered, success, cancel, editorParams){
 		}
 	});
 	
-	function onChange(e){
+	function onChange(){
 		var value = input.value;
 		
 		if(((cellValue === null || typeof cellValue === "undefined") && value !== "") || value !== cellValue){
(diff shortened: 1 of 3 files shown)


mastrdev:
The DeepScan flags are outside the changes I made. I saw that unused _e_ but left it alone.

olifolkerd:
Hey @mastrdev 

Thanks for the PR, the bulk of the content looks good but you have removed the new lines from the end of the editor files which is why the scan is failing.

All files must end on a new line as part of the Tabulator coding standard as it improves commitability of the content.

Please reinstate the ending new lines i the file and i will be happy to accept this PR.

Cheers

Oli :)

mastrdev:
Ok, for some reason the GitHub web editor stripped those ending new lines on its own. Will fix. Want me to remove the unused 'e' parameter in the onChange function so it clears the DeepScan check, or should we keep it around for style/consistency?

olifolkerd:
Great, thanks!

mastrdev:
Well, apparently my first commit had actually added ending newlines; they were missing before. I went ahead and removed the unused 'e' parameter for onChange() so the DeepScan passes without issue now. This PR should be ready for merge.

olifolkerd:
Hey @mastrdev 

Great work, thanks. i will merge that into master now and include it in the patch release in a few days.

Cheers

Oli :)

PR #4134, opened 2023-02-23
feat: renderer performance improvements

### LINKED ISSUE #4133: feat: renderer performance improvements
*Is your feature request related to a problem? Please describe.**

I have been playing with some changes in a fork which generally improve renderer performance for both fixed height and dynamic height cases.  

**Describe the solution you'd like**
Here is a summary of changes in the PR

- BasicHorizontal and BasicVertical Render will add to dom in bulk which reduces reflow
- Add elements to the dom in bulk instead of one at a time in VirtualDomVertical
- Only reRenderInPosition if the table height changes

lukecotter:
## Avoid block selection when clearing selection
This caused a large amount of reflow and it seemed like we did not need to block selection when it was about to be cleared anyway, but I may be missing something 😄 

## Only reRenderInPosition if the table height changes
This prevents `rerenderRows` from being called multiple times when the table has not resized.
This is especially useful for dynamic height tables.

## Add elements to the dom in bulk instead of one at a time in VirtualDomVertical
This will add a chunk of rows then check the height and determine if more rows are needed.
This is instead of adding 1 row then checking height and reduces reflow as a result which improves performance.

**Dynamic height case**
if there is no height on a table the virtual renderer will now attempt to render all rows.

Previously it would enter into a loop where some rows would be added, the table would resize then more rows would be added then resize until allow rows were added to the dom.

rows added -> table resize -> rows added -> table resize, repeat until all rows are added.

**Create table 93 rows, 4 columns: profile before**
dynamic-height-create-before

**Create table 93 rows, 4 columns: profile after**
dynamic-height-create-after

**Move row before**
dynamic-height-move-before

**Move row after**
dynamic-height-move-after

**Fixed Height case**
The number of rows to render is estimated upfront using the vDomRowHeight.

We then use this as a chunk size and keep updating in chunks until we have added enough rows to fit in the container + buffer.

This does mean is is possible we will now add more slightly rows to the dom, up to a maximum of one full chunk. 
This is likely to be around 80 rows.

**Create table 93 rows, 4 columns: profile before**
fixed-height-create-before

**Create table 93 rows, 4 columns: profile after**
fixed-height-create-after

**Move row before**
fixed-height-move-before

**Move row before**
fixed-height-move-after

I hope these changes are useful and ok. If not I am happy to chat about any of them. I can always split them up if needed.

resolves #4133

### DIFF
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -141,7 +141,7 @@ export default class RowManager extends CoreFeature{
 				this.dispatch("scroll-vertical", top, topDir);
 				this.dispatchExternal("scrollVertical", top, topDir);
 			}
-		});
+		}, {passive: true});
 	}
 	
 	////////////////// Row Manipulation //////////////////
@@ -1016,7 +1016,7 @@ export default class RowManager extends CoreFeature{
 				minHeight = isNaN(this.table.options.minHeight) ? this.table.options.minHeight : this.table.options.minHeight + "px";
 				
 				const height = "calc(100% - " + otherHeight + "px)";
-				this.element.style.minHeight = minHeight || "calc(100% - " + otherHeight + "px)";
+				this.element.style.minHeight = minHeight || height;
 				this.element.style.height = height;
 				this.element.style.maxHeight = height;
 			} else {
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -141,7 +141,7 @@ export default class RowManager extends CoreFeature{
 				this.dispatch("scroll-vertical", top, topDir);
 				this.dispatchExternal("scrollVertical", top, topDir);
 			}
-		}, {passive: true});
+		});
 	}
 	
 	////////////////// Row Manipulation //////////////////
@@ -1016,7 +1016,7 @@ export default class RowManager extends CoreFeature{
 				minHeight = isNaN(this.table.options.minHeight) ? this.table.options.minHeight : this.table.options.minHeight + "px";
 				
 				const height = "calc(100% - " + otherHeight + "px)";
-				this.element.style.minHeight = minHeight || height;
+				this.element.style.minHeight = minHeight || "calc(100% - " + otherHeight + "px)";
 				this.element.style.height = height;
 				this.element.style.maxHeight = height;
 			} else {
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -1081,7 +1081,9 @@ export default class RowManager extends CoreFeature{
 		this.table.tableWidth = this.table.element.clientWidth;
 		
 		if(!force){
-			resized && this.reRenderInPosition();
+			if(resized) {
+				this.reRenderInPosition();
+			}
 			this.scrollHorizontal(this.scrollLeft);
 		}else{
 			this.renderTable();
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -1020,7 +1020,6 @@ export default class RowManager extends CoreFeature{
 				this.element.style.height = height;
 				this.element.style.maxHeight = height;
 			} else {
-				resized = true;
 				this.element.style.height = "";
 				this.element.style.height =
 					this.table.element.clientHeight - otherHeight + "px";
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -1015,9 +1015,10 @@ export default class RowManager extends CoreFeature{
 			if(this.fixedHeight){
 				minHeight = isNaN(this.table.options.minHeight) ? this.table.options.minHeight : this.table.options.minHeight + "px";
 				
-				minHeight = minHeight || "calc(100% - " + otherHeight + "px)";
 				const height = "calc(100% - " + otherHeight + "px)";
-				this.element.style.cssText = `min-height:${minHeight}; height:${height}; max-height:${height};`;
+				this.element.style.minHeight = minHeight || "calc(100% - " + otherHeight + "px)";
+				this.element.style.height = height;
+				this.element.style.maxHeight = height;
 			} else {
 				resized = true;
 				this.element.style.height = "";
@@ -1082,7 +1083,7 @@ export default class RowManager extends CoreFeature{
 		
 		if(!force){
 			resized && this.reRenderInPosition();
-				this.scrollHorizontal(this.scrollLeft);
+			this.scrollHorizontal(this.scrollLeft);
 		}else{
 			this.renderTable();
 		}
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -1006,7 +1006,7 @@ export default class RowManager extends CoreFeature{
 	
 	//adjust the height of the table holder to fit in the Tabulator element
 	adjustTableSize(){
-		var initialHeight = this.element.clientHeight, minHeight;
+		let initialHeight = this.element.clientHeight, minHeight;
 		
 		if(this.renderer.verticalFillMode === "fill"){
 			let otherHeight =  Math.floor(this.table.columnManager.getElement().getBoundingClientRect().height + (this.table.footerManager && this.table.footerManager.active && !this.table.footerManager.external ? this.table.footerManager.getElement().getBoundingClientRect().height : 0));
@@ -1014,12 +1014,12 @@ export default class RowManager extends CoreFeature{
 			if(this.fixedHeight){
 				minHeight = isNaN(this.table.options.minHeight) ? this.table.options.minHeight : this.table.options.minHeight + "px";
 				
-				this.element.style.minHeight = minHeight || "calc(100% - " + otherHeight + "px)";
-				this.element.style.height = "calc(100% - " + otherHeight + "px)";
-				this.element.style.maxHeight = "calc(100% - " + otherHeight + "px)";
-			}else{
-				this.element.style.height = "";
-				this.element.style.height = (this.table.element.clientHeight - otherHeight) + "px";
+				minHeight = minHeight || "calc(100% - " + otherHeight + "px)";
+				const height = "calc(100% - " + otherHeight + "px)";
+				this.element.style.cssText = `min-height:${minHeight}; height:${height}; max-height:${height};`;
+			} else {
+				this.element.style.height =
+					this.table.element.clientHeight - otherHeight + "px";
 				this.element.scrollTop = this.scrollTop;
 			}
 			
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -1007,6 +1007,7 @@ export default class RowManager extends CoreFeature{
 	//adjust the height of the table holder to fit in the Tabulator element
 	adjustTableSize(){
 		let initialHeight = this.element.clientHeight, minHeight;
+		let resized = false;
 		
 		if(this.renderer.verticalFillMode === "fill"){
 			let otherHeight =  Math.floor(this.table.columnManager.getElement().getBoundingClientRect().height + (this.table.footerManager && this.table.footerManager.active && !this.table.footerManager.external ? this.table.footerManager.getElement().getBoundingClientRect().height : 0));
@@ -1018,6 +1019,8 @@ export default class RowManager extends CoreFeature{
 				const height = "calc(100% - " + otherHeight + "px)";
 				this.element.style.cssText = `min-height:${minHeight}; height:${height}; max-height:${height};`;
 			} else {
+				resized = true;
+				this.element.style.height = "";
 				this.element.style.height =
 					this.table.element.clientHeight - otherHeight + "px";
 				this.element.scrollTop = this.scrollTop;
@@ -1027,6 +1030,7 @@ export default class RowManager extends CoreFeature{
 			
 			//check if the table has changed size when dealing with variable height tables
 			if(!this.fixedHeight && initialHeight != this.element.clientHeight){
+				resized = true;
 				if(this.subscribed("table-resize")){
 					this.dispatch("table-resize");
 				}else{
@@ -1038,6 +1042,7 @@ export default class RowManager extends CoreFeature{
 		}
 		
 		this._positionPlaceholder();
+		return resized;
 	}
 	
 	//reinitialize all rows
@@ -1072,15 +1077,12 @@ export default class RowManager extends CoreFeature{
 	
 	//redraw table
 	redraw (force){
-		var left = this.scrollLeft;
-		
-		this.adjustTableSize();
-		
+		const resized = this.adjustTableSize();
 		this.table.tableWidth = this.table.element.clientWidth;
 		
 		if(!force){
-			this.reRenderInPosition();
-			this.scrollHorizontal(left);
+			resized && this.reRenderInPosition();
+				this.scrollHorizontal(this.scrollLeft);
 		}else{
 			this.renderTable();
 		}
--- a/src/js/core/rendering/renderers/VirtualDomVertical.js
+++ b/src/js/core/rendering/renderers/VirtualDomVertical.js
@@ -30,6 +30,7 @@ export default class VirtualDomVertical extends Renderer{
 
 		this.vDomTopNewRows = []; //rows to normalize after appending to optimize render speed
 		this.vDomBottomNewRows = []; //rows to normalize after appending to optimize render speed
+		this.initTableHeight = this.tableElement.clientHeight;
 	}
 
 	//////////////////////////////////////
@@ -266,7 +267,7 @@ export default class VirtualDomVertical extends Renderer{
 			const rowsToRender = Math.max(this.vDomWindowMinTotalRows, Math.ceil((containerHeight / this.vDomRowHeight) + (this.vDomWindowBuffer / this.vDomRowHeight)));
 			
 			let totalRows = 0;
-			while(((rowsHeight <= containerHeight + this.vDomWindowBuffer) || totalRows < this.vDomWindowMinTotalRows) && this.vDomBottom < rowsCount -1) {
+			while(((this.initTableHeight <= 0 || rowsHeight <= containerHeight + this.vDomWindowBuffer) || totalRows < this.vDomWindowMinTotalRows) && this.vDomBottom < rowsCount -1) {
 				// console.debug('1', i, this.vDomWindowMinTotalRows);
 				let renderedRows = [];
 				const dFrag = document.createDocumentFragment();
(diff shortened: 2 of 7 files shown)


olifolkerd:
Hey @lukecotter 

Thanks for the PR, and for the detailed explanation that is very helpful.

That is an interesting approach you have taken there, and I would certainly be interested in including it in the library!

There are a couple of points i would like to discuss before this gets merged in:

## Tables with no fixed height but  Max Heigh


On the whole I think you are on to something good here, but there is one area that I am a little concerned has been overlooked in the current approach *(and I totally may have missed something in your implementation so welcome feedback)*.

One of the main reasons for the row by row addition is to allow for variable height tables that then become fixed when a `maxHeigh` is reached. In the existing implementation it checks for this and stops adding new rows when a max height has been reached and the virtualDOM window buffer has been filled. But in this PR, it assumes it is always going to show all rows and draws them all, which will cause DOM blocking for large data sets. 

Maybe in that scenario if there is a max height set we need to look at rendering in chunks to get the best of both worlds, maybe in chunks the size of the virtual DOM window buffer.

## Code styling and minor tweaks
Running through the code, there are a few minor tweaks that would be needed to bring things inline with Tabulators coding style:
- You changes to the RowManager on line 1084 you have added the line `resized && this.reRenderInPosition();`. For clarity and consistent, Tabulator requires that all conditional logic be contained in curly braces and have condition and action clearly separated, could you please wrap this in an `if` statement with curly braces
-  in the `_clearSelection` function of `Tabulator.js` you have removed the functions the `tabulator-block-select`, this is needed to prevent text selection visual glitching, and is also used by developers outside of Tabulator to handle styling while the selection is being reset. Please reinstate this.
- In the `renderRows` funcrion of  `BasicVertical` you have added your table frag as a const after the variable definitions at the top of the function, please include this as part of the variables list (this should be the same with any variable list, please include it at the top
- In the `VirtualDomVertical` you have removed the `minWidth` clearing on line 47, please leave this in unless you have a specific reason to remove it
- In `VirtualDomVertical` there are lots of const variables declared as they are assigned in the `if` statements and other logic, please handle all variable declarations at the top of the function definition and assign them when they are used.
- At line 295 & 305 in the `VirtualDomVertical` you have duplicate for loop and if logic, you could combine both loops and just call the two functions one after the other. you also have another iterator over the `renderedRows` object immediately after that which could be combined
- On line 312 of the `VirtualDomVertical` you are defining a `me` variable. you don't need this variable if you use an arrow function for the `forEach` call, then the scope of the parent is preserved and you can just use `this`


Let me know your thoughts, i am planning a release in a couple of weeks time, so if we can get this finished by then i  would love to include it.

Cheers

Oli :)

src/js/core/RowManager.js
```diff
@@ -1072,15 +1077,14 @@ export default class RowManager extends CoreFeature{
...
-		
-		this.adjustTableSize();
-		
+		const resized = this.adjustTableSize();
 		this.table.tableWidth = this.table.element.clientWidth;
 		
 		if(!force){
-			this.reRenderInPosition();
-			this.scrollHorizontal(left);
+			if(resized) {
```
lukecotter, line 1084: >You changes to the RowManager on line 1084 you have added the line resized && this.reRenderInPosition();. For clarity and consistent, Tabulator requires that all conditional logic be contained in curly braces and have condition and action clearly separated, could you please wrap this in an if statement with curly braces

braces added :)

lukecotter:
>in the _clearSelection function of Tabulator.js you have removed the functions the tabulator-block-select, this is needed to prevent text selection visual glitching, and is also used by developers outside of Tabulator to handle styling while the selection is being reset. Please reinstate this.

Reintroduced

src/js/core/rendering/renderers/VirtualDomVertical.js
```diff
@@ -44,7 +44,6 @@ export default class VirtualDomVertical extends Renderer{
 
 		element.style.paddingTop = "";
 		element.style.paddingBottom = "";
-		// element.style.minWidth = "";
```
lukecotter, line 47: >In the VirtualDomVertical you have removed the minWidth clearing on line 47, please leave this in unless you have a specific reason to remove it

I actually just removed commented out code but I can reintroduce if you want.

src/js/core/rendering/renderers/VirtualDomVertical.js
```diff
@@ -223,25 +222,34 @@ export default class VirtualDomVertical extends Renderer{
...
 		topPad = 0,
 		rowsHeight = 0,
+		rowHeight = 0,
 		heightOccupied = 0,
 		topPadHeight = 0,
 		i = 0,
 		rows = this.rows(),
 		rowsCount = rows.length,
-		containerHeight = this.elementVertical.clientHeight;
+		index = 0,
```
lukecotter, line 236: >In VirtualDomVertical there are lots of const variables declared as they are assigned in the if statements and other logic, please handle all variable declarations at the top of the function definition and assign them when they are used.

move vars to top

src/js/core/rendering/renderers/VirtualDomVertical.js
```diff
@@ -261,54 +269,95 @@ export default class VirtualDomVertical extends Renderer{
...
+				});
+
+				renderedRows.forEach((row) => {
+					if(!row.heightInitialized) {
+						row.setCellHeight();
+						
+					}
+				});
+
+				renderedRows.forEach((row) => {
```
lukecotter, line 324: >On line 312 of the VirtualDomVertical you are defining a me variable. you don't need this variable if you use an arrow function for the forEach call, then the scope of the parent is preserved and you can just use this

Switched to arrow functions

lukecotter:
@olifolkerd 
> Hey @lukecotter
> 
> Thanks for the PR, and for the detailed explanation that is very helpful.
> 
> That is an interesting approach you have taken there, and I would certainly be interested in including it in the library!
> 
> There are a couple of points i would like to discuss before this gets merged in:
> 
> ## Tables with no fixed height but Max Heigh
> On the whole I think you are on to something good here, but there is one area that I am a little concerned has been overlooked in the current approach _(and I totally may have missed something in your implementation so welcome feedback)_.
> 
> One of the main reasons for the row by row addition is to allow for variable height tables that then become fixed when a `maxHeigh` is reached. In the existing implementation it checks for this and stops adding new rows when a max height has been reached and the virtualDOM window buffer has been filled. But in this PR, it assumes it is always going to show all rows and draws them all, which will cause DOM blocking for large data sets.
> 
> Maybe in that scenario if there is a max height set we need to look at rendering in chunks to get the best of both worlds, maybe in chunks the size of the virtual DOM window buffer.
> 
> ## Code styling and minor tweaks
> Running through the code, there are a few minor tweaks that would be needed to bring things inline with Tabulators coding style:
> 
> * You changes to the RowManager on line 1084 you have added the line `resized && this.reRenderInPosition();`. For clarity and consistent, Tabulator requires that all conditional logic be contained in curly braces and have condition and action clearly separated, could you please wrap this in an `if` statement with curly braces
> * in the `_clearSelection` function of `Tabulator.js` you have removed the functions the `tabulator-block-select`, this is needed to prevent text selection visual glitching, and is also used by developers outside of Tabulator to handle styling while the selection is being reset. Please reinstate this.
> * In the `renderRows` funcrion of  `BasicVertical` you have added your table frag as a const after the variable definitions at the top of the function, please include this as part of the variables list (this should be the same with any variable list, please include it at the top
> * In the `VirtualDomVertical` you have removed the `minWidth` clearing on line 47, please leave this in unless you have a specific reason to remove it
> * In `VirtualDomVertical` there are lots of const variables declared as they are assigned in the `if` statements and other logic, please handle all variable declarations at the top of the function definition and assign them when they are used.
> * At line 295 & 305 in the `VirtualDomVertical` you have duplicate for loop and if logic, you could combine both loops and just call the two functions one after the other. you also have another iterator over the `renderedRows` object immediately after that which could be combined
> * On line 312 of the `VirtualDomVertical` you are defining a `me` variable. you don't need this variable if you use an arrow function for the `forEach` call, then the scope of the parent is preserved and you can just use `this`
> 
> Let me know your thoughts, i am planning a release in a couple of weeks time, so if we can get this finished by then i would love to include it.
> 
> Cheers
> 
> Oli :)

**Small changes**
I have added separate comments where I have made each of the changes you asked for.

**Tables with no fixed height but Max Height**
For this I have made a change in `VirtualDomVertical` to appends to the dom in chunks then check the table size and added more if needed. 

I have re run my performance tests too and added in max height

dynamic height

|        | before      | After    |
| ------ | ----------- | -------- |
| create | 850 ms      | 270.60ms |
| redraw | 400 ms (X3) | 135.12ms |

The before case (with my data) the render happens 3 times due to adjustTableSize causing the draw to repeat so total is closer to 1000 ms

fixed height - 100%

|        | before    | After     |
| ------ | --------- | --------- |
| create | 368.01 ms | 240.96ms  |
| redraw | 253.12 ms | 135.76 ms |

max height - 100%

|        | before    | After     |
| ------ | --------- | --------- |
| create | 441.53 ms | 280.12ms  |
| redraw | 200 ms    | 150.10 ms |

Let me know if you want any additional changes


**Intersect observer**
I have also been thinking about using the intersect observer in order to get a virtual rendered that will only render what is in the viewport. This should mean regardless of height performance would be the same.
I would obviously do this separately, but I wanted to know what you thought about the idea?

@olifolkerd

src/js/core/RowManager.js
```diff
@@ -141,7 +141,7 @@ export default class RowManager extends CoreFeature{
 				this.dispatch("scroll-vertical", top, topDir);
 				this.dispatchExternal("scrollVertical", top, topDir);
 			}
-		});
+		}, {passive: true});
```
lukecotter, line 144: I have changed to a passive event listener here for scroll performance.
I am not sure if you want a polyfill or feature check in unsupported browsers?

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#browser_compatibility

olifolkerd: This cannot be a passive event listener because there are touch events listened to by the interaction manager that can trigger interactions that will result in it being a non passive option

olifolkerd:
Hey @lukecotter 

Thanks for the speedy updates. Those are some great performance numbers there. Great Work! The chunking seems to work well there. Also thanks for the feedback on your commits, i had missed the read/write layout thrashing :)

And i just want to say at this point, thankyou very much for clearly taking the time to understand the renderer before making changes. Ive had a lot of attempted PR's on the renderer that broke large chunks of the Table and this is the first that shows care and consideration.


I have a couple more points of discussion for you:

### Passive Event Listeners
Could i ask a quick favour, i have noticed that you have introduced a passive event listener in your latest round of commits. Im keen to merge in your proposed solution, but if you keep on adding additional improvements in addition to the chunked rendering then with all the explanation I need to give around interactions with other parts of the table I worry this will never get merged in.

Could i ask that for now you focus only on the feedback from the initial PR without making further optimizations and then we can tackle the others as the come along.

The reason i ask this is the render is a complex beast and there are many usage cases that you may be unaware of. so can we tackle one change per PR to make it easier to get things reviewed and merged in. Im deffo keen to see and discuss any other improvements you may be looking to make, but i just dont want to turn this into a never ending PR

Could you remove the passive mode lister you have added at line 144 of the `RowManager`, this is deliberately not a passive listener *(you will notice there are passive listers already employed else where in the code base where appropriate)*. While i understand this has a minor performance impact it is because there are explicitly non passive actions triggered in other modules and the interaction monitor that are blocked by this and result in console errors.

### Variable Height Rows
The next challenge to tackle is variable height rows. *(and please correct me if I have misunderstood your logic here)*

One of the primary reason that Tabulator rendered the table row by row, is that a lot of usage cases see tabulator using rows of unknown height. To do this it rendered each row, calculated its height and then adjusted the rowsHeight calculation to match as it went.

Your current render math works out the number of rows to show based on the initial avg row height, this works fine when all the rows are the same height, but when tables have massive differences in rows *(some usage cases ive seen have rows that can vary from 20px to 1000px)* the it could lead to an very over filled or underfilled buffer.

Any thoughts on how you might deal with this with the new approach?

### Branch Change
Ive moved this PR to merge into the 5.5 branch so that it can be integrated into the other new features for the feature release later this month

### Intersection Observer
I love a good IntersectionObserver, and actually already use one to check table visibility, i would certainty be interested in hearing your thoughts on this in more detail, and certainly in another PR. 

A couple of points to consider on this, we don't want to just render the rows in the viewport there is a buffer on the VDOM is to allow all additions and removals to happen while elements are hidden outside of the viewport of the table, this ensure smooth scrolling when rows are added one by one as the user scrolls, the virtualRenderFill function is only called when a whole new set of data is added, the table is fully redrawn, or a person scrolls so fast that they move the viewport down its entire height, the rest of time rows are added and removed one at a time as the table scrolls and to do this efficiently and smoothly it needs the buffer. The second point to consider is how you are binding the intersection observers, one or two may result in improvement, but if you have a table with 100,000 rows that could be a lot of observers using a lot of resources.

But we can cover this in more detail in a separate PR once this one is finished,.

Let me know your thoughts on the variable height rows issue.

Cheers

Oli :)

src/js/core/RowManager.js
lukecotter: sorry about that I did not notice, I have reverted the change.

lukecotter:
> Hey @lukecotter
> 
> Thanks for the speedy updates. Those are some great performance numbers there. Great Work! The chunking seems to work well there. Also thanks for the feedback on your commits, i had missed the read/write layout thrashing :)
> 
> And i just want to say at this point, thankyou very much for clearly taking the time to understand the renderer before making changes. Ive had a lot of attempted PR's on the renderer that broke large chunks of the Table and this is the first that shows care and consideration.
> 
> I have a couple more points of discussion for you:
> 
> ### Passive Event Listeners
> Could i ask a quick favour, i have noticed that you have introduced a passive event listener in your latest round of commits. Im keen to merge in your proposed solution, but if you keep on adding additional improvements in addition to the chunked rendering then I worry this will never get merged in. Would you mind focusing on the feedback from the initial PR without making further optimizations and then we can tackle the others as they come along.
> 
> The reason i ask this is the render is a complex beast and there are many usage cases that you may be unaware of. so can we tackle one change per PR to make it easier to get things reviewed and merged in. Im deffo keen to see and discuss any other improvements you may be looking to make, but i just dont want to turn this into a never ending PR
> 
> Could you remove the passive mode lister you have added at line 144 of the `RowManager`, this is deliberately not a passive listener _(you will notice there are passive listers already employed else where in the code base where appropriate)_. While i understand this has a minor performance impact it is because there are explicitly non passive actions that can occur as a result of this event, (it broadcasts internal events that cascade throughout Tabulator) which cause issues in other modules and the interaction monitor that are blocked by this and result in console errors. When i added the passive listeners to the rest of Tabulator i experimented with adding one here and it resulted in errors.
> 
> ### Variable Height Rows
> The next challenge to tackle is variable height rows. _(and please correct me if I have misunderstood your logic here)_
> 
> One of the primary reason that Tabulator rendered the table row by row, is that a lot of usage cases see tabulator using rows of unknown and variable height. To do this it rendered each row, calculated its height and then adjusted the rowsHeight calculation to match as it went.
> 
> Your current render math works out the number of rows to show based on the initial avg row height, this works fine when all the rows are the same height, but when tables have massive differences in rows _(some usage cases ive seen have rows that can vary from 20px to 1000px)_ the it could lead to an very over filled or underfilled buffer.
> 
> Any thoughts on how you might deal with this with the new approach?
> 
> ### Branch Change
> Ive moved this PR to merge into the 5.5 branch so that it can be integrated into the other new features for the feature release later this month
> 
> ### Intersection Observer
> I love a good IntersectionObserver, and actually already use one to check table visibility, i would certainty be interested in hearing your thoughts on this in more detail, and certainly in another PR.
> 
> A couple of points to consider on this, we don't want to just render the rows in the viewport there is a buffer on the VDOM is to allow all additions and removals to happen while elements are hidden outside of the viewport of the table, this ensure smooth scrolling when rows are added/removed one by one as the user scrolls.
> 
> The virtualRenderFill function is only called when a whole new set of data is added, the table is fully redrawn, or a person scrolls so fast that they move the viewport down its entire height, the rest of time rows are added and removed one at a time as the table scrolls and to do this efficiently and smoothly it needs the buffer.
> 
> The second point to consider is how you are binding the intersection observers, one or two may result in improvement, but if you have a table with 100,000 rows that could be a lot of observers using a lot of resources.
> 
> But we can cover this in more detail in a separate PR once this one is finished,.
> 
> Let me know your thoughts on the variable height rows issue.
> 
> Cheers
> 
> Oli :)

Thank you for the nice feedback @olifolkerd 

**Passive Event Listeners**
I have reverted this change, sorry for trying to smuggle more changes in. You are right, much better to get the rendered changes stable and in.

**Intersection Observer**
Yup lets pick this up separately

**Variable Height Rows**
You are right here, I actually realised this could cause over rendering when rows were large but I sort of assumed it was not likely.

I am wondering if I can render the first chunk based on the buffer height and then use the heights of the first chunk to work out the average for the next chunk heights.

olifolkerd:
Yeah, rendering of variable height rows happens all the time.

I like the idea of checking after the table height chunk, that would be a good compromise. I won't mind if it is a little bit over the buffer, just as long as it isnt underfilled that's the most important bit as that could result in corrupted render. Given the significant improvement in render time it should hopefully offset a little overdraw.

What we could look at doing is looking at the rowHeight setup option, if that is set then we can safely render out rows without checking height which might offer a performance boost for those that want it.

Cheers

Oli :)

lukecotter:
> 

I have made that changes and hopefully this works. I have changed it to render the first chunk using `vDomWindowMinTotalRows` then it will work out the average row height and use that to work the number of rows needed to fill container height + buffer. We may over fill by `vDomWindowMinTotalRows`.

I also added a check for `this.table.options.rowHeight` which will be used if it exists.

Here are the new numbers, the dynamic height one should be unaffected by the change.
The time was lower in this test for dynamic height so I putting this down to other stuff that was happening on my pc at the same time

**Dynamic Height**
|        | original    | perf     | var height changes |
| ------ | ----------- | -------- | ------------------ |
| create | 850 ms      | 270.60ms | 240 ms             |
| redraw | 400 ms (X3) | 135.12ms | 115 ms             |

The before case (with my data) the render happens 3 times due to adjustTableSize causing the draw to repeat so total is closer to 1000 ms

**fixed height - 100%**

|        | original  | perf      | var height changes |
| ------ | --------- | --------- | ------------------ |
| create | 368.01 ms | 240.96ms  | 250 ms             |
| redraw | 253.12 ms | 135.76 ms | 120 ms             |

**max height - 100%**

|        | original  | perf      | var height changes |
| ------ | --------- | --------- | ------------------ |
| create | 441.53 ms | 280.12 ms | 295 ms             |
| redraw | 200 ms    | 150.10 ms | 170 ms             |

olifolkerd:
Hey  @lukecotter 

Great work!

I will get that merged into the 5.5 branch and will include it for release later this moth.

Thanks for taking on all the feedback there, once it is all merged in I will have a bit more of a tinker and run it through all my standard example to see how they fair.


If you are on the discord channel, drop me a DM and i will give you a contributor role/badge.

Looking forward to seeing your next PR.

Cheers

Oli :)

Issue #4091, opened 2023-01-16
Add header filter support to persistence module


andimg93:
The persistence mode seems to work for all except the filtering. I tried
```javascript
persistence:{
	sort: true, //persist column sorting
	filter: true, //persist filter sorting
	group: true, //persist row grouping
	page: true, //persist page
	columns: true, //persist columns
}
```

and the default, which should activate all available persistence features:
```javascript
persistence: true
```

I talk about header based filtering => https://tabulator.info/docs/5.4/filter#header

olifolkerd:
Please describe in detail what type of filtering you are expecting to persist

andimg93:
well, as I linked - I mean the header based filtering, like in your documentation. 
1. Clicking on demo you go for an example.
2. As an example, based on that demo: If I enter now in the first column (name) "Oli", I get only the result "Oli Bob" and I would expect now that when I refresh the page and persistence mode/config is active that it simply takes into account / remembers the used / entered filters.

olifolkerd:
Hey @olifolkerd 

Im afraid that header filters are not currently available in the persistence functionality, as stated in the docs:

> Note: Header filters are not currently stored persistently, this feature will be coming in a future release.

https://tabulator.info/docs/5.4/persist#filter

Cheers

Oli :)

andimg93:
Oh I completely overlooked that, can you estimate when that will be approximately? That would be a really important feature

olifolkerd:
@andimg93 not sure at the moment, hopefully one of the next few releases. 

Every feature request I receive is a "really important feature" for someone so I have to balance out the needs of the community against my time.

Will change this to a feature request and let you know when it is available

Cheers

Oli :)

andimg93:
I absolutely understand, everybody has his own prio views - in any case many thanks for your effort!

olifolkerd:
Hey All,

I have just merged this in to the 5.5 branch and will include it later in todays v5.5 release.

Cheers

Oli :)

cgountanis:
Going to upgrade today, this is great news, been waiting for this feature.

PR #4156, opened 2023-04-06
Header Filter Persistence support

mind-ra:
I added the support for saving the Header Filter to the persistence mode, as requested in the https://github.com/olifolkerd/tabulator/issues/4091 issue.

I need it for one of my projects and looking throught the Persistence module adding this feature seems to me pretty straigthforward.

I added an option to the Persistence configuration 
```
persistence: {
  headerFilter: true
}
```
It has no options, like the `filter` configuration can be active or disabled.


olifolkerd:
Hey @mind-ra 

Great addition, thanks for the code, i will include this in todays 5.5 release.

If you are on the discord server, drop me a DM and i will add a contributor badge to your user

Cheers

Oli :)

PR #4160, opened 2023-04-14
added a check to ensure that the row exists in the DOM before removin…

melchanti:
I'm using the tabulator in my application but I'm using my own pagination.
When I select on a row, keep the row selected and move to a different page of the table, the tabulator throws an error saying `Cannot read properties of undefined (reading 'remove')`.

After investigating, it seems that the tabulator repo is trying to remove a class from a row that is no longer in the DOM.

This simple change will prevent it from removing a class unless the row is in the DOM.

image


leandrosalgado:
I get the same error but when the editor is trying to remove a validation class:

https://github.com/olifolkerd/tabulator/blob/f9a65455c9bb34ebe96f0136a201b5c446fc1d4e/src/js/modules/Validate/Validate.js#L63

I can in the Validate.js many places with the same code so the same validation could be added.

olifolkerd:
Hey @leandrosalgado 

Thanks for your PR,

That is a great idea you have there, but im afraid you seem to have included the dist files in your  PR, please only include src files in a PR, the dist files are generated prior to a release, and including them in the PR will result in merge conflicts and file overwites when I try and merge things in,

I would normally provide feedback and give you a chance to update, but im putting together the 5.5 release today, so im just going to make the change directly to the 5.5 branch and close this PR.

If you are on the discord server, drop me a DM and i will award your user the contributor badge.

Thanks for your contribution.

Cheers

Oli :)

leandrosalgado:
Hi @olifolkerd,

I'm not the author of this PR. I just mentioned that I get the same error in another part of the code.

I can create a PR to handle this scenario though.

Cheers

melchanti:
thanks @olifolkerd, I sent you a message on discord

PR #4181, opened 2023-05-09
Add option to retain selection in selectable row

### LINKED ISSUE #4178: Selecting text in a cell not possible in selectable row
**Describe the bug**
Selecting text inside a cell in a selectable row selects the row and deselect the text.

**Tabulator Info**
5.4

**Working Example**
https://tabulator.info/examples/5.4#selectable

**To Reproduce**
Just try to select any name in the "Name" column as if you want to copy it somewhere (Select and press Ctrl + C). After releasing the mouse, the row is selected and the name is not. Therefore Ctrl + C wont copy it to clipboard.

**Expected behavior**
I would expect the text to stay selected (highlighted) without selecting the row. Only clicking on the row should select it.

 - OS: Windows 10
 - Browser: Chrome
 - Version: newest

**Additional context**
Is there any workaround for this (except pressing Ctrl + C when mouse button is not released yet)? I would like to have the ability to select cell value of selectable row without actually selecting the row. Thankx :)

nirhu:
Hi Oli,

I opened this PR to try to resolve #4178 by adding new option "retainSelection". What are your thoughts?

Thanks

### DIFF
--- a/src/js/modules/SelectRow/SelectRow.js
+++ b/src/js/modules/SelectRow/SelectRow.js
@@ -90,6 +90,10 @@ class SelectRow extends Module{
 					element.addEventListener("click", this.handleComplexRowClick.bind(this, row));
 				}else{
 					element.addEventListener("click", function(e){
+						if (self.table.options.retainSelection && window.getSelection().toString().length !== 0) {
+							return;
+						}
+
 						if(!self.table.modExists("edit") || !self.table.modules.edit.getCurrentCell()){
 							self.table._clearSelection();
 						}


azmy60 (changes requested):
Hey @nirhu, sorry for the very long delay. A few things I'd like to address:

1. The new option's name feels ambiguous. It'd be clearer if it's closer to the actual intent, which is "preserving the native text selection".
2. Since this is part of the SelectRow module, options here use `selectableRows` prefix (see the `constructor`). 
3. The new option also needs to be registered in the `constructor` of the same class.

I'd suggest something like `selectableRowsPreserveTextSelection`. Open to other naming suggestions :)

rathboma:
Closing this as stale. @nirhu if you get chance to come back to this the contribution would be welcome!

Issue #4190, opened 2023-05-16
Add write options to xlsx download


sergehauri:
*Is your feature request related to a problem? Please describe.**
The SheetJS library allows configuring the file writing, most notably the output format ("bookType") to support file formats other than "xlsx" (https://docs.sheetjs.com/docs/api/write-options#writing-options).
Tabulator sets these options to hardcoded values, but it would be useful if they could be passed in the options argument in the download method.

**Describe the solution you'd like**
The options argument of the .download method supports a property "writeOptions", which will be passed to the XLSX.write method:

```
table.download(
    "xlsx",
    "example.xls",
    {sheetName:"sheet1", writeOptions: {bookType:'biff8', bookSST:true}}
);
```
The writeOptions should passed through to XLSX.write as they are, with the exception that .type is always set to 'binary' (as otherwise the download method will not work correctly).

**Describe alternatives you've considered**
If you want to change the SheetJS output format, there's no way to do it with tabulator right now. You'd have to implement the download function yourself.

The passing of the options could be handled differently, e.g. as a callback function passing the default options object as an argument (similar to "documentProcessing"). This way you could keep the current tabulator default options as they are and just add/update properties on the default options object. In my opinion this would be an unecessary overhead though.

**Additional context**
This change is small and does in no way affect current usage and functionality.

olifolkerd:
Hey @sergehauri 

This functionality will be included in todays 5.5 release.

Cheers

Oli :)

PR #4191, opened 2023-05-16
Add write options to xlsx download

sergehauri:
Feature request #4190


olifolkerd:
Hey @sergehauri 

Thanks for getting in touch, 

That is a great idea, i will merge this into the 5.5 branch and include it in the next release.

Could i ask in future, that you dont declare variable anywhere other than at the top of the function, as otherwise all you end up with is the browser hoisting the variable. You will notice as the top of the function other variables are created based on the options, so for consistency your update should do the same.

If you are on the discord server, drop me a DM and i will award your usr the contributor badge.

Thanks for your contribution.

Cheers

Oli :)

PR #4259, opened 2023-08-04
Fix filtering of non-string values

bodschwa:
The unique column values are stored as keys of the output array of the  _uniqueColumnValues method. When they are used for filtering it has to be ensured that they are compared to strings in order to avoid false negatives.


olifolkerd:
Hey @bodschwa 

Im afraid there are a couple of issues with the PR.

First off you assume that all data is stored as a string. it is not, it could be any data type the user chooses to use.

Second, please do not submit PRs on the dist files, these are generated from the source files and will be overritten at deployment.

Checkout the Contribution gudes for more info:
https://tabulator.info/community/contribute

Cheers

Oli :)

Issue #4269, opened 2023-08-13
header misaligned when using frozen columns group on the right side


ajaborsk:
**Describe the bug**
When using a columns group **and** this group is frozen **and** this group is at the end (right side), the group header position seems not well computed. It looks like the right side of the *group* is at the position of the right side of the *first column* of the group.

**Tabulator Info**
This problem exists, I think, since 5.4 and is still present in 5.5.1

**Working Example**
Here is a minimal example which shows the problem :

https://codepen.io/Alexandre-JABORSKA/pen/yLQmOWO

**To Reproduce**
When scrolling (right/left) the table, the group header is sticky (as expected) but not exactly where it should be. I've put different grouped columns widths to show better

**Expected behavior**
I think the columns group header should be exactly at the right end of the table

**Screenshots**
![image](https://github.com/olifolkerd/tabulator/assets/3057244/cc2a9cd1-9297-4834-a8b8-08bdbc2ec9da)
![image](https://github.com/olifolkerd/tabulator/assets/3057244/326342b4-1903-44b0-b711-ad3420dc6f08)

**Desktop (please complete the following information):**
 - OS: Linux (Archlinux : uname -a = "`Linux jkf21.local 6.4.10-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 11 Aug 2023 11:03:36 +0000 x86_64 GNU/Linux`")
 - Browser : Firefox 
 - Version 116.0.2

**Additional context**
The (mis)computed position is in the element style so it can't be overwritten by a CSS rule. I didn't find any possible workaround...

ajaborsk:
Just tested the codepen.io with Edge and Chrome (on the same platform) and got the exact same behavior.

lukeclimen:
@olifolkerd would you be able to take a look at this PR when you get a chance? I was about to open an issue for the same problem when I saw this one from August. Would be greatly appreciated!

SGreenland:
Came across the same issue, and as a quick workaround I added a cssClass property on the column group object which then sets the style to `right: 0` . If using bootstrap or tailwind should be as simple as adding 'end-0' or 'right-0'.

PR #4270, opened 2023-08-14
Fix right frozen group header position

ajaborsk:
This PR intends to fix the issue #4269 I created yesterday. I simply reproduced the method used for left frozen groups to right frozen groups. Works for me. A little more testing may be necessary, thought there is few risk.

Alexandre

PS : The base code is very clean and clear. Congrats !


azmy60 (changes requested):
Good catch! Looks good to me even though the header line looks slightly shifted (see image below), but that could be fixed in a different PR.

Image

Also, I noticed that the `leftColumns` and `rightColumns` code are super identical. To avoid future mistakes, would you be able to merge them into a single function? At the end, it'd probably look like this:

```js
layoutColumnPosition(allCells) {
  this.layoutFrozenColumns(allCells, this.leftColumns);
  this.layoutFrozenColumns(allCells, this.rightColumns);
}

private layoutFrozenColumns(allCells, columns) {
  // ...
}
```

If all good, will be merging this soon :)

rathboma:
@azmy60 either take this over of close it plz

azmy60:
I'll close this since there's no update. Feel free to reopen if feedback is addressed.

Issue #4280, opened 2023-08-25
bug: incorrect key binding for copyToClipboard on mac


lukecotter:
**Describe the bug**
The current keybinding for macos is `meta + y (meta + 89)` to match windows it should be  `meta + c (meta + 67)` 

**Tabulator Info**
- v5.5.1

**Working Example**
https://jsfiddle.net/Lt810vdk/2/

**To Reproduce**
A step by step guide to recreate the issue in your JS Fiddle or Codepen:
1. Go to the fiddle on mac 
2. focus the table and press `cmd + c` 
3. nothing is in the clipboard and can not be pasted
4. focus the table and use `cmd + y` instead and it will work

workaround is to specify own keybindings
`keybindings: { copyToClipboard: ['ctrl + 67', 'meta + 67'] },`

**Expected behavior**
`cmd + c (meta + c)` should copyToClipboard 

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS ventura 13.5
 - Browser: chrome
 - Version: 116.0.5845.96

**Smartphone (please complete the following information):**

**Additional context**

olifolkerd:
Hey  @lukecotter 

Thanks for the headsup and the fix, i have merged this into mater and will include it in todays patch release.

Cheers

Oli :)

PR #4281, opened 2023-08-25
fix: copy to clipboard key to meta + c

lukecotter:
The current keybinding for macos is meta + y (meta + 89) to match windows it should be meta + c (meta + 67)

fix: copy to clipboard key to meta + c
It was `meta + y` instead of `meta + c`

fixes: #4280


olifolkerd:
Hey @lukecotter 

Thanks for the PR, the concept of your PR is great, but im afraid you seem to have altered all the spacing in the file for some reason.

Please dont change the formatting of files in your PR's, only change the lines of code that actually result in the functionality changes you want.

As im doing a patch release today i will make this change directly myself, otherwise i would normally push it back for remediation.

Thanks for the fix.

Oli :)

lukecotter:
@olifolkerd 
Sorry about that I think my prettier auto formatted.
I will disable it for this project.

PR #4297, opened 2023-09-13
Sanitize keywords string

dylanFdvhuset:
Today if we filter using "Keywords", if we have spaces at the beginning, the end or extra spaces in the middle of a string, the search breaks.

I created this to sanitize the string from extra spaces to keep the essential.

### DIFF
--- a/src/js/modules/Filter/defaults/filters.js
+++ b/src/js/modules/Filter/defaults/filters.js
@@ -55,7 +55,11 @@ export default {
 
 	//contains the keywords
 	"keywords":function(filterVal, rowVal, rowData, filterParams){
-		var keywords = filterVal.toLowerCase().split(typeof filterParams.separator === "undefined" ? " " : filterParams.separator),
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @dylanFdvhuset 

Great idea there, thanks

I will include this in the 5.6 feature release.

Cheers

Oli :)

olifolkerd:
Hey @dylanFdvhuset 

Sorry, with the complexity of getting the cell range module integrated in the 5.6 release,  i have had to push this to the 5.7 release in march.

Cheers

Oli :)

olifolkerd:
Hey @dylanFdvhuset 

Looking into this more, im not sure if this is the right approach, the user can set the separator they want to use, so trimming spaces would be the wrong approach if for example their strings included spaces and they were using a comma seperator.

I think for this to be included it would need to use the separator variable rather than just trim spaces, and would need to be managed through a param rather than be the default behaviour.

Cheers

Oli :)

PR #4301, opened 2023-09-21
Fixing De-Selection after selecting all rows

tim-vogt:
Depending of the source of calling `_deselectRow` the selectedRow in this comparison could be a either a Row or a Row encapsulated into a Proxy (`Proxy(Row)`). You can reproduce this behavior as followed:

### Scenario 1 (working)

1. Select a certain row
2. Deselect that row

Here it will compare `Proxy(Row)` with `Proxy(Row)` and it would behave as expected.

 ### Scenario 2 (not working)

1. Use the Select All feature of the table header, for selecting all rows
2. Deselect one of the Rows from the table

Now it will compare `Row` with `Proxy(Row)` and therefor the the comparison will fail. This is leading to the next code block is completely ignored, the deselection is not applied and the deselection event is not fired.

Although It would be better to find the source for the different contents of `selectedRows` this can be a possible quick fix to avoid that bug.

### DIFF
--- a/src/js/modules/SelectRow/SelectRow.js
+++ b/src/js/modules/SelectRow/SelectRow.js
@@ -354,7 +354,7 @@ class SelectRow extends Module{
 		
 		if(row){
 			index = self.selectedRows.findIndex(function(selectedRow){
-				return selectedRow == row;
+				return selectedRow.position == row.position;
 			});
 			
 			if(index > -1){


olifolkerd:
Hey @tim-vogt 

Thanks for the PR, You have clearly identified the issue here but im afraid your proposed solution is based on properties rather than object comparison so will have accuracy issues in certain circumstances.

With tabulator the proxy element is the public row component and the row is the internal row component, it is possible to retrieve one from the other, so the correct approach here (and the one used everywhere else in Tabulator) is to use the `findRow` function on the row manager to lookup the internal component from the argument and then pass that to the function.

As im about to do the 6.0 release today, i will close this PR and add the relevant fix in now.

Thanks for rasing the issue.

Cheers

Oli :)

olifolkerd:
Hey @tim-vogt

Looking into this further im afraid i am unable to replicate this behaviour it always works for me wether im calling functions or interating with the tabe. Could you please mockup a JS fiddle where you are experiencing this problem?

With tabulator the proxy element is the public row component and the row is the internal row component, it is possible to retrieve one from the other, so the correct approach here (and the one used everywhere else in Tabulator) is to use the findRow function on the row manager to lookup the internal component from the argument and then pass that to the function.

looking at the code both the select and deselect function use the findRow function on the row manager to lookup the internal row objects, so this comparison should always work.

If you have managed to find a reproducible example of where this is not the case i would need to see an example of the table setup to understand things further

Cheers

Oli :)

Issue #4294, opened 2023-09-09
New module: Spreadsheet Mode - Google Sheets style row, column, and cell selection.


rathboma:
This has been brought up before (#3095), but I think it's time to build it.

I plan on having myself, or another Beekeeper Studio team member, tackle this work. So I'm detailing below all the different features this needs.

Broad goal: Column, row, and cell selection should 'just work like Google Sheets'.

## Row selection features:
- [ ] The first column of the table is module-provided, and displays the row number. We call this the 'row header' (similar to the column header)
- [ ] Clicking the row header selects that row
- [ ] Clicking row-header x and dragging vertically  to row y selects all rows that were moused over in range {x..y}
    - Note, the mouseover might move into the table cells, this is ok.
- [ ] Clicking a row header x, then shift-clicking row-header y selects all rows in the range {x..y}
- [ ] Clicking a row header x then ctrl/meta cliking row-header y selects rows {x, y}
- [ ] Ctrl/Meta + C when rows are selected copies them to clipboard in an excel-friendly format

https://github.com/olifolkerd/tabulator/assets/279769/8a26e8af-5475-42fb-8dae-2260e29abede

## Column selection features
Column selection behaviors are the same as row selection behaviors.
- [ ] The column header is modified such that clicking the header does NOT sort by that column, instead a button is added to the right of the column title for actions such as sorting and filtering
- [ ] Clicking the column header selects that row
- [ ] Clicking col-header x and dragging horizontally  to col y selects all cols that were moused over in range {x..y}
    - Note, the mouseover might move into the table cells, this is ok.
- [ ] Clicking a col header x, then shift-clicking col header y selects all cols in the range {x..y}
- [ ] Clicking a col header x then ctrl/meta cliking col-header y selects cols {x, y}
- [ ] Ctrl/meta + C when cols are selected copies them to clipboard in an excel-friendly format

https://github.com/olifolkerd/tabulator/assets/279769/961687e3-9480-4cb7-a2c3-5fc105bb166c

## Cell selection features
This is the most complex behavior. It behaves with similar primitives to the row and column selection
- [ ] Clicking a cell selects that cell (easy)
- [ ] Clicking cell a,x and dragging horizontally or vertically  to cell b,y selects all cells that were moused over in range {a..b, x..y}
    - Note, that this is a rectangle of cells
- [ ] Clicking a cell a,x, then shift-clicking cell b,y selects all cells in the range {a..b, x..y}
[...]

PR #4313, opened 2023-10-04
Spreadsheet module

azmy60:
In attempt to support spreadsheet mode as mentioned in issue #4294 

Animation4

Usage:

```js
import { TabulatorFull } from 'tabulator-tables';

const table = new TabulatorFull("#table", {
  spreadsheet: true, // enable it
  clipboard: true, // will copy the active range using via Clipboard module
  spreadsheetRowHeader: { // optional: row header column*
    field: '--row-header',
    contextMenu: [...],
  },
});

// get a range component if exists
cell.getRange()
row.getRange()
column.getRange()

range.getData()
range.getCells()
range.getStructuredCells() // get cells in 2d array (row x column)

table.getSelectedData() // get data from the active range
table.getActiveRange()
```

* Row Header is just a new column and same as the other columns. You can override the default definition by following the [Column Definition](https://tabulator.info/docs/5.5/columns#definition).

Notes:
- Enabling Spreadsheet module would disable SelectRow module and disable the `headerSort`. Clicking the column header would select the cells of the column.
- A range refers to a selection of cells. An active range is the last range registered.


olifolkerd:
Hey @azmy60 

Love it!, im just catching up on things after a couple of months away from the project, but will review your PR in the next week or so.

Loving the video though, looks great.

Cheers

Oli :)

azmy60:
Hey @olifolkerd !

I have a little follow up commit that'll be coming up tomorrow, and then I'll mark this as ready to review.

Also, looking forward to your feedbacks! Thanks! :)

rathboma:
I've been using spreadsheet mode all morning.

Overall: It's amazing, and behaves exactly as I'd expect. Great stuff. It even keeps selected ranges when scrolling in a virtual table. Great work on that.

Some bugs:
1. Tabulator is much much slower when loading data, or paginating a virtual table when using this plugin. (in BKS: try select * from film, then scroll)
2. Resizing columns is very very slow also (in Beekeeper try this with the `film` table, it basically hangs)

Some missing features that I only notice now I'm using it
3. Arrow key support: up/down/left/right should move the selected cell around
4. Shift+arrow - selects a range, starting at the initial cell (test this in google sheets)
5. Ctrl+arrow - moves the cursor to the next non-empty position in the arrow direction (again see in google sheets)
6. Ctrl+shift+arrow - selects the range up until and including the final non-empty cell {current cell: last non-empty cell}

olifolkerd:
Hey @rathboma thanks for the nudge this one had slipped off my todo,

Your notes on user interaction seem spot on, and should be handled through the keybiding module and the internal event bus.

I will carve out some time this week to check through each of the events, but from a quick skim i can see a few of the render events could be handled through the built in helper function that come with the module, there are also more optimal ways to refresh the DOM when you are only looking to redraw certain parts, which might be behind some of the performance issues you have noticed.

I will have a skim through and put togeather some programatic recommendations, probs by monday eve at this point.

If we can work to polish this up i would love to include it in a november release :)

Cheers

Oli :)

olifolkerd:
Hey @azmy60 

So i have had a chance to look over things, and first off id like to say great work this is a really intersting idea and has been well implemented, i have a few points of feedback to disucss, a bit on coding style to match the library, some feedback on how to make the module more efficient and some ideas for further extensions to bring it even more into the Tabulator ecosystem, that i would be happy to help with.

In this post i will talk about some changes that would need to be made to bring the PR inline with Tabulator coding style, in the next post i will talk about some possible further extensions

### Setup Options Naming Conventions
To be consistent with the rest of the tables setup option could you split the configuration options for the row header into their own options starting with the module name, so for example it should be `spreadsheetRowHeaderField` that would take the name for the row header., the same for any context menu options etc

### Code Formatting

#### Variable Definitions
You are defining variables throughout the functions of your code, you will notice from the failures in the deepscan checks (see the orange box at the bottom of this chat stream) that variables MUST be declared at the top of the function in which they are used and then referenced when needed.

If you are using custom variables inside of a for loop then these must be of type `let` to avoid thier scope being acceessible outside the loop, and should also be declared at the top of the loop.

In the case of you swich statements, please define your 'nextRow' and 'nextCol' variables at the top of the function and then reference them in the swtich.

#### Component Module Variables.
I notice you are adding custom properties directly to the column component object, please do not do this as it isnt safe, there is a modules object defined on the column object where modules can safely store custom variables, the convention is to add a property to this object of the name of your variable and then store the data in there, eg:
```js
column.modules.spreadsheet = {
    editable:true,
}
```
You can then access and edit the `column.modules.spreadsheet` object as needed and it will be safely persisted on the column component

This is also true of rows if you need to store data on a row.

as a result of this tou shouldnt need to use the 'this.registerColumnOption' calls in the constructor. these should be used to register properties expected in the column definition, not on the internal component.

#### Column Editor
Could you confirm what you are aiming to do with the attempt to remove the editor, Im unsure what the current approach is intended to do, but it wont interact with the editor module with the current setup. if you could advise your intention i would be happy to suggest the correct approach


#### Table Header Tab index 
I notice that you have updated the 'ColumnManager.js' to set a tab index on the 'table-header' element, could you please make this change in your module as part of your `initializeTable` function rather than in the column column manager itself. this change could break some of the accessibility navigation in other modules so should only be made if the spreadsheet module is enabled. you can access this element on `this.table.columnManager.element` inside your module

#### Render event
You appear to have added a `render-virtual-dom` event in a great many positions throughout the code base, this will be triggering a huge number of unnecessary redraws which will be behind a lot of the performance issues that we are experiencing as you are triggering a full redraw on event a small incremental update of the DOM, the correct approach is to update only the rows/columns that have been added if they sit in one of the ranges. Could you clarify the intent behind this, and i can outline the correct approach using exiting  events to render this smoothly.

There appears to be an issue when using the virtual dom (set a height on the table) that the formatting gets cleared, and i think that is also directly tied to the current approach, but the good news is it should be easy to fix :)

If you like, id be happy to have a zoom call to take you through the backend of Tabulator and discuss some next steps

Thanks for the great work!

Cheers

Oli :)

olifolkerd:
Hey @azmy60 

In addition to my last post, i think there are a few small additions we could make to really make this new module fit in well to the Tabulator ecosystem, and im sure @rathboma would have some additional ideas too

### Row Header Menus
I notice in your example you are mentioning a context menu (though i cant see it in the code of the PR itself) If you want to add in menu options for the row header that is no problem, but that wants to sit inside the `Menu`  module which manages all menus and will need to implement the click, dblClick and context menu options to be consistent with the other areas where menu clicks occur.  _(I would be more than happy to help add this in myself if you would like)_

### Row Header Interaction Events
It may also be worth adding in the standard interaction handlers for the row header in the core `InteractionMonitor` tool to make them accessible to users too,  _(I would be more than happy to help add this in myself if you would like)_

### External Events
You should also consider adding some external events that are triggered when the spreadsheet selection is changed that pass out the selected cells.
for example:
```js
this.dispatchExternal("spreadsheetRangeChanged", ranges, data);
```

Would be happy to chat through these on a zoom call too

Cheers

Oli :)

azmy60:
Hey @olifolkerd,

The feedbacks are awesome! Thank you very much for the reviews! I'm very happy to take all the suggestions and feedbacks since this library is huge and we don't want to break other functionalities. It's very extensive and I learn a lot from it. Again, thank you.

As to your questions:

## Column Editor

The main intention of removing column `editor` and `editable` temporarily was for allowing us to trigger the cell edit with a double click instead of a single click. Without removing the `editor`, the cells edit will always trigger by a single click after it gets a double click for the first time. This is kind of a hack, so I'm aware this is not a good approach :D. I'd be happy to get your ideas on this.

## Keybindings on Edit Module

This change is kind of unrelated to spreadsheet. I commented the left and right keys because these were never triggered before. These should triggered by `navLeft` and `navRight` which didn't exists in the `src/js/modules/Keybindings/defaults/bindings.js`. 

But even though we have them enabled, they don't respect the editor inputs.

Navigating with left and right arrows

Notice that, after I click the cell, I want to go to the previous character by pressing the left arrow. But it doesn't work because the Edit module prevent that using the `navLeft` event. This can be reproduced even without the spreadsheet mode.

## Rendering & Debouncing

The way that range rendering works is that it simply takes two cells - a top left cell and a bottom right cell - and then draw the range based on where the two cells are. When the virtual dom is enabled, and the cells don't exist in the document yet, it will try to find the other two cells that are aligned and exist in the document.

image

The `render-virtual-dom` events are added so I can update the range after every rows or columns are added/removed from the document. Due to that event being called very frequently, I need a debounce helper to limit it. My first approach was to add something like `scroll-vertical-stopped` event, but I wasn't sure where to add that. This is something I'm aware that it's not  good, so I'd love to get your suggestions on this.

## Further Discussions

And yeah, of course. It'd be great to discuss about this further in a zoom call. You can email me at mazmiagus2003@gmail.com :)

In the meantime, I'm going to update the codes with the other feedbacks you mentioned.

azmy60:
@olifolkerd  Sorry for my ambiguous example. What I meant was I put a `getRange()` function to each cell/row/column component, which can be used in a situation like getting data from `contextMenu`. No need to extend the `Menu` module.

```js
contextMenu: [{
  label: "Copy as JSON",
  action(e, cell) {
    const range = cell.getRange()
    copyToClipboard(JSON.stringify(range.getData()))
  }
}]
```

rathboma:
@olifolkerd we merged our branch into Beekeeper Studio, and it's working great. How much more work do you need before this can be merged?

It makes me nervous not being on an official release of Tabulator :-S

olifolkerd:
Hey @rathboma @azmy60 

Thanks so much for all your hard work and patients there. Ive just done a patch release to start catching up on some bits while i was away.

Im going to look to work through a bit more of the backlog to iron out bits that have been found while i was away and then get this merged into a feature release.

Cheers

Oli :)

olifolkerd:
Hey @azmy60 

Thanks for all the continued work on this PR, im putting togeather the 5.6 release at the moment and would like to include this in that release.

I have created a 5.6-spreadsheet branch and merged this PR into it, there are few tweaks im going to need to make to bring this inline with other modules, so i will make them on this branch and test them before it gets merged in.

I will stick some feedback on here when im done with the tweaks so you can see what was needed for the production release.

Cheers

Oli :)

rathboma:
@azmy60 please see Oli's comments above, can you review his branch then create new PRs for the other functionality that we added along the way?

@olifolkerd the bundling of features is my fault really as I asked azmy to do it this way to keep the Beekeeper fork simple for us to integrate into the app given how long the branch was open.

@olifolkerd do you have a list of the features that you removed so we can make sure we capture each one as an individual PR?

olifolkerd:
@rathboma The main ones I have removed is the column resize functionality and the rownum pagination updates (the last one needs some work as it doesn't account for grouped row scenarios)

I've also completely refactored the codebase of the pr to better isolate the module from the rest of the table (rolled back most of the tweaks to other modules and then implemented them in this module to interact with them through events, as most of the code that reached outside this module broke other table configs), simplified the code, bring it inline with tabulators coding style and break it out into responsible components, the main module file was enormous!

As well as this I've also removed the concept of it as a spreadsheet module as it was trying to reconfigure large chunks of the table, and that was really pushing beyond the scope of an isolated module. This has been replaced with the range selection module which has the same functionality as this but I have broken down the functionality into a series of options that can be enabled per table to give the developer a load of choices over how the module functions as well as added a load of new functionality to apply it to a bigger range of scenarios.

Ive almost finished on my updates and will merge them into the 5.6 branch later this week. I would suggest that all further PR's be based on the 5.6 branch once that merge is complete, especially if they affect selection, otherwise I won't be able to merge them in

Cheers

Oli :)

olifolkerd:
I have merged these changes into the 5.6 branch so it should now be safe for you to add other changes there as separate PR's.

Im happy to run you through an changes ive made if you have questions.

Cheers

Oli :)

azmy60:
Hi @olifolkerd , Thanks so much! I'll go through the feedbacks and make the PRs soon!

olifolkerd:
I have also just added some cool new range paste functionality that will allow for paste behaviour similar to a spreadsheet, in terms of allowing pasting a value across multiple cells

olifolkerd:
Hey @azmy60 @rathboma 

Version 5.6 has just been released! 

Including [Cell Range Selection](https://tabulator.info/docs/5.6/range#overview)
and an example of how to [Setup the table as a Spreadsheet](https://tabulator.info/docs/5.6/range#spreadsheet)

Checkout the [Release Notes](https://tabulator.info/docs/5.6/release) for full details of all the tweaks and features ive added on top of this feature :)

Would love to hear your thoughts on it all.

Cheers

Oli :)

rathboma:
Great docs Oli! Looking great.

olifolkerd:
Ive just added some tweaks to the 6.0 branch that add some functionality with the range selection module:

- The concept of the "row header" is now built in as standard to the table core functionality and is available to all modules
- Modules now have the ability to inject extended config options directly into other modules from within the original modules codebase, so it is easier than ever to confine a modules functionality for its own code, for example the range module can now inject its keybindings, actions, clipboard range functionality and export functionality all from within the range module
- The range module now supports right to left text direction

I'm working on a few other exciting additions i think you might like over the next few days, will give you a shout when they are in.

Cheers

Oli

rathboma:
Awesome!

Issue #4364, opened 2024-01-11
Incorrect table rendering if the element has any transforms applied


AdiDD:
**Describe the bug**
If the DOM element on which we create the `Tabulator` has any transforms applied while the table is being drawn, it will not render properly. (table's width and header's height)

**Tabulator Info**
5.5.2

**Working Example**
Open the modal, and click on a table row in order to redraw it while no `transform` is being applied to the element.
https://jsfiddle.net/asce90/jxdv1Lpn/42/

**To Reproduce**
1. Create an element that has initially `scale: 0;`
2. Set `scale:1` and initialize or redraw the table

**Expected behavior**
The tabulator should have the element's layout width and height, instead it had the element's rendering width and height.

**Screenshots**

- how the table renders 
![image](https://github.com/olifolkerd/tabulator/assets/16980592/9ce4466a-74ed-461f-b370-ec34bdbe81f7)
- how it should render 
![image](https://github.com/olifolkerd/tabulator/assets/16980592/784ce1a7-566a-4a3b-b0e8-912a32a2ab58)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 120.0.6099.217

AdiDD:
I investigated the problem, I will submit a PR shortly.

olifolkerd:
Hey @AdiDD 

Thanks for the PR, but your suggested solution will break the rendering in other ways im afraid.

Have you tried applying the transform to a parent element of the table rather than the table itself?

Cheers

Oli :)

PR #4365, opened 2024-01-11
fix initial height and width measurements on table draw

AdiDD:
Fix for https://github.com/olifolkerd/tabulator/issues/4364. 
When calculating the table size, we should use the layout with and height, not the rendered width and height.


olifolkerd:
Hey @AdiDD 

Thanks for the PR, unfortunately this will break other calculations used by the table (this has been done before and had to be reverted, so im afraid a different approach will be needed

Cheers

Oli

Issue #4403, opened 2024-02-19
More columnResized events


lofcz:
Currently, only `columnResized` is exposed which fires once the interaction is done. Our header filters are running on a custom JS virtual-select engine and to properly synchronize the position of the dropdown menus we need to run some code every time the mouse is moved during the resize interaction. Exposing `columnResizeStart` would be sufficient for us to add a mouse-moved event listener, run the code we need, and dispose of the listener once the interaction is finished.

**Describe the solution you'd like**
Add `columnResizeStart` event.

**Describe alternatives you've considered**
Keep the current behavior as with the combination of `headerMouseMove`, `headerMouseDown`, and `headerMouseUp` this can be  emulated albeit not with a 100% precision in v5.6

olifolkerd:
Hey @lofcz 

Thanks for the PR,

I can see where you were going with your PR, but im afraid the implementation wont work correctly based on the positioning of those events.

As it happens i have already been tinkering in that module, merging in some new functionality that i think would make these events unneeded. new resize guide functionality is coming in the 6.0 release that will allow you to resize a column using a guide placeholder without resizing the column itself till after the resize is complete, which will mean the existing `columnResized` event should work nicely for you

Having said that i can see a few scenarios where some additional events could be useful, so i have also just pushed to the 6.0 branch the following events:
- `columnResizing` - A column has started to be resized
- `columnWidth` - A columns width has changed (from any method of changing column width, not just resize module)
- `rowResizing` - A rowhas started to be resized
- `rowHeight` - A rows height has changed (from any method of changing row height, not just resize module)

Cheers

Oli :)

PR #4404, opened 2024-02-19
Add missing column resize & move events

lofcz:
Adds two events: `columnResizeStart`, `columnResizeMoved` and `columnMoveStart`. The need for them is explained in #4403.
I'm open to renaming the events to meet the conventions used elsewhere in the project.

With the added events we were able to integrate https://sa-si-dev.github.io/virtual-select in the header columns.

### DIFF
--- a/src/js/modules/MoveColumns/MoveColumns.js
+++ b/src/js/modules/MoveColumns/MoveColumns.js
@@ -201,6 +201,7 @@ class MoveColumns extends Module{
 		}
 		
 		this.moveHover(e);
+		this.table.externalEvents.dispatch("columnMoveStart", column);
 	}
 	
 	_bindMouseMove(){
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @lofcz 

Thanks for the PR,

I can see where you were going with your PR, but im afraid the implementation wont work correctly based on the positioning of those events.

As it happens i have already been tinkering in that module, merging in some new functionality that i think would make these events unneeded. new resize guide functionality is coming in the 6.0 release that will allow you to resize a column using a guide placeholder without resizing the column itself till after the resize is complete, which will mean the existing `columnResized` event should work nicely for you

Having said that i can see a few scenarios where some additional events could be useful, so i have also just pushed to the 6.0 branch the following events:
- `columnResizing` - A column has started to be resized
- `columnWidth` - A columns width has changed (from any method of changing column width, not just resize module)
- `rowResizing` - A rowhas started to be resized
- `rowHeight` - A rows height has changed (from any method of changing row height, not just resize module)

Cheers

Oli :)

PR #4410, opened 2024-02-22
fix broken selection when the first column is invisible

azmy60:
When starting tabulator with `SelectRange` enabled, the selection is not on the top left. This is because we have a table that the first column is invisible and that column get processed, which shouldn't.

Demo: https://jsfiddle.net/zqd0kgwr/


olifolkerd:
Hey @azmy60 

Thanks for the fix, i will include it in the 6.0 release.

Cheers

Oli :)

azmy60:
Ah, I see why that needs to be on top. Thanks!

PR #4411, opened 2024-02-22
fix broken range UI when using SelectRange and virtual horizontal renderer

azmy60:
This PR fixes the issue of range element that doesn't cover all selected cells if all columns are selected. This happens if we have a frozen column (row header) and the virtual horizontal renderer is enabled. In this image, I selected all columns but notice that the range lines don't cover the `internal` column.

image

Demo: https://jsfiddle.net/fo97hyrk/

Also, I added some warnings for folks to be cautious to use frozen columns along with `SelectRange`. For now, I think it's best to have a frozen column just for row header.

### DIFF
--- a/src/js/modules/SelectRange/Range.js
+++ b/src/js/modules/SelectRange/Range.js
@@ -165,6 +165,10 @@ class Range extends CoreFeature{
 		_vDomLeft = this.table.columnManager.renderer.leftCol,
 		_vDomRight = this.table.columnManager.renderer.rightCol,		
 		top, bottom, left, right, topLeftCell, bottomRightCell;
+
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @azmy60 

Thanks for this, im working on the 6.0 release with some cool new features and will include this in there.

Nice idea on the warnings, its good to let people know of misconfigurations.

One small tweak i am going to make after i merge things in is to check for the frozen column module and asses if there are frozen columns in there as it will give a more accurate count of whether frozen columns are actually present and in-place. For example if the module is missing the definition wont affect things etc.

Cheers

Oli :)

azmy60:
Right, that's very fair.

PR #4414, opened 2024-02-26
Fix bug of getRowFromPosition function when groupComponent exists.

lstgithub578:
Fix the bug of `getRowFromPosition` function in `RowManager`
When the result returned in `getDisplayRows ()` contains `GroupComponent`, an error will be reported.
image


olifolkerd:
Hey @lstgithub578 

Thanks for the fix, id say using range selection on a grouped table is a weird choice, but this does undoubtably fix the issue so i will include it in the 6.0 release.

Cheers

Oli :)

Issue #3654, opened 2022-03-10
Tabulator vertical scroll skipping / jumping


jarb99:
Hi There,

I am a big fan of Tabulator, thanks for the great work! There is a scrolling issue I have not been able to overcome and cannot see another bug report for. This is the only issue we have found that is preventing us adopting in our app so super keen to resolve.

**Description:**

- When scrolling up in a table with several rows, the table skips suddenly jumping down.
- This behavior seems to be triggered by a window / container resize although not always. 
- This behavior can be reduced _some of the time_ with **table.redraw(true)** as recommended in the docs -This is a problematic band aid as it will reset the scroll to the top of the table which interrupts user experience
- The tabulator internal debug log reports this as a single downwards scroll event, despite no downward scroll input from the user.
- The skipping is much worse for rows that are spread over multiple lines using ‘textarea’ or custom formatter (our intended use).
- I have reproduced this on multiple devices, Firefox, Edge & Chrome.

**Tabulator Info:** version 5.1

**Desktop:** Windows 10/11, OSX

**Working example:**
•	I have found this issue appears on the [Tabulator website](http://tabulator.info/examples/5.1) in the vertical DOM example [(Gif of behaviour attached)](https://drive.google.com/file/d/16iYA1q0RT4f0eqqR48cDSDaIYnUXtIoq/view?usp=sharing). This seems easier to reproduce behavior than the code sandbox example.
•	Codesandbox: https://codesandbox.io/s/still-hooks-39032m?file=/src/App.js

![Animation](https://user-images.githubusercontent.com/76447823/157566800-3d3d3c03-0b06-483c-b023-36788a53268d.gif)
gif showing example of vertical scroll skipping after window resize

**Steps to reproduce**
- Go to tabulator example website, Virtual DOM – vertical example.
- Scroll to bottom of table.
- Resize browser window horizontally, forcing tabulator container to resize
- Start scrolling up. Sometimes it takes a couple of goes to reproduce the behavior. Is most evident using the middle mouse button click to lock smooth scrolling in direction of cursor.

**Expected behaviour**
- Smooth scrolling. Perhaps there is an internal fix Tabulator could incorporate which detects this scroll misfire

gerric:
I have also found this issue. It is even easier to see when you're scrolling using the scrollbar and moving large steps. It will randomly lock into the last position the table was fully redrawn.

jarb99:
[...]

PR #4425, opened 2024-03-07
Fix incorrectly calculated height on collapes rows

antman3351:
Hey Oli, 

as mentioned in the issue #3654 ( So i'm going to copy and paste a bit 😜 )

I noticed in the VirtualDomVertical module in _removeTopRow the row height is always the same even though the rows are not the same height.

I think the issue is the column height isn't recalculated after the rows are collapsed, so I've added a call to `calcHeight()` after the rows are collapsed.

Cheers,
Antonio

### DIFF
--- a/src/js/modules/ResponsiveLayout/ResponsiveLayout.js
+++ b/src/js/modules/ResponsiveLayout/ResponsiveLayout.js
@@ -249,6 +249,7 @@ class ResponsiveLayout extends Module{
 			if(contents){
 				el.appendChild(contents);
 			}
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @antman3351 

Thanks for the PR, it definitly should recalc row height after the collapse event, so thanks for that, but this wont fix the issue outlined in #3654 which is a wider issue to do with how the vdom handles adjustment of variable height rows in its calculations.

I will merge this in as it should certainly be included

Cheers

Oli :)

PR #4431, opened 2024-03-13
fix movable columns and range selection

azmy60:
This PR should allow us to have movable columns and also column range selection. Currently, enabling both options would results in weird behavior.

DEMO (to replicate): https://jsfiddle.net/4pqox6gu/

mygif

### DIFF
--- a/src/js/modules/MoveColumns/MoveColumns.js
+++ b/src/js/modules/MoveColumns/MoveColumns.js
@@ -209,7 +209,7 @@ class MoveColumns extends Module{
 		
 		this.moveHover(e);
 
-		this.dispatch("column-start-move", e, this.moving);
+		this.dispatch("column-moving", e, this.moving);
 	}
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @azmy60 

Looking good there, one quick tip on naming, the Tabulator convention is to use "ing" for when a thing starts and "ed" for when it ends

so it would be `column-moving` and `column-moved` if you would be ok with the changes. There is already an existing `column-moved` event, would the exiting one work for you without needing your new one? it is triggered at about the same time, just from a different point in the code?

Cheers

Oli :)

azmy60:
Hey @olifolkerd ,

Yeah, that's fine. The code still works with the existing event.

marsrover4:
could someone please merge this? 0:)

marsrover4:
Thanks for the merge! :)

PR #4452, opened 2024-04-04
make autoFocus optional for SelectRange

azmy60:
We recently found a bug related to SelectRange.

https://github.com/beekeeper-studio/beekeeper-studio/issues/2075

Basically, text editor (codemirror) outside tabulator becomes unusable when tabulator tries to gain focus. I think it's related to `document.createRange()` in SelectRange and the way that codemirror didn't handle that properly I guess (?).

So, I think it would be great to make the `autoFocus` optional.


rathboma:
Thanks Azmi!

olifolkerd:
Hey @azmy60 

Thanks for this, as it is a new option i will include it in the next feature release.

Cheers

Oli :)

olifolkerd:
Hey @azmy60

Cheers

Oli :)

PR #4456, opened 2024-04-04
Node bump for lint pipeline

rathboma:
- rollup doesn't support node 14 anymore

This should fix the linting pipeline


olifolkerd:
Hey as of version 6 of tabulator it has a minimum node version of 20

olifolkerd:
Ahh with you, update to linter config. Should have checked code first lol

rathboma:
Also - Beekeeper is still running on Node 16 due to Electron, so don't break too much for us :-)

PR #4457, opened 2024-04-04
fix navigation with rowheader

azmy60:
Navigating to left with keyboard can make it select the `rowHeader`. 

Here's the demo to replicate: https://jsfiddle.net/dbjzcw0e/

image

- Try using keyboard to navigate to `rowHeader`
- Compare that with clicking `rowHeader` with mouse

To fix this, we can check if `rowHeader` exists in SelectRange. Range is not responsible for checking `rowHeader` for the most part anymore. Also, that seems to simplify the code a lot.


azmy60:
This PR causes other bugs. Closing in favor of #4459

PR #4478, opened 2024-04-22
Break expanded range after pressing arrow keys

azmy60:
**To reproduce:**
- Select all cells by pressing `Ctrl+Shift+Arrow`
- Press Up Arrow without `Ctrl+Shift`

**Expected:** Only one cell should be selected and it appears in the view
**Actual:** It does nothing


olifolkerd:
Hey @azmy60 

Thanks for this suggestion, id be happy to include this, but as ive mentioned before the coding style for Tabulator does not allow for getter properties.

If you would move the rowHeaderWidth getter into a function i would be happy to include it.

Cheers

Oli :)

azmy60:
@olifolkerd Thanks for the feedback! Just fixed the getter and merge conflict as well.

azmy60:
Hey @olifolkerd ! Would you take a look when you have time to review this? Thanks! :)

olifolkerd:
Hey @azmy60

Cheers

Oli :)

PR #4481, opened 2024-04-23
BasicVertical renderer restores scroll position in rerenderRows

### LINKED ISSUE #4469: Progressive Load: Scroll on Basic Renderer jumps to top while scrolling
**Describe the bug**
Basic Vertical Renderer bounces to top when using Progressive Load Scroll. This happens when new data is fetched and added from the server.

**Tabulator Info**
- Tabulator 6.2.0

**Working Example**
https://jsfiddle.net/qsk5zLwh/2/

**To Reproduce**
1. Construct table with `renderVertical:"basic"` and `progressiveLoad:"scroll"`
2. Connect table to remote source with enough rows to require multiple requests
3. Scroll table towards bottom
4. Notice table jumps to top and scroll bar shrink as new rows were added

**Expected behavior**
Smooth scrolling. Fetching the next page should not reset the scroll position.

**Desktop:**
 - OS: Windows 10
 - Browser Chrome
 - Version 123.0.6312.122

**Additional context**
BasicVerticalRenderer `rerenderRows` clears all rows then renders them again. This shrinks the container element and resets the scroll position. Documentation for renderers states that `rerenderRows` should maintain scroll position.

Have encountered scrolling issues using the default virtual dom renderer like #4353 and #3654 both of which are resolved by switching to the basic renderer. 

I have used the suggestion in #4424 and made the basic renderer's `rerenderRows` remember the `scrollTop` before clearing and then reapplying it after rendering to resolve the issue.

Bobcat64:
Makes the `BasicVertical` renderer remember the scroll position when `rerenderRows` is called so it can be restored after clearing and rendering the rows.

This fixes #4469  when Progressive Scrolling is used with the Basic renderer, the table scrolls back to the top when a new page is loaded. This is because all rows are cleared and re-added.

Side note/Suggestion: The documentation at https://tabulator.info/docs/6.2/render#vertical for `rerenderRows` says that the rows should be cleared AND rendered before the callback is called. However the current code for both the Basic and Virtual renderers clear then call the callback and then render the rows. Might be prudent to update the documentation to reflect this.


olifolkerd:
How does this resolve the issue? 

In basic render mode the table generally doesn't have a height set and so the scroll is not on the table element but usually on the page body and it is the redrawing of the content that results in the loss of scroll position. Which changing the internal scroll position of the table will not resolve?

Cheers

Oli

Bobcat64:
It resolves it when the table element does have a scroll position. Such as when the height is explicitly set like in the example in the linked issue.

Only way I can think to resolve your scenario with the page itself scrolling would be to temporarily add some padding or something before the rows are removed so the total height doesn't change. I'm however unsure if that would just cause more performance issues with the browser redrawing. (I am not well versed in when/what causes browser re-renders)

I suppose I'm also a bit at a loss why adding new rows to the end should end up with the basic renderer clearing and recreating all the rows. Perhaps there should be a method on renderers to update/render specific rows (or just new ones)? This would allow the basic and/or virtual renderer to determine if it needs to draw/add elements as necessary. 

Is there an issue with setting the scroll position to its previous value after rendering the rows?

azmy60:
Hey @olifolkerd , just wanted to check in on this PR. Could you take another look when you have time?

rathboma:
@azmy60 take this over or close it plz

Issue #4378, opened 2024-01-28
Tabulator ESM with Sveltekit and Vite


tadmi:
Tabulator 5.5.4
Sveltekit 2.5.0 (built with Vite 5.0.12)

Possible problem
![image](https://github.com/olifolkerd/tabulator/assets/63556979/b3a257bb-9b5f-44b1-9e2c-3a540f5508f5)

and solution: add in tabulator-tables package.json **type: module**.

The problem appeared with Vite 5.0.8:
specific change — fix(ssr): check esm file with normal file path (#15307) (1597170), closes #15307

olifolkerd:
Hey @tadmi ,

The challenge with your proposed solution is that Tabulator is not just a module, it offers both ESM and UMD options, which is why you will see in the package.json that it offers both a `main` and a `module`.

This works by allowing your task runner to pick whichever version is correct depending on whether you are running as a module  or not.

You may need to update your projects package.json to handle this. If i add your proposed solution, it will break it for all UMD users.

I use tabulator in many ESM projects with the current package.json without any issues.

Cheers

Oli :)

novecento:
sorry: what does it mean "You may need to update your projects package.json to handle this." ?

novecento:
If I use 
import * as TabulatorFull from 'tabulator-tables';
const { Tabulator } = TabulatorFull;

Tabulator doesn't work well (ie tree not works)

If I use 
import {TabulatorFull as Tabulator} from 'tabulator-tables';

Error when evaluating SSR module /src/lib/components/common/Tabulator.svelte: failed to import "tabulator-tables"
|- SyntaxError: [vite] Named export 'TabulatorFull' not found. The requested module 'tabulator-tables' is a CommonJS module, which may not support all module.exports as named exports.

apologies if it's a stupid question..

thisjt:
I think we need to reopen this issue. Here's a minimum reproducible repo of the issue using SvelteKit.

(File is in `src/routes/+page.svelte`)
https://github.com/thisjt/tabulator-import-issue

Here is the result on dev run
![image](https://github.com/olifolkerd/tabulator/assets/64672207/b35330ad-7466-431c-8a78-76f453441869)

I didn't have this issue when I was on Vite 4, but when I was forced to update to Vite 5 as it is a requirement for SvelteKit 2.0 due to Vercel recommending me to migrate from SvelteKit 1 to 2, this issue started showing up on my imports.

thisjt:
[...]

PR #4490, opened 2024-05-05
Add Require and Import Values in `package.json` for better `ESM` Support

thisjt:
`build/Bunder.mjs` has been changed so that `tabulator_esm` output in `dist/js` will have the `.mjs` extension.

`package.json` has been changed so that the file that the bundler creates is properly referenced.

Referencing bug report: https://github.com/olifolkerd/tabulator/issues/4378

### DIFF
--- a/build/Bundler.mjs
+++ b/build/Bundler.mjs
@@ -1,180 +1,181 @@
-import { createRequire } from 'node:module';
+import { createRequire } from "node:module";
 const require = createRequire(import.meta.url);
 
 import { nodeResolve } from "@rollup/plugin-node-resolve";
 import terser from "@rollup/plugin-terser";
 
-import license from 'rollup-plugin-license';
-import {globbySync} from 'globby';
-import fs from 'fs-extra';
+import license from "rollup-plugin-license";
+import { globbySync } from "globby";
+import fs from "fs-extra";
 
 import postcss from "rollup-plugin-postcss";
 
-export default class Bundler{
-	
-	constructor(version, env){
+export default class Bundler {
+	constructor(version, env) {
 		this.bundles = [];
-		
+
 		this.env = env;
-		this.version = "/* Tabulator v" + version + " (c) Oliver Folkerd <%= moment().format('YYYY') %> */";
+		this.version =
+			"/* Tabulator v" +
+			version +
+			" (c) Oliver Folkerd <%= moment().format('YYYY') %> */";
 	}
-	
-	_suppressUnnecessaryWarnings(warn, defaultHandler){
+
+	_suppressUnnecessaryWarnings(warn, defaultHandler) {
 		const ignoredCodes = {
-			"FILE_NAME_CONFLICT": true,
+			FILE_NAME_CONFLICT: true,
 		};
-		
-		var suppressed = false, 
-		codeHandler = ignoredCodes[warn.code];
-		
-		if(codeHandler){
-			suppressed = typeof codeHandler === "function" ? codeHandler(warn) : codeHandler;
+
+		var suppressed = false,
+			codeHandler = ignoredCodes[warn.code];
+
+		if (codeHandler) {
+			suppressed =
+				typeof codeHandler === "function" ? codeHandler(warn) : codeHandler;
 		}
-		
-		if(!suppressed){
+
+		if (!suppressed) {
 			defaultHandler(warn);
 		}
 	}
-	
-	_suppressCircularDependencyWarnings(warn){
-		const ignoredCircularFiles = [
-			"Column.js",
-			"Tabulator.js",
-		];
 
-		return ignoredCircularFiles.some(file => warn.importer.includes(file));
+	_suppressCircularDependencyWarnings(warn) {
+		const ignoredCircularFiles = ["Column.js", "Tabulator.js"];
+
+		return ignoredCircularFiles.some((file) => warn.importer.includes(file));
 	}
-	
-	bundle(){
-		if(this.env){
+
+	bundle() {
+		if (this.env) {
 			this.watch(this.env);
-		}else{
+		} else {
 			this.build();
 		}
-		
+
 		return this.bundles;
 	}
-	
-	watch(env){
+
+	watch(env) {
 		console.log("Building Dev Package Bundles: ", env);
-		switch(env){
+		switch (env) {
 			case "css":
-			this.bundleCSS(false);
-			break;
-			
+				this.bundleCSS(false);
+				break;
+
 			case "esm":
-			this.bundleESM(false);
-			break;
-			
+				this.bundleESM(false);
+				break;
+
 			case "umd":
-			this.bundleUMD(false);
-			break;
-			
+				this.bundleUMD(false);
+				break;
+
 			case "wrappers":
-			this.buildWrappers();
-			break;
-			
+				this.buildWrappers();
+				break;
+
 			default:
-			this.bundleCSS(false);
-			this.bundleESM(false);
-			break;
+				this.bundleCSS(false);
+				this.bundleESM(false);
+				break;
 		}
 	}
-	
-	build(){
+
+	build() {
 		console.log("Clearing Dist Files");
-		
+
 		this.clearDist();
-		
+
 		console.log("Building Wrappers");
-		
+
 		this.buildWrappers();
-		
+
 		console.log("Building Production Package Bundles");
-		
+
 		this.bundleCSS(false);
 		this.bundleCSS(true);
-		
+
 		this.bundleESM(false);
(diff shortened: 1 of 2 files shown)


dominikg:
thanks for working on improving esm compatibility :heart: 

Please keep in mind that adding an exports map is almost always a breaking change and you have to list all exports that are needed.

olifolkerd:
@dominikg do you want to clarify this further? Possibly with any examples of areas of specific concern there?

olifolkerd:
@thisjt could you revert the change to the bundler please and only include the e package change. Plenty of people still use the .js file directly when pulling from cdns so the .js file cannot be removed without disabling a huge number of users. Instead I will make the bundler duplicate the output to the new extension and deprecate the .js and remove it at the next major version change in a year or two to give people a chance to adapt

thisjt:
@olifolkerd done. You may review the changes and then take over from there.

dominikg:
I don't know the details behind the current structure of this package, just provided links to information how to make this esm compatible. Most ways to do this are a breaking change due to the way it changes how exports of your package can be reached. Without an exports map, everything that is exported can be imported by users. With an exports map, only the entries in the exports map can. publint and the guide by antfu go into more detail about this.

olifolkerd:
@thisjt is the exports bit actually needed? If you point the modules option to the .mjs file does it work in sveltkit?

thisjt:
@olifolkerd yes I assume it's required.

This repo (https://github.com/thisjt/tabulator-import-issue/tree/rename-import) imports (https://github.com/thisjt/tabulator-sveltekit2fix/tree/tabulator-import-rename). This throws an error. This is what its `package.json` looks like.
```
...
  "main": "dist/js/tabulator.js",
  "module": "dist/js/tabulator_esm.mjs",
  "sideEffects": [
    "**/*.css",
    "**/*.scss"
  ],
...
```

This repo (https://github.com/thisjt/tabulator-import-issue/tree/rename-import-with-exports) imports (https://github.com/thisjt/tabulator-sveltekit2fix/tree/tabulator-import-with-exports). This works fine and the table is being rendered correctly. This is what its `package.json` looks like.
```
...
  "main": "dist/js/tabulator.js",
  "module": "dist/js/tabulator_esm.mjs",
  "exports": {
    ".": {
      "require": "./dist/js/tabulator.js",
      "import": "./dist/js/tabulator_esm.mjs"
    }
  },
  "sideEffects": [
    "**/*.css",
    "**/*.scss"
  ],
...
```

olifolkerd:
Hey All,

This will be included in todays patch release.

Cheers

Oli :)

lazka:
fyi, this removed the export for the package.json file which made my build fail (I haven't looked in detail yet):

```
[!] Error: Package subpath './package.json' is not defined by "exports" in /home/runner/work/app/app/node_modules/tabulator-tables/package.json
    at new NodeError (node:internal/errors:405:5)
    at exportsNotFound (node:internal/modules/esm/resolve:366:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:713:9)
    at resolveExports (node:internal/modules/cjs/loader:590:36)
    at Function.Module._findPath (node:internal/modules/cjs/loader:664:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1126:27)
    at Function.resolve (node:internal/modules/helpers:188:19)
```

thisjt:
> fyi, this removed the export for the package.json file which made my build fail.
> 
> I used the path to package.json to figure out where the CSS is to copy that during bundling. I guess I can use the default import and go up the directory tree until I find the package.json instead. Other ideas welcome :)
> 
> ```
> [!] Error: Package subpath './package.json' is not defined by "exports" in /home/runner/work/app/app/node_modules/tabulator-tables/package.json
>     at new NodeError (node:internal/errors:405:5)
>     at exportsNotFound (node:internal/modules/esm/resolve:366:10)
>     at packageExportsResolve (node:internal/modules/esm/resolve:713:9)
>     at resolveExports (node:internal/modules/cjs/loader:590:36)
>     at Function.Module._findPath (node:internal/modules/cjs/loader:664:31)
>     at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1126:27)
>     at Function.resolve (node:internal/modules/helpers:188:19)
> ```

We might need to add this to `package.json` to fix the import css issue. I haven't really encountered this issue as I am importing the css file through `node_modules` (which is a bad practice I know, but I haven't bothered touching it since 2~ years ago)

```json
"exports": {
    ".": {
      "require": "./dist/js/tabulator.js",
      "import": "./dist/js/tabulator_esm.mjs"
    },
    "./dist/*.css": {
      "import": "./dist/*.css",
      "require": "./dist/*.css"
    }
}
```

https://github.com/vitejs/vite/discussions/2657

olifolkerd:
@thisjt Thanks for the headsup, ive just made an updated patch release on 6.2.3. let me know how that works for you.

Cheers

Oli :)

thisjt:
> fyi, this removed the export for the package.json file which made my build fail.
> 
> I used the path to package.json to figure out where the CSS is to copy that during bundling. I guess I can use the default import and go up the directory tree until I find the package.json instead. Other ideas welcome :)
> 
> ```
> [!] Error: Package subpath './package.json' is not defined by "exports" in /home/runner/work/app/app/node_modules/tabulator-tables/package.json
>     at new NodeError (node:internal/errors:405:5)
>     at exportsNotFound (node:internal/modules/esm/resolve:366:10)
>     at packageExportsResolve (node:internal/modules/esm/resolve:713:9)
>     at resolveExports (node:internal/modules/cjs/loader:590:36)
>     at Function.Module._findPath (node:internal/modules/cjs/loader:664:31)
>     at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1126:27)
>     at Function.resolve (node:internal/modules/helpers:188:19)
> ```

Can you confirm if this works on your end @lazka? Try editing `node_modules/tabulator-tables/package.json` and modify the exports to look like this

```json
  "exports": {
    ".": {
      "require": "./dist/js/tabulator.js",
      "import": "./dist/js/tabulator_esm.mjs"
    },
	"./dist/*.css": {
		"import": "./dist/css/*.css",
		"require": "./dist/css/*.css"
	}
  },
```

Then try importing the css in your scss stylesheet like this
```scss
@import 'tabulator-tables/dist/tabulator_site.css';
```

Oli has not published `6.2.3` yet, but this would be the expected edit for the export. Hopefully it will now work on your end.

---

@olifolkerd since `6.2.3` is not yet out, would it be possible to also add this to the exports? It would be nice to be able to import scss stylesheets directly from the package and take advantage of modifying the scss variables.

```json
  "exports": {
    ".": {
      "require": "./dist/js/tabulator.js",
      "import": "./dist/js/tabulator_esm.mjs"
    },
	"./dist/*.css": "./dist/css/*.css",
	"./src/scss/**/*.scss": "./src/scss/**/*.scss"
  }
```

So that the scss files can be imported~
```scss
// like this
@import 'tabulator-tables/src/scss/themes/tabulator_site.scss';

// instead of this
@import '../../node_modules/tabulator-tables/src/scss/themes/tabulator_site.scss';
```

The css and scss exports can be simplified as shown above as they are not affected on whether you're loading them in ESM, UMD, or CJS as per https://github.com/vitejs/vite/discussions/2657#discussioncomment-10050134

Or maybe I can raise a new PR for this.

lazka:
> Can you confirm if this works on your end @lazka?

The package.json export is still missing in your example (as in `"./package.json": "./package.json"`) , so I doubt it changes anything. And I'm not using SCSS.

Anyway, I solved it already via the workaround described above: https://github.com/digital-blueprint/toolkit/commit/47d250d945d11a5f9997cddfb5d764a49f36ea60

So feel free to ignore my use case :)

Thanks for asking!

dominikg:
as mentioned earlier, adding an exports map was a breaking change. not just for the reported package.json, but everything in src is now inacessible too. 

should have published it as 7.0

olifolkerd:
@dominikg could you explain what you are trying to access in src, none of those files are safe to include for production runs?

olifolkerd:
The same could be said if I change an internal variable name. But I don't expect people to do that and I don't update the major version for that either.

I would consider it a change in behaviour if it breaks expected standard documented behaviour which this doesn't, and if access to the src folder is important then I am more than happy to do a patch release with a wildcard export for that folder to make it accessible again making this a non breaking change :)

dominikg:
It is your package and you know your users better than I do. I'd caution against adding a glob in exports map as they bring back the problem you just described yourself. every rename or delete of a file matched by that glob would be a breaking change again. So the exports map limited to dist gives you a lot more freedom internally, you should keep it that way.

PR #4514, opened 2024-06-06
feat ResizeTable adds anti-shake redraw

Teadeno:


olifolkerd:
Hey @Teadeno 

Thanks for the PR,

Could you explain a bit about what the purpose of this PR is and why it is needed, things are a little light on details here beyond the title?

Cheers

Oli :)

PR #4516, opened 2024-06-11
fix navigation conflict with SelectRange and Edit

azmy60:
Using Up/Down Arrow keys while editing a cell can move the editor focus which is an uncommon behavior (comparing to Google Sheets and Ms. Excel). And it looks like we have 2 modules that can control table navigation => `SelectRange` module and `Edit` module. I think it's best if we can let **only** one module  to control navigation.

## Tabulator

tabulator-nav

Expected: Pressing arrow keys while editing should move the cursor inside the editor instead of continuing to navigate the table. 

## Google Sheets

gsheet-nav

### DIFF
--- a/src/js/modules/Edit/Edit.js
+++ b/src/js/modules/Edit/Edit.js
@@ -178,6 +178,10 @@ export default class Edit extends Module{
 		
 		if(cell){
 			
+			if(cell.column.modules.edit.navigationBlocked){
+				return false;
+			}
+
 			if(e){
 				e.preventDefault();
 			}
@@ -208,6 +212,10 @@ export default class Edit extends Module{
 		
 		if(cell){
 			
+			if(cell.column.modules.edit.navigationBlocked){
+				return false;
+			}
+		
 			if(e){
 				e.preventDefault();
 			}
@@ -238,6 +246,10 @@ export default class Edit extends Module{
 		
 		if(cell){
 			
+			if(cell.column.modules.edit.navigationBlocked){
+				return false;
+			}
+		
 			if(e){
 				e.preventDefault();
 			}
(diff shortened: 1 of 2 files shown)


olifolkerd:
Hey @azmy60 

Thanks for the PR. Got a few questions for you here.

First off, do you think you could manage this via confirm events, rather than direct module interaction, as when i merged in the spreadsheet module to start with, the idea is to decouple the modules rather than have them directly manipulate eachother. (for example other modules in future may also want to control this behaviour, so a general event is more useful.

Second, several of the editors control this behaviour directly. for example the `textarea` editor that allows the setting of the `verticalNavigation` property to control if arrow keys navigate or edit. How does this change interact with that? https://tabulator.info/docs/6.2/edit#editor-textarea

Cheers

Oli :)

olifolkerd:
Hey @azmy60 

Just wondering what your thoughts were on this one?

Cheers

Oli :)

azmy60:
Thanks for the feedback! I didn't know the `verticalNavigation` exists so I'll check that and also the confirm events sound very good! I'll try putting everything together next week.

azmy60:
@olifolkerd Sorry for getting back to this for a while. I tried using the confirm events and it ran an endless loop because both SelectRange and Edit module subscribe to "navigation-key-*" events which were dispatched by the keybinding module. Using `this.confirm` inside the listener caused an infinite loop. Do you have any suggestions? or I might miss something.

azmy60:
@olifolkerd I just updated the code to not use the module interaction directly. I don't use the `confirm` because it ends up in an infinite loop condition because both modules use keybinding events.

If you would prefer a similar approach to `confirm` events, we could add our version of `event.preventDefault` that's handled by the keybindings module. If this sounds better to you, I can give it a try!

azmy60:
@olifolkerd I managed to use the `confirm` like you suggested. Let me know what you think!

PR #4517, opened 2024-06-11
blur editor after pressing next/prev

azmy60:
There is a bug similar to this #4516. Pressing Tab/Shift+Tab would open the editor for the next cell which is uncommon if we compare it to other tools like Google Sheets and Ms. Excel.

Expected: Pressing tab should blur the editor and then go to the next cell without opening a new cell editor.

## Tabulator

tabulator-nav2

## Google Sheets

gsheets-nav2

### DIFF
--- a/src/js/modules/SelectRange/SelectRange.js
+++ b/src/js/modules/SelectRange/SelectRange.js
@@ -127,8 +127,8 @@ export default class SelectRange extends Module {
 		this.subscribe("edit-editor-clear", this.finishEditingCell.bind(this));
 		this.subscribe("edit-blur", this.restoreFocus.bind(this));
 		
-		this.subscribe("keybinding-nav-prev", this.keyNavigate.bind(this, "left"));
-		this.subscribe("keybinding-nav-next", this.keyNavigate.bind(this, "right"));
+		this.subscribe("keybinding-nav-prev", this.keyNavigate.bind(this, "prev"));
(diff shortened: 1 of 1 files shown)


olifolkerd:
Hey @azmy60 

Thanks for the PR.

On this one, while i accept that this behaviour deviates from excel, this is not a spreadsheet application. This is first and foremost a Table application that has recently had spreadsheet functionality added, which is very much the edge usage case as far as the main uses of the library are concerned.

Would you be ok moving this into an option for the module so that this different behaviour can be enabled by the user if desired?

Cheers

Oli :)

olifolkerd:
Hey @azmy60 

Just wondering what your thoughts were on this one?

Cheers

Oli :)

azmy60:
Hey @olifolkerd , sorry I missed looking into this one earlier, and of course I can make this as an option instead of the default behavior. Expect me to work on this next week :)

azmy60:
@olifolkerd I just added a new option to the Select Range module. Let me know what you think!

azmy60:
I think now we also need to avoid using edit module like this so I'm gonna work on that. :)

azmy60:
@olifolkerd So I just changed the code so the Select Range uses `this.chain` to check if Edit module has an editing cell. I also used this.dispatch("edit-cancel-cell") to cancel edit from Select Range. I'm not sure if this is the approach you'd like but please let me know if you have a better approach :)

Issue #4538, opened 2024-07-16
Facing issues with tabulator version 6.2.2.


sathamabb:
**Describe the bug**
Facing issues with tabulator version 6.2.2.

**Tabulator Info**
- 6.2.2

**Working Example**
![image](https://github.com/user-attachments/assets/836d2ce2-9bc5-474b-8ea4-d7744350a402)

I am using the tabulator latest version 6.2.2. After upgrading to this version, while starting the local server, I am getting the error like `./dist/css/tabulator_simple.min.css not exported from package`.

![image](https://github.com/user-attachments/assets/6cd49499-a6b5-4833-8b37-1c54da1a015d)
 
**To Reproduce**
1. Install the tabulator version 6.2.2.
2. Run the npm start command.
3. You can see the below error in the terminal.

**Screenshots**
![image](https://github.com/user-attachments/assets/836d2ce2-9bc5-474b-8ea4-d7744350a402)
![image](https://github.com/user-attachments/assets/94c6a321-d9cb-4a8c-95e1-288c18a8a55e)

**Expected behavior**
The local server should run and the tabulator should work as expected.

**Additional context**
Tabulator Version: 6.2.2

tadmi:
similarly

olifolkerd:
Have you tried 6.2.3, there was a regression on 6.2.2 due to the package.json change

tadmi:
problem remains on 6.2.3
![image](https://github.com/user-attachments/assets/8d30af60-2334-464f-b23c-ebd2fecf2b6f)

tadmi:
this works
![image](https://github.com/user-attachments/assets/897394b6-0c17-4103-b01c-80d33ad9d895)

olifolkerd:
@tadmi  Good spot, thanks for the info, ive just published 6.2.4, does that work for you?

Oli :)

tadmi:
@olifolkerd thanks, everything is working fine in the ESM project

sathamabb:
Hi @olifolkerd,

We also depend on Renderer, Row, Cell, and Module.

![image](https://github.com/user-attachments/assets/8e4c688f-02d2-406c-a9b5-4c09ea66674c)

![image](https://github.com/user-attachments/assets/3430b052-5835-42b3-951d-96d544e69fe5)

Could you please add `src/js` also in the export statement?

![image](https://github.com/user-attachments/assets/16bdafcd-5647-4f3d-bcba-59679952837d)

[...]

PR #4542, opened 2024-07-17
Add default export for JS/CSS dist files

jonatanschroeder:
As a followup to #4538: exports for CSS/JS files in dist have been set for import and require, but this does not include cases like bundling using esbuild, where the context is neither import nor require (it is browser/default). This causes an error when trying to import these files in the CSS or JS asset to be compiled/bundled.

This change ensures these files can be exported regardless of context.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,79 +1,73 @@
 {
-  "name": "tabulator-tables",
-  "version": "6.2.4",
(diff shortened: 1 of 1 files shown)


olifolkerd:
This has been included in the 6.2.5 patch release today. thanks for the PR :)

Issue #4380, opened 2024-02-01
Bootstrap 5 theme support


MarcJo:
In order tabulator works with bootstrap 5.3 themes you have to do following:

1. Change into sass variables tabulator.scss darken and lighten mixins
2. Change remaining hard coded colors into sass variables

I test this by manually change darken / lighten mixins with css filter:brightness and set sass variables like this

```scss
:root,
[data-bs-theme=light] {
	--tt-table-active-bg : #9ABCEA;
	--tt-table-active-bg-hover: color-mix(in srgb, var(--tt-table-active-bg), black 30%)
}

//Main Theme Variables
$backgroundColor: var(--bs-table-bg) !default; //background color of tabulator
$borderColor:var(--bs-table-border-color) !default; //border to tabulator
$textSize:16px !default; //table text size

//header theming
$headerBackgroundColor:var(--bs-table-bg) !default; //border to tabulator
$headerForegroundColor:inherit !default; //border to tabulator
$headerSeparatorColor:var(--bs-table-border-color) !default; //header bottom separator color
$headerTextColor:var(--bs-table-color) !default; //header text color
$headerBorderColor:var(--bs-table-border-color) !default;  //header border color
$headerMargin:4px !default; //padding round header

//
$cellPadding:12px !default; //padding round header

//column header arrows
$sortArrowHover: color-mix(in srgb, var(--bs-primary), var(--bs-gray-500) 40%) !default; //#555 !default;
$sortArrowActive: color-mix(in srgb, var(--bs-primary), black 0%) !default; //#666 !default;
$sortArrowInactive: color-mix(in srgb, var(--bs-gray-500), white 30%) !default; //#bbb !default;


//row theming
$rowBackgroundColor:var(--bs-table-bg) !default; //table row background color
$rowTextColor:var(--bs-table-color) !default; //table text color
$rowAltBackgroundColor: var(--bs-table-striped-bg) !default; //table row background color
$rowBorderColor:var(--bs-table-border-color) !default; //table border color
$rowHoverBackground:var(--bs-table-hover-bg) !default; //row background color on hover

//
$rowSelectedBackground: var(--tt-table-active-bg) !default; //row background color when selected
$rowSelectedBackgroundHover: var(--tt-table-active-bg-hover)  !default; //row background color when selected and hovered


$editBoxColor:var(--bs-primary) !default; //border color for edit boxes
$errorColor:var(--bs-danger) !default; //error indication


//footer theming
[...]

PR #4548, opened 2024-07-30
fix dark mode for Bootstrap 5

DastardlySky:
This issue has been raised before on here and on SO:
https://github.com/olifolkerd/tabulator/issues/4380

https://stackoverflow.com/questions/77524543/tabulator-js-5-apply-bootstrap-js-5-dark-theme

However, the issue still seems to persist when applying table-dark to bootstrap tables.

Currently, this is what a table looks like with the "table-dark" class:
https://jsfiddle.net/Lz8eacs7/

image

With the changes to the .scss files, when the compiled .css file is used it looks like this:
https://jsfiddle.net/tyvawr40

image


liepumartins:
Please merge this. 

As stated switching between light/dark boostrap5.3 themes is broken otherwise. Tabulator remains ugly light in a dark page.

BartNSTCL:
Looking to see when this will be merged into a release

DastardlySky:
@liepumartins, thanks for pointing that out, I've now amended it so that it will also work with Bootstrap 5's color modes e.g. the data-bs-theme attribute

DastardlySky:
Minimal reproducible example:
https://jsfiddle.net/2uwrjp4z/115/

data-bs-theme="light"

image

data-bs-theme="dark"

image

rathboma:
Thanks for the contribution! This looks fine for me. Thanks for the community review.

Issue #4540, opened 2024-07-16
TopCalc adds "space" between headers and rows


jaguarxii:
**Describe the bug**
Setting the topcalc option, add space (<br>) between column headers and rows.

**Tabulator Info**
- 6.2.3

**Working Example**
[CodePen sample](https://codepen.io/jaguarito/pen/yLjwQmy)

**To Reproduce**
A step by step guide to recreate the issue in your JS Fiddle or Codepen:
1. 
![image](https://github.com/user-attachments/assets/dd9c2c8e-3de1-4ed4-a490-9cb99b3fb796)

2. 
![image](https://github.com/user-attachments/assets/64ccda1b-c729-48f8-aaca-02c98b1bb4cb)

There is an "extra" `<br>` between headers and rows
![image](https://github.com/user-attachments/assets/1518c26a-8788-4b8d-84bb-6bf8ac3be104)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Waterfox (latest), Firefox (latest), Chrome (latest), Edge (latest)

**Smartphone (please complete the following information):**
 - Device: NOT TESTED

PR #4551, opened 2024-08-10
fix topCalc has blank line bug when group exists

lstgithub578:
issue #4540

A "br" was inserted in "initializeTopRow", but it was not deleted in "removeCalcs".
image


smb:
Is this going to merged? this bug prevents setting/removing grouping programmatically as the header grows on each toggle/change with the mentioned blank lines

PR #4601, opened 2024-10-17
Fix For #4600 - Error When Only Frozen Columns Are Present and render Horizontal is Enabled

### LINKED ISSUE #4600: Error When Only Frozen Columns Are Present and renderHorizontal is Enabled
**Describe the bug**
When only frozen columns are present in the DOM and the renderHorizontal option is enabled in Tabulator, an error is thrown during table rendering. The issue might be arises because the system attempts to access non-existent unfrozen columns, causing a crash.

**Tabulator Info**
- 6.3

**Working Example**
https://jsfiddle.net/uwaj5Lgt/

**To Reproduce**
1. You should have only frozen columns and `renderHorizontal` should be enable.
2. Go to provided jsfiddle link
3. Run
4. Check the console error.

**Expected behavior**
It shouldn't throw an error

**Screenshots**
image

 - OS: fedora, Linux
 - Browser [chrome]
 - Version [129.0.6668.100]

**Additional context**
I would be more than happy to do a PR for this Issue. Thanks.

kunaljaykam:
Fixes #4600

### DIFF
--- a/src/js/core/rendering/renderers/VirtualDomHorizontal.js
+++ b/src/js/core/rendering/renderers/VirtualDomHorizontal.js
@@ -190,9 +190,11 @@ export default class VirtualDomHorizontal extends Renderer{
 	}
 	
 	reinitializeColumnWidths(columns){
-		for(let i = this.leftCol; i <= this.rightCol; i++){
-			this.columns[i].reinitializeWidth();
-		}
+		columns.forEach(column => {
+			if(column.visible) {
+				column.reinitializeWidth();
+			}
+		});
 	}
 	
 	//////////////////////////////////////


olifolkerd:
Hey @kunaljaykam 

Thanks for the PR, unfortunatly this fix reslved the issues by looking through all columns and calling the function. The whole point of the horizontal renderer is that it only does calculation on the columns visible in the view port, other wise it becomes very inefficient with large numbers of columns. but your idea can be easily replicated without the need to look at the whole tables worth of columns, so i will apply the appropriate patch in the next patch release.

Cheers

Oli :)

Issue #4563, opened 2024-08-21
Left/Right arrow keys wont move text caret on inputs when editing a cell with selectableRange enabled


cpgo:
**Describe the bug**
When editing a cell and the table has `selectableRange` enabled left and right arrow keys will not move the input cursor

**Tabulator Info**
- Which version of Tabulator are you using?
- 6.2

**Working Example**
https://jsfiddle.net/4fc5s8oa/6/
Or the example on docs https://tabulator.info/docs/6.2/range#clipboard

**To Reproduce**
A step by step guide to recreate the issue in your JS Fiddle or Codepen:
1. Enable `selectableRange`
2. Open any cell editor via either `Enter` key or single/double clicking depending on `editTriggerEvent`
3. Try hitting `left` key to edit a single letter on the cell input

**Expected behavior**
I would expect the `input` to work like a normal input, meaning I can navigate with keyboard or click somewhere on the text to move the input caret

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Chrome
 - Version 127.0.6533.119

**Smartphone (please complete the following information):**

**Additional context**
I see that the text area input (https://tabulator.info/docs/6.2/edit#editor-textarea) has a `verticalNavigation` option that almost does what I expect, but as the name suggest only navigating up or down, I tried looking at the code to see if I could implement a `horizontalNavigation` option but had no luck

cpgo:
Found this workaround for now, not sure if this will break anything else
```
    Tabulator.extendModule("keybindings", "actions", {
      rangeExpandLeft: function(e) {
        if (this.table.modules.edit.currentCell) {
          return;
        }

        this.dispatch("keybinding-nav-range", e, "left", false, true);
      },
      rangeExpandRight: function(e) {
        if (this.table.modules.edit.currentCell) {
          return;
        }

        this.dispatch("keybinding-nav-range", e, "right", false, true);
      },
      navLeft: function(e) {
        if (this.table.modules.edit.currentCell) {
          return;
        }

        this.dispatch("keybinding-nav-left", e);
      },
      navRight: function(e) {
        if (this.table.modules.edit.currentCell) {
          return;
        }

        this.dispatch("keybinding-nav-right", e);
      },
    });
```

Still trying to keep the editor open when clicking inside the input but still havent found what module to extend to do that.

ValyaSHmelev:
Faced the same problem, thanks for the temporary solution

[...]

PR #4620, opened 2024-11-08
Fix left/right movement of caret while editing with selectRange enabled.

andy-5:
This is a fix for #4563.

I'm not sure why `keyNavigate` and `keyNavigateRange` always called `e.preventDefault()`, but using just the commented line inside the `if` block fixes caret movement while editing.

Or is there any reason to call `e.preventDefault()` even if range-navigation is not allowed?
I think the event handlers should not have any effect in this case, as implemented in this PR.

### DIFF
--- a/src/js/modules/SelectRange/SelectRange.js
+++ b/src/js/modules/SelectRange/SelectRange.js
@@ -529,9 +529,8 @@ export default class SelectRange extends Module {
 			}
 
 			this.layoutElement();
-			
-			return true;
 		}
+		return true;
 	}
 	
 	rangeRemoved(removed){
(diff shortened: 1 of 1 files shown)


azmy60 (changes requested):
It was commented because the page would scroll if you press arrow keys when you're no longer able to navigate at the side of the table. It was from this PR: https://github.com/olifolkerd/tabulator/pull/4474

I should've implemented a better solution, so `.preventDefault` inside the `if` statement is fine. But we also need to address the issue I mentioned. Probably `this.navigate` should return `true` even if the range is not moved.

andy-5:
I changed it so that `this.navigate` returns `true` even if the range is not moved, as you suggested. This seems to work great (tested also with the example linked in #4474).

azmy60 (approved):
Awesome! Thanks!

Issue #4633, opened 2024-11-17
no range will be created when selection starts at first column in first row


stefanhoelzl:
**Describe the bug**
When `selectableRange` is activated (e.g. set to `true`) it is not possible to create a range starting at the first column in the first row.

**Tabulator Info**
- Which version of Tabulator are you using? `6.3.2`

**Working Example**

[Screencast From 2024-11-17 17-02-40.webm](https://github.com/user-attachments/assets/b4635edc-22b7-4c91-889b-274021d1bfa3)

**To Reproduce**
A step by step guide to recreate the issue in your JS Fiddle or Codepen:
1. Go to https://jsfiddle.net/pqm372r6/
2. Click on `1` in column
3. `range added` will be appear in console output
4. Click on `0` in column
5. no console output `range added`

**Expected behavior**
The range event `rangeAdded` should fire when a selection start in the first cell

**Desktop (please complete the following information):**
 - OS: Fedora 41
 - Firefox
 - Version

LeonidShv:
I use "tabulator-tables": "^6.3.1" together with "vue": "^3.5.13", and it is still not possible to create a range starting from the first column in the first row.

My solution is to use the cellMouseUp event along with the getRanges() method:

```
  tabulator.on("cellMouseUp", (e, cell) => {
    selectedCells= tabulator.getRanges()[0].getCells().flat().map(item => item._cell)
  })
```

stefanhoelzl:
> I use "tabulator-tables": "^6.3.1" together with "vue": "^3.5.13", and it is still not possible to create a range starting from the first column in the first row.

there hasn't been a release since this fix was merged

PR #4658, opened 2024-12-23
fix range selection starting at top-left cell.

stefanhoelzl:
fixes https://github.com/olifolkerd/tabulator/issues/4633

`Range` gets initialized with cell 0,0, when this `Range` gets added and `setBounds` is called with cell 0,0 no update is triggered and therefore no events are dispatched.

The solution is to set the initial `Range` start and end position as `undefined`


azmy60 (approved):
I'm not sure `undefined` or `null` is the best way to represent initial values. But it should be good for now.

PR #4659, opened 2024-12-23
default range selection initialization can be disabled

### LINKED ISSUE #4634: default range cannot be removed
**Describe the bug**
When `selectableRange` is enabled (e.g. set to `true`) the first cell is selected by default, but it is not possible to remove the range and select no cell.

**Tabulator Info**
- Which version of Tabulator are you using? `6.3.2`

**To Reproduce**
1. Go to https://jsfiddle.net/qas4bwod/24/
2. Wait until table is loaded
3. value `0` is still selected

**Expected behavior**
I expect that there is nothing selected after I remove the range.

 - OS: Fedora 41
 - Browser Firefox
 - Version 132

**Additional context**
Even better I would expect that nothing is selected in the first place.

stefanhoelzl:
fixes https://github.com/olifolkerd/tabulator/issues/4634

by default cell [0,0] is selected, when range selection is disabled. With this change this can be disabled, so that on table init no range is selected.


azmy60:
Thanks for making a PR! Question - what happens if we press keyboard to navigate while the range does not exist?

stefanhoelzl:
> Thanks for making a PR! Question - what happens if we press keyboard to navigate while the range does not exist?

when the range does not exist, keyboard presses to navigate do nothing. You first need to select a range before you can navigate by keyboard.

src/js/modules/SelectRange/SelectRange.js
```diff
@@ -908,27 +910,29 @@ export default class SelectRange extends Module {
 		return range;
 	}
 	
-	resetRanges() {
+	resetRanges(forceNewRange = true) {
```
azmy60, line 913: I'm not sure If I like adding a `forceNewRange` flag like this. This could be a sign that we probably need to make a new function. Is it possible to remove this?

src/js/modules/SelectRange/SelectRange.js
```diff
@@ -908,27 +910,29 @@ export default class SelectRange extends Module {
...
+
+			if(this.table.rowManager.activeRows.length){
+				visibleCells = this.table.rowManager.activeRows[0].cells.filter((cell) => cell.column.visible);
+				cell = visibleCells[this.rowHeader ? 1 : 0];
+
+				if(cell){
+					range.setBounds(cell);
+					if(this.options("selectableRangeAutoFocus")){
+						this.initializeFocus(cell);
+					}
```
azmy60, line 930: I think this whole thing could be a new method? We can call it `setInitialRange()` or something similar.

src/js/modules/SelectRange/SelectRange.js
```diff
@@ -873,6 +874,7 @@ export default class SelectRange extends Module {
 	
 	
 	getActiveCell() {
+		if(!this.activeRange) return;
```
azmy60, line 877: Since this could be `undefined`, could you make sure that other places that use `getActiveCell()` handles `undefined` properly?

src/js/modules/SelectRange/SelectRange.js
```diff
@@ -758,7 +759,7 @@ export default class SelectRange extends Module {
 	redraw(force) {
 		if (force) {
 			this.selecting = 'cell';
-			this.resetRanges();
+			this.resetRanges(false);
```
azmy60, line 762: If you can make the new `setInitialRange` method, you don't need to pass this boolean flag.

azmy60 (changes requested):
It's a great idea! But there are some changes needed before merge. I would prefer not to pass boolean flags to function like this if we can avoid that.

src/js/modules/SelectRange/SelectRange.js
stefanhoelzl: It is used in only two places. One already checked the return value and to the other one I  added an check.

src/js/modules/SelectRange/SelectRange.js
stefanhoelzl: I refactored `resetRanges` into `resetRanges` and `setDefaultRange`

src/js/modules/SelectRange/SelectRange.js
stefanhoelzl: yes, the flag is gone no. (see also other comments)

azmy60 (approved):

PR #4672, opened 2025-01-23
Fix errors when a grouped table is shown

### LINKED ISSUE #4219: Unable to scroll to bottom on table with large amount of groups
**Describe the bug**
Strange scrolling behaviour on larger amount of groups (1000). Nearly unable to scroll to bottom by mousewheel, unexpected jumping of group rows.

**Tabulator Info**
- Tabulator V5.5.0

**Working Example**
https://jsfiddle.net/7oLv4ub9/3/

**To Reproduce**
Scroll down the table by mousewheel

**Expected behavior**
Should scroll down line by line

 - OS: Windows 10
 - Browser Firefox, Brave
 - Version FF114.0.1

**Additional Information
Seems to be, that the problem does not occur if startet with groupStartOpen:true, or with unlimited tableHeight

### LINKED ISSUE #4525: Grouped views do not allow for vertical scrolling & Group Expansion Outside Initial Viewport
Regarding ticket #4219, here is another attempt to report this bug to find a (better) solution. Unfortunately, there is no solution to this problem since 06.2023. Setting the renderVerticalBuffer to a high pixel value is only a conditional solution and will greatly affect the display speed for large tables.

**Describe the bug**

**Issue 1: Vertical Scrolling in Grouped Views**
In version 5.5.0, grouped views do not allow for vertical scrolling without manually setting a large renderVerticalBuffer.
It appears that the automatic calculation of the buffer is not functioning correctly for grouped views.

**Issue 2: Group Expansion Outside Initial Viewport**
When expanding groupings that are outside the initial visible section of the window, the grouping expands but the view jumps to the beginning of the table instead of keeping the expanded section in the visible area.

I expect the same behaviour as up to version 5.4.4. I think that a change/optimisation of the render is responsible for the behaviour. (https://tabulator.info/docs/5.5/release#performance)

**Tabulator Info**
Tabulator affected Version: >= 5.5.0
Previous Version (without issues): 5.4.4

Both problems are specific to version 5.5.0 and higher
These problems are also present in the latest version 6.2.1

**Working Example**
https://jsfiddle.net/syh7kjrm/9/

The problems do not occur in version 5.4.4.
https://jsfiddle.net/5o09y8nd/20/

**To Reproduce**
**Issue 1:**
1. scroll down

**Issue 2:**
1. scroll past the first screen 
2. click on a category to open it

**Expected behavior**
**Issue 1:**
I expect to be able to scroll to the end without any problems.

**Issue 2:**
I expect that when opening a category, the system does not jump to another area.

**Screenshots**

**Desktop:**
 - OS: Windows 10
 - Browser chrome
 - Version 125

**Additional context**
**Possible solution**
One possible solution is to set the "renderVerticalBuffer (https://tabulator.info/docs/6.2/layout#virtual-dom-buffer)".
But I think this is just a workarount and does not fix the error.

### LINKED ISSUE #4867: Scrolling is not possible when grouping
[index.html](https://github.com/user-attachments/files/26566761/index.html)

**Describe the bug**
Hello, I’ve got a strange problem: when I’ve hidden grouped elements and try to scroll, the scroll bar keeps jumping back to the top, so I can’t scroll any further down; and even when I open one, it jumps to a different spot. Have I done something wrong? I’ve uploaded a video and an index.html file with some sample code.

**Tabulator Info**
- 6.4.0

**Working Example**
[YOU MUST include a link to a JS Fiddle or Codepen that demonstrates the problem, it is very hard to diagnose an issue from a simple description.](https://jsfiddle.net/3btn7gs8/1/)

**To Reproduce**
1. Go to 'https://jsfiddle.net/3btn7gs8/1/'
2. Try scrolling down

**Expected behavior**
It must be possible to scroll down easily whether all groups are collapsed or expanded

**Screenshots**
[If applicable, add screenshots to help explain your problem.](https://github.com/user-attachments/assets/822ca6c9-fe11-43cd-b695-c14653486f31)

 - Edge Version 146.0.3856.109

 - None

**Additional context**

[index.html](https://github.com/user-attachments/files/26566764/index.html)

janhaertel:
If a table has **groups** which are all **collapsed** and the **VirtualDomVertical** renderer is used, there are some issues **scrolling** the table:

**Example**
https://jsfiddle.net/isaraystanz/tpgh5vze/11/

- On initial display there is a horizontal scrollbar even though **fitColumns** is enabled and the colums do fit the view without a scrollbar being needed.

I noticed that RowManager.renderTable() will call **VirtualDomVertical.renderRows() => _virtualRenderFill()** which triggers **GroupRows.virtualRenderFill()**.
In this function the minWidth is adjusted if there are only group rows causing a horizontal scrollbar to be displayed.
Later in RowManager.renderTable(), the call to RowManager.layoutRefresh() will lead to filtColumns.scaleColumns() to be invoked, which changes the width of the columns.
After this, the columns are thin enough to fit the view but the horizontal scrollbar is still being shown, due to the minWidth that was set.
I added a call to GroupRows.virtualRenderFill() after this, so the minWidth will be adjusted and the horizontal scrollbar disappears.

- The table cannot be scrolled to the bottom, because the view always jumps back to the top.

RowManager._virtualRenderFill() will call **row.calcHeight()** after rows have been rendered causing Group.outerHeight to be set internally. This is always 0 because the implementation for Group.calcHeight() is empty.
The rowsHeight calculated later will then also be 0 since all groups are collapsed.
When the view is scrolled so far, that the next rows have to be added at the bottom and some have to be removed from the top, 
**VirtualDomVertical._removeTopRow(**) tries to sum up row heights until none have to be removed anymore.
Since the height is always 0, all rows are added to a list of removable rows then **all** are **removed**.
I added a minimal implementation to Group.calcHeight(), so height will not be 0.

- The table scrolls to the top every time a group is collapsed/expanded.

Collapsing/Expanding a group will lead to_virtualRenderFill(), (GroupRows.updateGroupRows () ... => RowManager.reRenderInPosition() ...=> _virtualRenderFill()) which sets scrollTop after rendering. Here the position is not correct due to the groupRow height being empty.
This is fixed by the code from the previous issue.

Is the implementation for GroupRow.calcHeight() empty on purpose?
If so the height may have to be calculated differently.

These problems are also addressed in the github issues #4219 and #4525.

fixes #4219
fixes #4525


Kallius74:
This pull request fixes many of my scrolling issues with the vertical virtual renderer and row groups, but not all. Scrolling down with these code changes now works smoothly: if you hold down the thumb of the scrollbar and move it down, it follows the mouse pointer and maintains its vertical distance from the mouse pointer.
However, if you reach the bottom of the scroll area, hold down the scroll bar thumb and move it up again, you will notice jumps and an increasing distance between the scroll bar thumb and the mouse pointer.

The issue seems to arise from the line...
table.insertBefore(row.getElement(), table.firstChild);
... in VirtualDomVertical.js in _addTopRow:

Adding an element to the top changes the position of the vertical scrollbar, and this triggers a scroll event with an increased scroll position value compared to the last scroll action. 

The event listener attached in initialize() in RowManager.js recognizes this as a “scrolldown” action (even if the user has scrolled up).

This leads to a "_removeTopRow" and "_addBottomRow" call in VirtualDomVertical.js.

azmy60:
This is great! I'm trying to understand the whole situation carefully as well as looking into the @Kallius74 issue. Hopefully we can merge this soon!

src/js/core/RowManager.js
```diff
@@ -929,6 +929,11 @@ export default class RowManager extends CoreFeature{
 				}
 				
 				this.layoutRefresh(true);
+				
+				// handle resized columns
+				if(this.table.modExists("groupRows")) {
+					this.table.modules.groupRows.virtualRenderFill();
```
azmy60, line 935: Could you do this inside the [GroupRows](https://github.com/olifolkerd/tabulator/blob/d6b654c4385beefd32fdc808cc695a6577e4cd6e/src/js/modules/GroupRows/GroupRows.js) module? Just making sure that we are isolating group rows related codes within its own module.

You can probably subscribe to a `table-layout` event from the GroupRows module and see if that works. If not, please let me know, and we can figure something out.

azmy60 (changes requested):
Thanks for submitting the PR - awesome work! Just one small change request.

rathboma:
@janhaertel Just checking in to see if you can make the change Azmi requested above? If so, we can merge once the change is made

azmy60:
Hey all, I made a new PR based on @janhaertel's work for cleaner commits and changes (#4870). I'm taking over the issue and hopefully getting it merged this week.

PR #4695, opened 2025-02-17
remove tabulator-layout attribute when destroying table

Yay295:


Yay295:
Tabulator also leaves a `role` attribute on the element, but since that's a generic attribute it may have already been there from the start.

PR #4703, opened 2025-02-28
Add Basic E2E and Unit Tests Setup

azmy60:
As the project grows, keeping track of regressions is getting way harder. This PR adds basic setups for E2E tests and Unit tests for Tabulator. We can add more tests over time as we find bugs or add new features.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -12,7 +12,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions/setup-node@v4
       with:
-        node-version: lts/*
+        node-version: 20.15.1
     - name: Install dependencies
       run: npm ci
     - name: Install Playwright Browsers
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -8,11 +8,14 @@ jobs:
   test:
     timeout-minutes: 60
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [18, 20, 22]
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-node@v4
       with:
-        node-version: 20.15.1
+        node-version: ${{ matrix.node-version }}
     - name: Install dependencies
       run: npm ci
     - name: Install Playwright Browsers
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,29 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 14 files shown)


.gitignore
```diff
@@ -3,4 +3,12 @@
 
 node_modules/
 examples/
-npm-debug.log
\ No newline at end of file
+npm-debug.log
+
+dist/
```
rathboma, line 8: I don't think Oli usually excludes dist, because it's used to download a specific version for GitHub

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
+      with:
+        node-version: lts/*
```
rathboma, line 15: Do we want to be explicit about which node.js versions this runs on? What versions do we officially support with Tabulator -- I'd make this list the same.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,15 @@
+name: Unit Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
rathboma, line 9: What version of node js does this run on? I'd expect us to test on at least the same versions that we support,

Probably that's at least 18, 20, 22

test/e2e/SelectRange.spec.js
```diff
@@ -0,0 +1,54 @@
...
+	await page.keyboard.press('Enter')
+	await page.keyboard.press('ArrowLeft')
+	await page.keyboard.press('ArrowLeft')
+	await page.keyboard.type('-')
+	await page.keyboard.press('ArrowRight')
+	await page.keyboard.type('+')
+	await page.keyboard.press('Enter')
+	await page.keyboard.press('ArrowDown')
+
+	expect(page.getByText("a-b+c")).toBeVisible()
```
rathboma, line 53: Great stuff! Nice to see the first Playwright test :heart:

rathboma (changes requested):
Let's make sure we're supporting enough node.js versions

.gitignore
azmy60: Oh my bad. It was supposed to be for my local text editor.

.github/workflows/unit-tests.yml
rathboma: No, this should be a matrix, and should test node 20, 18, and 22

rathboma: @azmy60 see how we do it in queryleaf - https://github.com/beekeeper-studio/queryleaf/blob/main/.github/workflows/test.yml

azmy60: Done!

rathboma (approved):

PR #4710, opened 2025-03-07
Modern css setup and tailwindcss theme

tikiatua:
We recently started using tabulator for various projects and really love the library. 

This pull requests adds two features:
- Modern CSS styling with custom properties and nested CSS
- Tailwindcss theme with modern look, inspired by tailwindcssplus examples

What benefits does this change offer:
Styling tabulator for a specific project is currently somewhat cumbersome. With new CSS features, now in baseline support for modern browsers (https://web.dev/baseline) and the limited SASS-functionality actually used by tabulator, it is now possible to use native CSS without any preprocessor.

By replacing SASS variables with custom CSS properties, these variables can be very easily overridden in a specific project or tabulator instance, by just redefining the necessary variables in the respective scope. This makes adaptions for colors and font sizes easier. Details on how to use CSS custom properties can be found here: https://css-tricks.com/a-complete-guide-to-custom-properties/

Since many web developers are now using tailwindcss for styling, us included, we decided to also develop a tailwindcss theme, which is inspired by the table examples on https://tailwindcss.com/plus, with various enhancements.

A demo page is included in the theme folder and can be seen here:
https://codepen.io/saccilottor/pen/GgREMZb

Future enhancements:
In-depth styling of tabulator is still somewhat cumbersome due to the limited amount of variables defined and the often deep CSS nesting. This could be enhanced by exposing additional variables (i.e. text color for selected rows) and reducing the nesting for the CSS definitions.

Drawbacks of the pull request:
Older browsers, such as Internet Explorer, are not supported. This would lead to a duplication of work for CSS theming, if these should still be supported in the future.

### DIFF
--- a/src/css/tabulator.css
+++ b/src/css/tabulator.css
@@ -1,0 +1,1617 @@
+@import "./variables.css";
+
+.tabulator {
+	position: relative;
+
+	border: 1px solid var(--tabulator-borderColor);
+
+	background-color: var(--tabulator-backgroundColor);
+
+	font-size: var(--tabulator-textSize);
+	text-align: left;
+	overflow: hidden;
+
+	transform: translateZ(0);
+
+	&[tabulator-layout="fitDataFill"] {
+		.tabulator-tableholder {
+			.tabulator-table {
+				min-width: 100%;
+			}
+		}
+	}
+
+	&[tabulator-layout="fitDataTable"] {
+		display: inline-block;
+	}
+
+	&.tabulator-block-select {
+		user-select: none;
+	}
+
+	&.tabulator-ranges {
+		.tabulator-cell:not(.tabulator-editing) {
+			user-select: none;
+		}
+	}
+
+	/* column header containing element */
+	.tabulator-header {
+		position: relative;
+		box-sizing: border-box;
+
+		width: 100%;
+
+		border-bottom: 1px solid var(--tabulator-headerSeparatorColor);
+		background-color: var(--tabulator-headerBackgroundColor);
+		color: var(--tabulator-headerTextColor);
+		font-weight: bold;
+
+		white-space: nowrap;
+		overflow: hidden;
+
+		user-select: none;
+
+		outline: none;
+
+		&.tabulator-header-hidden {
+			display: none;
+		}
+
+		.tabulator-header-contents {
+			position: relative;
+			overflow: hidden;
+
+			.tabulator-headers {
+				display: inline-block;
+			}
+		}
+
+		/* individual column header element */
+		.tabulator-col {
+			display: inline-flex;
+			position: relative;
+			box-sizing: border-box;
+
+			flex-direction: column;
+			justify-content: flex-start;
+
+			border-right: 1px solid var(--tabulator-headerBorderColor);
+			background: var(--tabulator-headerBackgroundColor);
+
+			text-align: left;
+			vertical-align: bottom;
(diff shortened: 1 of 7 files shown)


rathboma:
Dang! This is a huge PR, thanks for all the work you've put in here. Tabulator only supports modern browsers officially, so I don't think the lack of support for IE is a blocker at all.

I'm less familiar with this part of Tabulator, so I'll probably need @azmy60 to review.

@tikiatua Quick question -- does this break any of the existing themes?

tikiatua:
@rathboma Thank you for your feedback. 

The short answer: 
It does not break any existing themes.

The longer explanation:
The existing themes use the scss variables defined in the file _src/scss/tabulator.scss_. All current themes build on this base definition. 

To not break the themes, I created a separate folder _src/css_. The file _src/css/tabulator.css_ corresponds to the _src/scss/tabulator.scss_ file and the tailwind theme would be a first example of how a theme can be specified using the new css setup. The DOM structure of the tabulator elements is exactly the same as before.

After merging the pull request, we could gradually transition the existing themes from scss to modern css and then finally drop the scss dependency. I am happy to help with this.

It might also make sense to expand the css variables provided in the tabulator base css, to allow for more fine grained adaptation of the theme, without having to overwrite existing css classes.

tikiatua:
@olifolkerd Ah, yes. I see the problem. I think it should be possible to actually define the scss variables on top of the css custom properties. 

```scss
// file src/scss/tabulator.scss

// import the custom css properties
@import "../css/tabulator.css";

// define scss variables based on custom css properties
$backgroundColor: var(--tabulator-backgroundColor);
$borderColor: var(--tabulator-borderColor);
```

This way tabulator could be setup with modern css while still supporting the scss approach. Would you think this is a feasible approach? I am happy to try this out and see if this would work.

Jordan-Hall:
> @olifolkerd Ah, yes. I see the problem. I think it should be possible to actually define the css custom properties using scss variables.


You are correct. This is how i use scss for a long time.  I think its worth making this change so nothing seriously broken. You example code is correct

src/css/tabulator.css
```diff
@@ -0,0 +1,1617 @@
...
+						white-space: normal;
+						text-overflow: initial;
+					}
+
+					/* element to hold title editor */
+					.tabulator-title-editor {
+						box-sizing: border-box;
+						width: 100%;
+
+						border: 1px solid #999;
```
Jordan-Hall, line 144: Shouldnt the border come from css var too? Ideally here you do not want to hard code theme & typeography settings

Jordan-Hall (changes requested):
There are a few instances in Tabulator where typography and theme settings are hard-coded. These should be moved to variables. To maintain backwards compatibility, use the second parameter in `var(--tabulator-x, defaultValue)` and set the hard-coded values as defaults.

yardensachs:
im excited for this one

Giveback007:
What is the overal status of this? I would love to use the library for an upcoming project, and tailwind css is how we are building all our UIs.

AlePiccin:
Any updates?

azmy60:
@tikiatua Yes, defining css props in scss would work. I believe this is how it looks like:

```scss
@use "../scss/tabulator.scss" as *;
:root {
  --tabulator-backgroundColor: #{$backgroundColor};
}
```

AslanAmca:
Any updates?

PR #4712, opened 2025-03-11
Fix popup positioning at the bottom edge of the screen

4wincode:
This should allow popups to start on the bottom edge even when there is no parent element.


rathboma:
@4wincode this looks good to me. Quick favor -- could you add a before/after screenshot or something like that, just so it's clear what this does?

Also -- is this solving an issue from the issue tracker? If so we should link it

Issue #4716, opened 2025-03-14
Editors do not activate if they are in two adjacent cells


miderose:
**Describe the bug**
Case: you some adjacent cells, each of them with custom editor.
You select one of them then move to the next cell with TAB key: the new selected cell won't go in edit mode.
In general a cell doesn't go in edit mode if the previous one was in edit mode.

**Tabulator Info**
6.3.0

**Working Example**
https://jsfiddle.net/0znwogys/3/
**To Reproduce**

1. Give the focus to a cell in the table
2. press TAB key to move to the next cell
3. note how the editor do activate only on "odd" cells.

**Expected behavior**
each cell with an editor should activate it when focused.

**Screenshots**
-

**Desktop:**
 - OS: Windows 10
 - Browser: faced on Chrome, Edge...
 - Version: Chrome 134.0.6998.89

**Smartphone:**
-  

**Additional context**
`src\js\modules\Edit\Edit.js:630`
https://github.com/olifolkerd/tabulator/blob/a3a2a2db277f6938d77a23a39d8c9cd01c38702e/src/js/modules/Edit/Edit.js#L631
maybe this could fix the issue:
```
if(this.currentCell){  
  if(this.invalidEdit && this.currentCell !== cell){  
    this.cancelEdit();  
    return;  
  }  
}
```

PR #4717, opened 2025-03-16
Fix(Edit module): fix for #4716

miderose:
This code fixes the case of 2 (or more) adjacent cells with an editor set (see issue #4716): when navigating with the TAB key the second (actually all the "even") editor wasn't activated


azmy60:
@miderose Good catch! Would it be possible if you could add a unit/e2e test for this? If you update your branch to master, you'll see there is a `test` folder in the root project.

rathboma:
Writing some tests for this, will create a new branch

PR #4724, opened 2025-03-25
perf: streamline cell content generation

lukecotter:
### What
Two changes to `Cell._generateContents`:
- Write `innerHTML` once instead of twice — the fallback set `innerHTML = value` and returned it, then the `switch` `default` wrote it again.
- Clear cell content with a `removeChild` loop (`_clearCellContent()`) instead of `innerHTML = ""`.

### Behaviour
Unchanged — Format tests pass. The common formatted-string path (`default: innerHTML = val`) is untouched. No new browser requirements.

### Performance (isolated, 250k renders; jsdom, directional)
- render, no formatter (2× → 1× innerHTML) 2971 → 1492 ms (~2.0×); clear cell (`innerHTML=""` → `removeChild`) 1257 → 489 ms (~2.6×).

### Grid impact
Negligible — with the Format module present the common path is unchanged, so full-grid render is unaffected (headless Chromium, virtual renderer). The win applies to no-Format builds and cell-clear paths. Baseline `upstream/master` @ `9539446`.

### DIFF
--- a/src/js/core/cell/Cell.js
+++ b/src/js/core/cell/Cell.js
@@ -98,22 +98,18 @@ export default class Cell extends CoreFeature{
 
 	//generate cell contents
 	_generateContents(){
-		var val;
-
-		val = this.chain("cell-format", this, null, () => {
-			return this.element.innerHTML = this.value;
+		const val = this.chain("cell-format", this, null, () => {
+			return this.value;
 		});
 
 		switch(typeof val){
 			case "object":
 				if(val instanceof Node){
-
 					//clear previous cell contents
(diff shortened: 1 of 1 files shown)


rathboma:
Hey! Was the previous implementation causing some issues for you?

lukecotter:
@rathboma 
Not a huge issue but when profiling I noticed quite a bit of time spent in parseHtml.

This speed up rendering for me especially when the virtual renderer is not being used. 

If you want a can provide some timings.

lukecotter:

PR #4734, opened 2025-04-11
perf: track deepClone circular refs with a WeakMap

lukecotter:
### What
Track circular references in `Helpers.deepClone` with a `WeakMap` (O(1)) instead of an array scanned per object (O(n)). Record each source object → its clone at entry, and look it up before recursing.

### Behaviour
Unchanged for normal clones. Also fixes two circular-reference bugs: the array version never recorded the root (a ref back to the root was cloned as a duplicate), and an earlier draft keyed the map by the parent but looked up the child. New `Helpers.spec.js` tests cover shared + nested + root circular references. Full suite green.

### Performance (isolated, Node, `deepClone` of a wide graph)
| nodes | before | after | speedup |
| ---: | ---: | ---: | ---: |
| 5,000 | 73.6 ms | 1.9 ms | ~38× |
| 20,000 | 1396 ms | 9.5 ms | ~147× |
| circular chain 3,000 | 7.9 ms | 0.8 ms | ~10× |

O(n²) → O(n).

### Grid impact
Negligible — `deepClone` runs off the render path (Accessor/Localize), and grid rows are small/non-circular (accessor `getData` over 20k rows moved ~7%). The large win is for big/deep/circular data structures. Baseline `upstream/master` @ `9539446`.

### DIFF
--- a/src/js/core/tools/Helpers.js
+++ b/src/js/core/tools/Helpers.js
@@ -32,7 +32,7 @@ export default class Helpers{
 		return output;
 	}
 
-	static deepClone(obj, clone, list = []){
+	static deepClone(obj, clone, circularRefs = new WeakMap()){
 		var objectProto = {}.__proto__,
 		arrayProto = [].__proto__;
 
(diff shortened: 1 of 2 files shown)


rathboma:
@lukecotter This does change the behavior though right? It changes it from using `.copy` to using the reference directly. Is that intentional? What are the side effects of that?

lukecotter:
@rathboma oof I forgot to add to the map. It’s in my local copy I will sort asap. Sorry about that.

lukecotter:

PR #4750, opened 2025-05-08
fix: grid not rendering on scroll drag

lukecotter:
fix: ensure height is not undefined

With a `variableHeight: true` on a column and a fixed height grid, if I dragged the scroll halfway down all rows would vanish.

Setting `dataTree: true` is a workaround for the issue.

This is because `rowsToRender` would end up being NaN after initial render due
to `row.getHeight()` being undefined and the sum for `rowsHeight` resulting in NaN, which caused `avgRowHeight` to also be NaN.

With `rowsToRender` and `avgRowHeight` nothing rendered.

Sometimes heightInitialized would be true because setCellHeight had be called.
The row outerHeight was undefined still, we now check offsetHeight and vDomRowHeight as fallback.


azmy60:
Hey @lukecotter ! Could you give us a little example of the code? This is my [code](https://jsfiddle.net/eL6v8xp1/1) but I can't replicate the issue you mentioned.

lukecotter:
@azmy60 
Sorry slightly got the cause wrong. 
Set `variableHeight:true` on a column to show the issue.
Here is a [fiddle](https://jsfiddle.net/nmbdhk2g/)

lukecotter:
Superseded by #3654, which fixes the same root cause (an undefined `row.getHeight()` producing NaN padding and collapsing the scroll height on drag with `variableHeight`). #3654 re-measures any row whose cached height is invalid/zero and falls back to `vDomRowHeight`, covering this PR's change. Closing.

Issue #4751, opened 2025-05-11
After the table was removed, the newly added rows were not immediately rendered


LTXHNN:
**Describe the bug**
After removing the Table from the page, I added rows to the table and then re added it to the page. However, I found that the newly added rows were not rendered. At that time, when you dragged it to the resize handle, you found that the newly added lines were rendered.

**Tabulator Info**
- Which version of Tabulator are you using?
6.3
**Working Example**
YOU MUST include a link to a JS Fiddle or Codepen that demonstrates the problem, it is very hard to diagnose an issue from a simple description.

[demo](https://jsfiddle.net/LTXHNN/L47s10gc/4/)

**To Reproduce**
1.Just click the refresh button and drag the resize handle to find this issue.

**Expected behavior**
I want the table to be removed from the page and displayed normally after being added to the page.

**Screenshots**

**Desktop (please complete the following information):**
 - OS:windows
 - Browser:edge

PR #4756, opened 2025-05-16
Fix: use latest IntersectionObserver entry to determine visibility (#4751)

mumberrymountain:
- Fixes a bug by using the latest IntersectionObserver entry to determine element visibility


azmy60:
hey @mumberrymountain , thanks for the PR!

Not sure what we are trying to fix here. Could you please add more details like steps to replicate the issue this PR is fixing? Probably a video to go through this would be very helpful as well! :)

mumberrymountain:
@azmy60  I can add tests if you like me do so. Is there any convention I suppose to follow? or do I just have to simply add new test file to test/unit folder?

azmy60:
Since this fixes the ResizeTable module, the new test should be in `test/unit/modules/ResizeTable.spec.js`.

azmy60:
Thanks!

Issue #4442, opened 2024-03-20
On Multiple row selection, deselction is not possible by clicking on the same row again


VaishnaviAdai:
**DESCRIPTION**
When you use keyboard shortcuts to select multiple rows by holding down the CTRL or SHIFT keys, you cannot deselect those rows using the checkboxes that usually allow for row selection.

We are using tabulator multi select row with checkbox. 
Click on the first row on the checkbox. Click on the third row on the checkbox with SHIFT Key. It works as expected. But when we click on the first row on the checkbox and again on the checkbox. It didn't deselect the row

**STPES TO REPRODUCE**

- Select a first row on checkbox. It selects the row.
- Click on the third row on checkbox by pressing SHIFT Key. It selects three rows.
- Click on the first row on checkbox. It selects the first row.
- Click on first row on checkbox. **It should deselect the first row. But it is not working.**

**Working Example**
https://jsfiddle.net/0djfhc2n/4/

**Expected behavior**
Deselection should happen on clicking the same row/checbox again

**Screenshots**
![image](https://github.com/olifolkerd/tabulator/assets/87337245/192138cc-0b81-42d6-be94-63e6989fec9e)

**Desktop (please complete the following information):**
 - OS: MAC OS/Windows
 - Browser : Chrome

VaishnaviAdai:
Hi @olifolkerd, Could you please confirm if this the feature requested from tabulator end

PR #4762, opened 2025-06-24
Fix: Refine row click behavior for selection and deselection

edmundophie:
This PR is to address an issue reported in https://github.com/olifolkerd/tabulator/issues/4442

### Current State
1. Select any row
2. Selecting the same row **doesn't deselect it**

### Expected State
1. Select any row
2. Selecting the same row again **will deselect it**

### Description
This PR tries to achieve the following behavior when a row is clicked without SHIFT/CTRL:
- If the clicked row is the only selected row: Deselect it.
- If the clicked row is one of multiple selected rows: Make it the only selected row (deselect others).
- If the clicked row is not currently selected: Deselect all others and select the clicked row.

This commit modifies the `handleComplexRowClick` function in the `SelectRow` module to correctly handle this scenario. Now, clicking an already selected row will deselect it, only if only it was the only one being selected previously.

### DIFF
--- a/src/js/modules/SelectRow/SelectRow.js
+++ b/src/js/modules/SelectRow/SelectRow.js
@@ -202,9 +202,8 @@ export default class SelectRow extends Module{
 		else if(e.ctrlKey || e.metaKey){
 			this.toggleRow(row);
 			this.lastClickedRow = row;
-		} else{
+		}else{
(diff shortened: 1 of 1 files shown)


azmy60 (approved):
There's no way to deselect a row, so I agree with this. Thanks for contributing!

Issue #4757, opened 2025-05-20
Missing event.touches when resizableColumnGuide is set to true and Chrome Dev-Tools are used


schlundus:
Describe the bug
When resizableColumnGuide is set to true, then the following error appears when resizing a column while
the chrome dev-Tools are open and the device is set to "Ipad Air" (happens also with Surface duo) ... no idea if this also happens with a real device..

![Image](https://github.com/user-attachments/assets/de3632a9-e3b6-4e02-9ce3-3adf631a9e8b)

Tabulator Info

Which version of Tabulator are you using?
6.2.0
Chrome 123.0.6312.123

Example can be found in documentation (just resize columns in with dev Mode enabled)....
https://tabulator.info/examples/6.3#resize-guides

PR #4769, opened 2025-07-06
Fix: script error during column resizing on mobile with resizableColumnGuide option enabled (#4757)

mumberrymountain:
This PR fixes issue #4757. Cannot generate issue on jsfiddle… check sample below instead.

[issue4757.zip](https://github.com/user-attachments/files/21088592/issue4757.zip)

Whereas default mobile column resizing behavior of Tabulator is to trigger resizing event during `touchmove` event, the `resizableColumnGuide` option changes this behavior so that the event is triggered on `touchend` event instead.

Generally, `touchend` event occurs when all fingers are lifted from the display, which means that `event.touches.length` will be 0 and `event.touches[0]` will be undefined, and accessing `event.touches[0].clientX` will cause a script error.

So, I fixed to use `e.changedTouches[0].clientX` instead when the `resizableColumnGuide` option is enabled, to avoid script error.

### DIFF
--- a/src/js/modules/ResizeColumns/ResizeColumns.js
+++ b/src/js/modules/ResizeColumns/ResizeColumns.js
@@ -210,9 +210,20 @@ export default class ResizeColumns extends Module{
 			component.modules.resize.handleEl.style.height = height;
 		}
 	}
+
+	getResizingClientX(e){
+		let x;
+		if (typeof e.clientX !== "undefined") return x = e.clientX;
+
+		const touch = this.table.options.resizableColumnGuide
+			? e.changedTouches?.[0]
+			: e.touches?.[0];
+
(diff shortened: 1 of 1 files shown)


azmy60 (changes requested):
This is great. But what happens if `getResizingClientX` returns `undefined`? maybe it's unlikely? Regardless, I think we should handle the `undefined` edge case too just to be safe. Maybe log a warning or something?

mumberrymountain:
@azmy60 

Well, think it's unlikely, but no harm in being cautious. I just added a guard for the undefined case.

azmy60:
Thanks @mumberrymountain !

azmy60 (approved):

Issue #3230, opened 2021-03-04
Please trigger events on the dom element additionally to callbacks


BernhardBaumrock:
Hi Oli, thx for the awesome library. I've been working with it now for quite some time and there's only one thing left that really causes me headache: Events and callbacks.

## The problem

I'm developing a module for the CMS https://processwire.com/ - my module will help closing the link between the data (cms) and the presentation (tabulator). It will have helpers for ajax calls etc... So far, so good. The problems arise when it comes to multiple developers providing multiple plugins for extendes features.

In case of formatters there is already the concept of extending modules: http://tabulator.info/docs/4.9/modules#extend; That is great, but only for modules and not for events!

Imagine me having a standard grid setup that ships with my module: A grid, that has headerfilter enabled by default and uses a predefined ajax endpoint so that I can manage data via my CMS on the PHP side.

Now imagine developer BOB wanting to add a great feature: When someone clicks on a row, he wants to show an alert "Hi, this is Bob, you clicked row ...". Right now - if I'm not missing anything - it's the only way to define that in the initial table setup:

```js
var table = new Tabulator("#example-table", {
 	...
 	rowClick:function(e, row){
 		alert("Hi, this is Bob, you clicked row " + row.getData().id + " Clicked!!!!");
 	},
});
```

Sounds simple enough? Yes and no... Now imagine me as the dev of my module wanting to add some magic when a user clicks on a row on any of my "rockgrids"... where would I define this callback? Or imagine developer John wanting to also show an alert "Hi, this is John, you really clicked on row...". And Bob wants to ship his feature as a module in a separate folder and John wants this as well.

That's simply not possible.

## Proposed solution

It would be really easy if tabulator exposed events on the dom element. Then me, Bob and John could listen for those events and simply fire our alerts without modifying (damaging) the initial table setup! Bob could do:

```js
table.on('rowClick', function(e, row) {
  alert('Hi, this is Bob, you clicked row ...');
});
```

And John could simply do the same:

```js
table.on('rowClick', function(e, row) {
  alert('Hi, this is John, you really clicked row ...');
});
```

Another example could be some kind of info-screen:

```js
table.on('rowSelectionChanged', function ...);
table.on('dataFiltered', function ...);
```

Those plugins could be loaded depending on a setting on the CMS side (eg via checkboxes "show header filter, show filter info, show column sums") and split into several separate js files. Right now I'd have to place all this logic into the initial table setup and that's a no-go.

[...]

PR #4772, opened 2025-07-17
Add "smart" + "smarter" filter to support fuzzy search + AND/OR operators

BernhardBaumrock:
Hi @olifolkerd maybe you can remember me from https://github.com/olifolkerd/tabulator/issues/3230 - still using your great library and still thankful for the event system :)

This PR adds two new filter implementation that I have been using for years as default in all my projects. Maybe you find them helpful and want to merge them to the core.

Docs can be found here (I'm using them in my module RockGrid for the ProcessWire CMS): https://www.baumrock.com/en/processwire/modules/rockgrid/docs/filters/

### DIFF
--- a/src/js/modules/Filter/defaults/filters.js
+++ b/src/js/modules/Filter/defaults/filters.js
@@ -105,4 +105,111 @@ export default {
 			return false;
 		}
 	},
+
+	// Smart filter
+	// Supports ., !, <, >, <=, >=, = and falls back to like filter
+	"smart": function (filterVal, rowVal, rowData, filterParams) {
+		const search = filterVal.trim();
+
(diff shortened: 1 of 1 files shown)


azmy60 (approved):
Hey @BernhardBaumrock. Thank you for adding these to Tabulator! I think this looks awesome.

BernhardBaumrock:
@azmy60 great to hear that you like it! I'd probably not call them "smart" and "smarter" as of today. Maybe just offer one single filter called "smart" filter which actually works like the "smarter" filter in my PR ?

azmy60:
@BernhardBaumrock Yes, that'd be simpler which is great!

PR #4802, opened 2025-09-10
fix(SelectRange): Fix frozen column width calculation

xxxace:
- Adjust z-index of range selection element for proper display
- Fix calculation to consider all occupied frozen columns


src/js/modules/SelectRange/Range.js
```diff
@@ -201,13 +209,36 @@ export default class Range extends CoreFeature{
...
+					occupiedFrozenColumnsWidth += column.width;
+				}
+			});
+
+			if (occupiedFrozenColumnsWidth === 0) {
+				// - z-index 10: No frozen columns are occupied by the selection
+				this.rangeManager.overlay.style.zIndex = 10;
+			} else {
+				// - z-index 11: Selection occupies at least one frozen column (ensures visibility above other elements)
+				this.rangeManager.overlay.style.zIndex = 11;
```
azmy60, line 225: Could we remove this, please? Increasing the z-index from css should be enough. No need to conditionally assign a style here.

azmy60 (changes requested):

PR #4818, opened 2025-11-03
Replace node-sass by sass (Dart Sass) to support NodeJS >= 20

### LINKED ISSUE #4817: Tabulator does not build with NodeJS > 18
Tabulator relies on the now deprecated package [`node-sass`](https://www.npmjs.com/package/node-sass), which supports NodeJS until v18 (due to some binaries that need to be recompiled).

This has no effect on end-users, but it means that it's not possible to run `npm install` with NodeJS > 18.

`node-sass` should be replaced by [`sass`](https://www.npmjs.com/package/sass) using Dart Sass, the official implementation for JS, as per the deprecation message.

Tyoneb:
Fixes #4817

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ npm-debug.log
 /playwright-report/
 /blob-report/
 /playwright/.cache/
+
+# VS Code
+.vscode/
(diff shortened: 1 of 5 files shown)


azmy60:
@Tyoneb Unfortunately, I just noticed this is a duplicate of #4809. But thanks for the PR! I'm closing this now.

Issue #4555, opened 2024-08-16
Getting accessibility errors related to ARC toolkit


anand24590:
While running ARC toolkit tool I am getting some accessibility errors related to Tabulator. Which I can't customize and has to be taken care in library code. Please check the list of bugs below:

1. ARIA widget has an incorrect owned element
    Description: The combobox element is not expected inside the grid role

<img width="1186" alt="Screenshot 2024-08-16 at 2 12 13 PM" src="https://github.com/user-attachments/assets/94b76816-57e5-470c-b71e-85c074c62eae">

2. ARIA widget missing required context role
    Description: The rowgroup role is missing the {{requiredContextRole}} required context role

<img width="1195" alt="Screenshot 2024-08-16 at 2 11 27 PM" src="https://github.com/user-attachments/assets/a76bef6a-45e7-4d08-aceb-0a9470d433f9">

3. Non-active element in tab order
    Description: The element has been placed in the Tab order using tabindex="0", but it lacks a specific role=" ..." attribute.

<img width="1223" alt="Screenshot 2024-08-16 at 2 13 15 PM" src="https://github.com/user-attachments/assets/9dc594c2-bbeb-4b0f-abb0-ef0de60bec67">

**Tabulator Info**
- Which version of Tabulator are you using?
- Version: 6.2.1

**To Reproduce**
1. Create table using tabulator with any basic setup

```
{
	data: this.data,
	columns: ["col1", "col2"],
	layout: 'fitColumns',
	textDirection: "ltr",
	reactiveData: true
}
```
5. Run [ARC Toolkit ](https://chromewebstore.google.com/detail/arc-toolkit/chdkkkccnlfncngelccgbgfmjebmkmce?hl=en)
6. and check for errors it received

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] iOS
 - Browser [e.g. chrome, safari] Chrome

**Additional context**
Please let me know if you need any code example. will share it

olifolkerd:
This will be fixed in todays release

PR #4836, opened 2026-01-14
Fix rowgroup roles on header/body (#4555)

andreanironi:
**Summary**

This pull request partially addresses the #2 issue reported on #4555 . 

**Details of Changes**

1. Removed the rowgroup role set on tabulator-header-contents, as it's redundant with the rowgroup role set on tabulator-header . And according to ARIA specifications, a rowgroup element cannot be a parent of another rowgroup
2. Added an element ID to the tabulator-table element, and connect it to the grid element via aria-owns. This allows to establish a parent-child relationship between the two elements in the accessibility tree while keeping the tabulator-tableholder element without a role


azmy60 (approved):
Thanks!

PR #4862, opened 2026-03-23
Fix multiselect list editor duplicate selections.

m3ftwz:
### Description

Fixes a weird behavior with multiselect in lists

#### Current

https://github.com/user-attachments/assets/8cbfadaa-0799-405c-a210-d5732ff9a802

#### Fixed

https://github.com/user-attachments/assets/a80b4547-388a-424f-9ee5-47d093136f22


src/js/modules/Edit/List.js
```diff
@@ -655,6 +655,10 @@ export default class Edit{
 			this.lastAction = "typing";
 		}
 		
+		if(this.params.multiselect) {
+			this.initialValues = null;
```
azmy60, line 663: @m3ftwz is this really meant to be null? could you explain a little bit, please?

m3ftwz: @azmy60 yes, initialValues is set to null after the first multiselect parse because it should only seed the initial selection once, while all later rebuilds must use currentItems as the real selected state source

azmy60:
Hey @m3ftwz , before this can be merged, would you be able to make a test for this? If it's possible, a simple test that would avoid regression is great.

m3ftwz:
Haven’t had much time lately, but this is on my mind, I'll test it more as soon as I can :)

m3ftwz:
@azmy60 ready, thank you for your patience

m3ftwz:
@azmy60 I solved conflicts, would you mind marking this as ready?

azmy60:
Hey @m3ftwz . Sorry for missing this. Will review this soon!

azmy60 (approved):
I don't see any regression so I'm merging this.

Note: I feel like the List.js has become super complex that the multi select functionality needs to be extracted to its own file.

PR #4864, opened 2026-03-25
Restore support basic, core Node-RED type for using node ui-tabulator.

teluse7:
Restore support basic, core Node-RED type for using node ui-tabulator.
After switching to luxon.js, support for Node-RED's basic, core Unix timestamp type was dropped. I've fixed file, returning the "x" type, similar to "iso."


azmy60 (approved):
LGTM. Thanks for contributing!

Issue #4869, opened 2026-04-13
https://tabulator.info/ website is down!


Anecho:
https://downforeveryoneorjustme.com/tabulator.info?proto=https

<img width="718" height="201" alt="Image" src="https://github.com/user-attachments/assets/5b41997a-a398-4039-82fb-62dbb0b597e9" />

Robinzon100:
I would be so good if the repo had a examples folder or an `examples.md` file that just has a dump of every example and it's code-block.
If the website comes back online again I'll do PR hopefully for that (if it doesn't already exist and I'm being ignorant here `:D` )

olifolkerd:
Website is back up. Thanks for the heads-up

PR #4872, opened 2026-04-14
feat(docs): add offline examples reference

Robinzon100:
### Motivation

During the recent website outage (#4869), the examples page at tabulator.info was unreachable. With no offline reference available, the only options were digging through source code or relying on external tools, neither ideal when you need to move fast.

### What this adds

- `docs/all_examples.md` — a Markdown mirror of the website's examples page, organized by feature
- All deep-dive links still point to the original docs; this is intended as a quick-reference index, not a replacement
- Keeps everything self-contained inside the repo so the GitHub page can serve as a fallback during outages

### Notes

- Content parity with the website's examples page as of this PR, no additions or omissions
- If this direction is welcome, I'm happy to expand the docs further so the repo can serve as a standalone reference

---

*This is my first open source contribution 
feedback on structure or conventions is very welcome.*

### DIFF
--- a/docs/all_examples.md
+++ b/docs/all_examples.md
@@ -1,0 +1,3560 @@
+# Examples | Tabulator
+[Virtual DOM - Vertical](#virtual-dom) [Documentation](https://tabulator.info/docs/6.4/virtual-dom)
+---
+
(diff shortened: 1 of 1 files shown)


rathboma:
Hey! Agreed that examples should be in the repo, but my plan is to move the docs site into the repo at some point to make it easy to hack on.

PR #4876, opened 2026-04-19
Change for Chrome DevTools to stop raising issues for pagination fields

victorprosa:
Currently, Chrome DevTools raises issues for tabulator tables with page size selection due to a lack of relation between the select input and it's label:
image

According to MDN, the recommended aproach is to always link the form and label elements with a "for" property, mainly due to accessibility:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label

Even if it's debatable whether DevTools should be raising these issues, trying to keep the browser issues list empty and following the best practices for accessibility, I propose this PR as a low priority minor annoyance fix that doesn't affect functionality.


rathboma:
@azmy60 should this use `instanceId`?

PR #4879, opened 2026-04-29
fix: correct data tree child row aggregation in column calcs

### LINKED ISSUE #4705: "dataTreeChildColumnCalcs" doesn't work, the group or table total doesn't sum the children
**Describe the bug**
With a grouped tree structure, totaling table rows no longer works. I noticed that the "table.getCalcResults()" function also does not consider children.

**Tabulator Info**
- 6.3.1

**Working Example**
https://jsfiddle.net/verdonfly/sn5yvuL6/20/

**To Reproduce**
you can see the result directly on the screen, or press the "Print total" button and view the result in the console, you will see that this is wrong

**Expected behavior**
I expect the total of the table and groups to also add the children since the "dataTreeChildColumnCalcs" parameter is true

**Screenshots**

 - OS:windows
 - Browser chrome
 - Version 133.0.6943.142

lukecotter:
**Summary**
This fixes an issue where column calculations for data tree rows could include the parent row repeatedly instead of the child row data.

In the data tree calculation path, [rowsToData] iterates filtered child rows and should push each child row’s data into the calc input. The current logic pushes the parent [row] object instead, which causes incorrect summed values for expanded data tree branches.

- Fixes incorrect totals/sums for data tree child rows
- Affects column calculations when [dataTreeChildColumnCalcs] is enabled and a tree node is expanded

fixes #4705

### DIFF
--- a/src/js/modules/ColumnCalcs/ColumnCalcs.js
+++ b/src/js/modules/ColumnCalcs/ColumnCalcs.js
@@ -411,7 +411,7 @@ export default class ColumnCalcs extends Module{
 
 			if(hasDataTreeColumnCalcs && row.modules.dataTree?.open){
 				this.rowsToData(dataTree.getFilteredTreeChildren(row)).forEach(dataRow =>{
-					data.push(row);
+					data.push(dataRow);
 				});
 			}
 		});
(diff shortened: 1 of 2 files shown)


rathboma:
Thank you!

Issue #4142, opened 2023-03-07
Text area unclickable when tabulator height set to a percentage


thearchitect3344:
**Describe the bug**
 When adding a table with a height set to a percentage, the text area editor becomes unclickable. It appears that tabulator is throwing the error "ResizeObserver loop limit exceeded" 

**Tabulator Info**
5.4.4

**Working Example**
https://jsfiddle.net/kmtg6pf1/

**To Reproduce**
1. Go to https://jsfiddle.net/kmtg6pf1/
2. Click on the first row of the column "info"
3. Notice the column cannot be edited

**Expected behavior**
The column can be edited when clicked

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser chrome
 - Version 110

Jojoshua:
Hey @olifolkerd, this seems to be an issue for quite a few of us. If we wanted to contribute a PR for this, could you give some pointers of where to begin?

AVerdolini:
I confirm the bug. the problem can be temporarily resolved by setting a fixed table height. 
This problem appears when you have the table height in percentage or "auto".

Tabulator Info
6.2.0
Here's another example:
 https://jsfiddle.net/verdonfly/9753zbjp/
to simulate it just change the value of "height" and try to modify the cell.

Jojoshua:
@azmy60 I saw 6.4 dropped but tested out the original fiddle in this issue with 6.4 and still seeing the same behavior.

rathboma:
@Jojoshua I'm planning for a June bugfix release and I'm going to try and include a fix for this issue (if we can replicate it)

Jojoshua:
> [@Jojoshua](https://github.com/Jojoshua) I'm planning for a June bugfix release and I'm going to try and include a fix for this issue (if we can replicate it)

@rathboma That would be great, been waiting for this for 3 years, let me know if you can't reproduce

Jojoshua:
@rathboma @not-night-but I can't reopen this but wanted to follow-up. It seems as if the original reproduction (with the updated dist files) still happening. See https://jsfiddle.net/qj8ouvx2/3/ 

Clicking in the cell still produces a resize observer error

<img width="560" height="575" alt="Image" src="https://github.com/user-attachments/assets/7eee123c-2327-4e42-98c4-e7f978d72303" />

not-night-but:
[...]

PR #4888, opened 2026-06-04
Fix textarea editor unusable with percentage table height

rathboma:
## Summary
Fixes issue #4142 where textarea editors become unclickable and unusable when the Tabulator table height is set to a percentage value. The problem was caused by a ResizeObserver feedback loop that would tear down the editor immediately after it was opened.

## Changes
- **ResizeTable.js**: Refactored resize observer callbacks to skip table redraws while a cell is actively being edited
  - Extracted edit state check into a new `editActive()` method for clarity and reusability
  - Updated both `resizeObserver` and `containerObserver` callbacks to use the new method instead of the previous mobile-specific logic
  - This prevents the editor from being torn down by resize-triggered redraws

- **Test files**: Added regression test suite
  - `editor-percentage-height.html`: Reproduction case matching the original issue's jsfiddle configuration
  - `editor-percentage-height.spec.js`: Playwright e2e test verifying the textarea editor opens, remains visible, stays focused, and accepts user input when the table height is a percentage

## Implementation Details
The root cause was that when a textarea editor opens with a percentage-based table height, the editor's content can grow the page, which grows the table container, triggering the ResizeObserver. This would fire a table redraw that immediately tears down the editor, creating a loop that left the editor unusable.

The fix prevents resize-driven redraws while editing is active. The table layout is automatically refreshed once editing ends, so no visual updates are lost. The `editActive()` method provides a cleaner alternative to the previous mobile-specific condition check.

https://claude.ai/code/session_01UGb71kmGbina8fQregNpLK


src/js/modules/ResizeTable/ResizeTable.js
```diff
@@ -127,6 +127,16 @@ export default class ResizeTable extends Module{
...
 	}
 	
+	//redrawing the table while a cell is being edited tears down the active
+	//editor. With a percentage based table height the editor itself can grow
+	//the page, which grows the table, which fires the resize observers in a
+	//loop and leaves the editor unusable (see issue #4142). Skip resize driven
+	//redraws while an edit is in progress; the layout is refreshed once editing
+	//ends.
+	editActive(){
+		return !!(this.table.modules.edit && this.table.modules.edit.currentCell);
```
azmy60: Not sure why we skip `browserMobile` check?

rathboma: The `browserMobile` flag wasn't doing anything on its own here — it only gated *whether* the "don't redraw while a cell is being edited" guard got applied. The old condition

```js
!browserMobile || (browserMobile && (!modules.edit || (modules.edit && !modules.edit.currentCell)))
```

reduces to:
- **desktop** (`!browserMobile`) → always redraw, edit state ignored
- **mobile** → redraw only when *not* actively editing

`editActive()` is exactly that "is there an active edit" check, now applied on every platform. So mobile keeps its existing behaviour (skip redraws during an edit), and desktop now gets the same guard — which is what fixes #4142: there the percentage-height feedback loop is a *desktop* resize that fires a redraw and tears the editor down mid-edit, so it could never be used.

In other words, dropping the `browserMobile` branch is intentional: it was only ever a switch for applying this guard, and the bug shows desktop needs it too. Happy to wrap it back in an explicit `browserMobile` short-circuit if you'd prefer to keep desktop redrawing during edits, but that would reintroduce the issue.

---
_Generated by [Claude Code](https://claude.ai/code)_

azmy60:
The fix was working but I had to add my changes because there's a regression. If we start editing, and then resize the table horizontally, the column widths are the same instead of resized.

Also, the ResizeTable is coupled to Edit. I don't think it needs to check whether edit is active or not to begin with.

All of those are fixed in my commit, and I added a test for the regression too.

Please let me know if we are good with that so we can merge the PR! :)

test/e2e/resize-table.html
```diff
@@ -0,0 +1,55 @@
...
+		re-fit when the table is resized during an edit.
+
+		Fixed height -> virtual render mode -> the ResizeObserver code path.
+	-->
+	<div id="wrapper">
+		<div id="resize-table"></div>
+	</div>
+
+	<script>
+	document.addEventListener("DOMContentLoaded", function () {
```
azmy60, line 36: We don't need to listen to this event to initiate the Tabulator.

not-night-but (approved):

PR #4890, opened 2026-06-04
Fix vertical scroll jumping with variable height rows

rathboma:
## Summary
Fixes issue #3654 where the vertical scroll would jump/skip when scrolling up after fast scrolling to the bottom of a table with variable height rows.

## Root Cause
The `_virtualRenderFill` method was summing row heights directly into padding totals without validating them. When a row was first measured while detached from the DOM, it would cache an `undefined` outerHeight but still be flagged as `heightInitialized`. This caused:
1. The row to be skipped during height re-initialization
2. `getHeight()` to return `undefined`
3. `rowsHeight += undefined` to produce `NaN` padding
4. CSS padding to collapse to 0, reducing scroll height
5. Content to lurch on each scroll-up step

## Key Changes
- **Height validation**: Modified the condition to re-initialize rows whose cached height is invalid or zero, not just those never initialized: `if(!row.heightInitialized || !row.getHeight())`
- **Fallback height**: Added fallback to `vDomRowHeight` estimate when a row's height is invalid: `row.getHeight() || this.vDomRowHeight`

These changes align with the guard logic already used in `_addTopRow`, `_addBottomRow`, and `_remove*` helpers.

## Testing
Added comprehensive e2e test coverage (`scroll-jump.spec.js` and `scroll-jump.html`) that:
- Reproduces the exact issue scenario (fast scroll to bottom, engage horizontal scrollbar, scroll back up)
- Measures content jump by tracking row movement vs. scroll distance
- Includes a control test for gentle scrolling to verify the metric reads ~0 for correct behavior
- Validates that worst-case jump is within acceptable tolerance (≤20px for fast scroll, ≤5px for gentle scroll)

https://claude.ai/code/session_015fVi8dcCLc7Rq16Bs8Jmna


azmy60 (changes requested):
1. The issue's author linked a page that contains the bug. I tried the latest version of the doc ([6.4](https://tabulator.info/examples/6.4)) and it probably has been fixed.
2. The tests lack `groupBy` feature. Some comments reported that disabling the group feature fixes the issue.

azmy60 (approved):
Looks good! Just refactored it a little bit for readability.

pvh-archifact:
During testing of this issue, the table content stuck in an endless loop. I noticed it resets/sets the height of the .tabulator .tabulator-tableholder element. each time it reaches bottom of the table.

All looping problems go away if I do not allow it to reset the height by setting thus overruling height calculation:
for example:
.tabulator .tabulator-tableholder{
height: 400px **!important;**
}

Issue #4421, opened 2024-03-02
CellClick callback does not fire if editor is defined but cell is not editable


Deejerz88:
**Describe the bug**
With the 5.6+ update, it seems that having an editor & cellClick defined prevents the cellClick callback from triggering if the cell is not editable. This is the case whether the properties are set in the columnDefaults or on the columns themselves. Previous to 5.6, this worked fine. 

My use case is that I want cells to be editable when the cell is selected, otherwise it would copy the cell contents to the clipboard. In these fiddles, I just fire a console.log on cellClick.

**Tabulator Info**
- Tabulator 5.6

**Working Example**

**Tabulator 5.6 Example**

**To Reproduce**
1. Go to https://jsfiddle.net/djmcmillan/zbr4ghL8/14/
2. Click on a cell in the Name column
3. Check console - no console.log was fired
4. Click the checkbox in the same row to select the row
5. Click the name cell in the same row.
6. The cell editor activates and console.log is fired
7. Deselect row
8. Commnet out or delete ```editor: 'input'``` (ln. 28)
9. Click "Run"
10. Click a cell in the Name column
11. Console.log is fired

**Expected behavior**
The cellClick callback should trigger if the cell editor is defined but the cell is not editable. This was the case in previous versions.

**Tabulator 5.5.1 Example**

**To Reproduce**
1. Go to https://jsfiddle.net/djmcmillan/0fmL3z7t/19/
2. Click on a cell in the Name column
3. Check console - console.log **was** fired
4. This remains to be true regardless of editor / editable configuration.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 122.0.6261.95 (Official Build) (64-bit)

feng-zh:
I see similar issue when using 5.6 or 6.0 code.

By looking into change from 5.6, the new feature "editTriggerEvent" introduce the change in "Edit" module. By default, the "editTriggerEvent" table option is "focus".

https://github.com/olifolkerd/tabulator/blob/6f4934420e4bb3ffd57769343df1a78e464a6aef/src/js/modules/Edit/Edit.js#L498-L513

When click in cell, the click event was first consumed in cell level event listener, Inside of edit call, the event is "stopPropagation".
https://github.com/olifolkerd/tabulator/blob/6f4934420e4bb3ffd57769343df1a78e464a6aef/src/js/modules/Edit/Edit.js#L656-L659

[...]

PR #4891, opened 2026-06-04
Fix click propagation for non-editable cells with editors

rathboma:
## Summary
Fixed an issue where click events on non-editable cells were being stopped from propagating, preventing the `cellClick` callback from firing when a column had an editor defined but the cell itself was not editable.

## Changes
- **Edit.js**: Moved the `e.stopPropagation()` call to only execute after confirming the cell is actually editable. Previously, the propagation was stopped before checking editability, which swallowed clicks on non-editable cells and prevented them from reaching parent listeners.

## Implementation Details
The fix ensures that:
- Click events on non-editable cells now properly bubble up to parent listeners, allowing the table-level `cellClick` callback to fire
- Click events on editable cells still have propagation stopped as intended
- The change is minimal and surgical, only reordering when `stopPropagation()` is called relative to the `allowEdit()` check

This resolves issue #4421 where the `cellClick` event was not firing for cells in columns with editors defined but marked as non-editable.

https://claude.ai/code/session_01Nt4ggeWkoLbhrLcYwciwts

### DIFF
--- a/test/unit/modules/Edit.spec.js
+++ b/test/unit/modules/Edit.spec.js
@@ -129,44 +129,44 @@ describe("Edit module", () => {
 
 	// Regression test for https://github.com/tabulator-tables/tabulator/issues/4421
 	//
-	// Tabulator wires the `cellClick` callback through a single click listener
+	// Tabulator dispatches the `cellClick` callback from a single click listener
 	// bound to the table element; clicks on cells bubble up to it. Since 5.6 the
 	// edit click handler calls `e.stopPropagation()` *before* checking whether
 	// the cell is actually editable. When a column defines an editor but the
 	// cell is not editable, the click is swallowed and never reaches the
 	// table-level listener, so the `cellClick` callback never fires.
-	//
-	// jsdom does not lay the table out, so rows are never attached to the live
-	// DOM and the real table-level listener can't receive bubbling clicks. To
-	// reproduce the defect we attach the cell element under an ancestor that
-	// stands in for that table-level listener and assert the click reaches it.
-	it("should not stop click propagation when editor is defined but cell is not editable (#4421)", async () => {
+	it("should fire cellClick when an editor is defined but the cell is not editable (#4421)", async () => {
+		const cellClick = jest.fn();
+
+		// A fixed height + basic vertical rendering forces the rows to be laid
+		// out into the live DOM under jsdom, so a real click bubbles up to the
+		// table-level listener that powers the cellClick callback.
 		table = await setupTable({
+			height: 300,
+			renderVertical: "basic",
 			columns: [
 				{ title: "ID", field: "id" },
 				// Editor defined, but the cell cannot be edited.
-				{ title: "Name", field: "name", editor: "input", editable: false },
+				{ title: "Name", field: "name", editor: "input", editable: false, cellClick },
 				{ title: "Age", field: "age", editor: "number" },
 			],
 		});
 
+		const externalCellClick = jest.fn();
+		table.on("cellClick", externalCellClick);
+
+		table.redraw(true);
+		await new Promise(resolve => setTimeout(resolve, 20));
+
 		const cellElement = table.getRows()[0].getCell("name").getElement();
-
-		// Stand-in for Tabulator's table-level interaction listener that powers
-		// the cellClick callback.
-		const ancestorClickHandler = jest.fn();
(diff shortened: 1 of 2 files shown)


rathboma:
## Root cause of #4421 and why this behavior change is correct

### Where the bug came from

This is an **accidental regression introduced by the `editTriggerEvent` feature in 5.6.0**, not a deliberate behavior change.

In **5.5.4** (working), `bindEditor` wired editing through two separate handlers — the click handler only focused the cell, and `edit()` was invoked from the **focus** handler:

```js
element.addEventListener("click", function(e){
    if(!element.classList.contains("tabulator-editing")){
        element.focus({preventScroll: true});   // click ONLY focuses
    }
});
element.addEventListener("focus", function(e){
    if(!self.recursionBlock){
        self.edit(cell, e, false);               // edit() gets the FOCUS event
    }
});
```

So `edit()` — and its long-standing `e.stopPropagation()` (which dates back to the 2021 modular rewrite) — only ever ran against the **focus** event. Focus events don't bubble, so that `stopPropagation()` was harmless: the actual **click** always bubbled up to the table-level interaction listener and `cellClick` fired.

In **5.6.0**, the `editTriggerEvent` option was added and the click handler was rewritten to call `edit()` directly with the **click** event:

```js
if(this.options("editTriggerEvent") === "focus" || this.options("editTriggerEvent") === "click"){
    element.addEventListener("click", function(e){
        if(!element.classList.contains("tabulator-editing")){
            element.focus({preventScroll: true});
            self.edit(cell, e, false);           // edit() now gets the CLICK event
        }
    });
}
```

Now the same old `e.stopPropagation()` runs against a **bubbling click** event, *before* the editability check:

```js
if(!cell.column.modules.edit.blocked){
    if(e){ e.stopPropagation(); }   // swallows the click...
    allowEdit = this.allowEdit(cell); // ...even when this is false
    ...
}
```

For a cell that has an editor defined but is **not** editable, the click is swallowed before it can reach the table-level listener, so `cellClick` (and any other click-driven handler) never fires. Nobody decided "non-editable editor cells should stop firing `cellClick`" — it fell out of passing a bubbling event where previously only a non-bubbling one was passed.

### Why changing the behavior is correct

- It **restores the pre-5.6 / 5.5.x behavior** that this issue reports as the regression, and makes a non-editable editor cell behave like any ordinary cell — consistent and least-surprising.
- The fix is minimal and surgical: it only gates `stopPropagation()` behind the `allowEdit || forceEdit` check, so propagation is stopped **exactly when the cell is actually being edited** (which is the only case that needs it, to avoid the click leaking to other handlers while an editor opens). The `editTriggerEvent` feature is untouched.
- **Editable cells are unchanged** — `stopPropagation()` still runs for them.
- **No double-firing**: the column-level `cellClick` is invoked directly inside `edit()` only on the editable branch; the non-editable case fires once, via normal bubbling.

### One thing worth flagging

Restoring propagation means it isn't *only* `cellClick` that fires for these cells — all click-driven consumers do, as they did pre-5.6: the external `cellClick` event, click-triggered menus (`clickMenu`) and popups (`clickPopup`), `SelectRange` click handling, and `rowClick`. This matches 5.5.x and is consistent with non-editor cells, but anyone who came to rely on the 5.6+ "editor cells swallow clicks" behavior would notice the difference. The alternative — dispatching `cellClick` explicitly for non-editable cells instead of restoring propagation — would diverge from pre-5.6 behavior and add complexity, so the current approach seems preferable.

Covered by a regression test in `test/unit/modules/Edit.spec.js`.

---
_Generated by [Claude Code](https://claude.ai/code)_

test/unit/modules/Edit.spec.js
```diff
@@ -115,17 +115,59 @@ describe("Edit module", () => {
...
+		});
+
+		const cellElement = table.getRows()[0].getCell("name").getElement();
+
+		// Stand-in for Tabulator's table-level interaction listener that powers
+		// the cellClick callback.
+		const ancestorClickHandler = jest.fn();
+		const ancestor = document.createElement("div");
+		document.body.appendChild(ancestor);
+		ancestor.appendChild(cellElement);
```
azmy60: I don't think this is correct. We should just listen to the cell click event from the table instance.

azmy60 (changes requested):

azmy60 (approved):

PR #4894, opened 2026-06-05
fix errors when handling BigInt

azmy60:
picked from this commit: https://github.com/beekeeper-studio/tabulator/commit/13f9a9b99a7fb7bb8359a6f991afbd874835e2fd


rathboma (changes requested):
This needs a replication test.

rathboma (approved):

PR #4896, opened 2026-06-05
Remove dist folder and create proper release process

not-night-but:
Remove the dist folder and add a workflow that builds and publishes to NPM and creates a draft release in Github. I'm not sure what we need to do with bower.


.github/workflows/publish.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Build
+        run: npm run build
+
+      - name: Release
+        uses: softprops/action-gh-release@v3
+        with:
+          draft: true
+          generate_release_notes: true
+
+      - name: Publish Package
```
rathboma, line 39: So it creates a draft release on GitHub, but a production release on npm? Probably it should just be production for both?

bin/make-release.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+
+set -euo pipefail
```
rathboma, line 3: Love our make release script

bin/make-release.sh
```diff
@@ -0,0 +1,109 @@
...
+# Step 5: Push tag
+if git tag | grep -q "$NEW_TAG\$"; then
+  echo "Tag $NEW_TAG already exists."
+else
+  echo "Creating and pushing tag $NEW_TAG..."
+  git tag "$NEW_TAG"
+  git push origin "$NEW_TAG"
+fi
+
+echo "Release process completed successfully."
```
rathboma, line 109: Yes murder the dist folder!

.gitignore
```diff
@@ -10,3 +10,5 @@ npm-debug.log
 /playwright-report/
 /blob-report/
 /playwright/.cache/
+
+dist/
```
rathboma, line 14: ❤️

package.json
```diff
@@ -19,6 +19,12 @@
     "**/*.css",
     "**/*.scss"
   ],
+  "files": [
+    "dist/",
```
rathboma, line 23: Is this right? dist and src listed? I think this should just be dist?

rathboma (approved):
Please merge this after looking at my questions and either dismissing them or dealing with them (up to you)

PR #4899, opened 2026-06-07
fix: eliminate O(n) lag on rectangular selection in large tables

KrzysztofDusko:
Rectangular selection via mouse drag caused severe lag proportional to scroll position. At the bottom of a 50k-row table, each mousemove event triggered three O(n) operations making selection unusable.

Root causes and fixes:

1. RowManager.getRowFromPosition() used .find() scanning from index 0, reaching O(n) for rows near the bottom. Replaced with O(1) direct array access displayRows[position - 1] with .find() fallback for grouped-row edge cases.

2. SelectRange.getTableRows() created a new filtered array copy on every call. Added _tableRowsCache with invalidation in resetRanges() and tableDestroyed().

3. handleCellMouseMove / handleColumnMouseMove had no throttling, causing the full layout pipeline (getActiveCell -> getCell x 2 per range -> range.layout) to execute on every mouse event (~120/sec). Wrapped in requestAnimationFrame to cap at one layout per frame.

### DIFF
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -194,8 +194,15 @@ export default class RowManager extends CoreFeature{
 	}
 	
 	getRowFromPosition(position){
-		return this.getDisplayRows().find((row) => {
-			return row.type === "row" && row.getPosition() === position && row.isDisplayed();
+		var displayRows = this.getDisplayRows();
+		var row = displayRows[position - 1];
+
+		if (row && row.type === "row" && row.getPosition() === position) {
+			return row;
+		}
+
+		return displayRows.find((row) => {
+			return row.type === "row" && row.getPosition() === position;
 		});
 	}
 	
--- a/src/js/modules/SelectRange/SelectRange.js
+++ b/src/js/modules/SelectRange/SelectRange.js
@@ -23,6 +23,10 @@ export default class SelectRange extends Module {
 		this.maxRanges = 0;
 		this.activeRange = false;
 		this.blockKeydown = false;
+		this._rafId = null;
+		this._pendingCell = null;
+		this._pendingColumn = null;
+		this._tableRowsCache = null;
 		
 		this.keyDownEvent = this._handleKeyDown.bind(this);
 		this.mouseUpEvent = this._handleMouseUp.bind(this);
@@ -232,6 +236,11 @@ export default class SelectRange extends Module {
 	_handleMouseUp(e){
 		this.mousedown = false;
 		document.removeEventListener("mouseup", this.mouseUpEvent);
+
+		if (this._rafId) {
+			cancelAnimationFrame(this._rafId);
+			this._rafId = null;
+		}
 	}
 	
 	_handleKeyDown(e) {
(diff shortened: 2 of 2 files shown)


rathboma:
Plan: Azmi reviews, then merges once good, or close rejected if not good.

We need:
1. regression tests
2. Tests confirming the fixes (when possible and not just performance)

src/js/core/RowManager.js
```diff
@@ -194,8 +194,15 @@ export default class RowManager extends CoreFeature{
 	}
 	
 	getRowFromPosition(position){
-		return this.getDisplayRows().find((row) => {
-			return row.type === "row" && row.getPosition() === position && row.isDisplayed();
+		var displayRows = this.getDisplayRows();
+		var row = displayRows[position - 1];
```
azmy60, line 198: Wouldn't it be better if we have an array, say `positionedRows`, which is cached somewhere? and then here, we can just do:

```ts
const row = this.positionedRows[position - 1];
if (row.isDisplayed()) {
  return row;
}
```

src/js/modules/SelectRange/SelectRange.js
```diff
@@ -374,8 +392,18 @@ export default class SelectRange extends Module {
 		if (!this.mousedown || this.selecting === "all") {
 			return;
 		}
-		
-		this.activeRange.setBounds(false, cell, true);
+
+		this._pendingCell = cell;
+
+		if (!this._rafId) {
+			this._rafId = requestAnimationFrame(() => {
```
azmy60, line 399: I don't think we need this.

src/js/modules/SelectRange/SelectRange.js
```diff
@@ -374,8 +392,18 @@ export default class SelectRange extends Module {
...
+				this._rafId = null;
+				if (this._pendingCell) {
```
azmy60, line 401: This is one step closer to a better approach.

I think we also want to check if the `_pendingCell` is not the current `cell`.

```ts
if (this._pendingCell !== cell) {
  this.activeRange.setBounds(false, cell, true);
  this._pendingCell = cell;
  // do not set _pendingCell to null here
}
```

And once we're done "selecting", we can set `_pendingCell` to null.

If you do this, I don't think we need `requestAnimationFrame`.

azmy60 (changes requested):

src/js/modules/SelectRange/SelectRange.js
```diff
@@ -340,11 +349,20 @@ export default class SelectRange extends Module {
 		if (column === this.rowHeader || !this.mousedown || this.selecting === 'all') {
 			return;
 		}
-		
-		this.activeRange.setBounds(false, column, true);
+
+		this._pendingColumn = column;
+
+		if (!this._rafId) {
+			this._rafId = requestAnimationFrame(() => {
```
azmy60, line 356: Please avoid using `requestAnimationFrame`, and use the same logic I mentioned in `handleCellMouseMove`.

azmy60 (changes requested):
I think this PR still needs some work. There's also a merge conflict that needs fixing. I'm closing this right now, but feel free to re-open any time if you're willing to continue the work!

PR #4904, opened 2026-06-15
perf(core): clone row template and slice active rows

lukecotter:
### What
- `Row.createElement`: clone a shared template (`cloneNode(false)`) instead of `createElement` + `classList.add` + `setAttribute` per row.
- `RowManager.setActiveRows`: copy with `slice()` instead of `Object.assign([], rows)` (drops a dead double self-assignment).

### Behaviour
Unchanged — full suite green.

### Performance
Baseline `upstream/master` @ `9539446` · branch @ `5f387975`.
- **Isolated** (Node, 250k rows): `setActiveRows` 23.2 → 0.14 ms (**166×**); `Row.createElement` 420 → 249 ms (1.7×).
- **Grid** (headless Chromium, virtual renderer, 20k rows, median of 9): filter **9.2 → 7.1 ms (−23%, 1.3×)** (from `setActiveRows` on the pipeline). Initial render is unchanged — row-create is a small share of a full render.

### DIFF
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -768,7 +768,7 @@ export default class RowManager extends CoreFeature{
 	}
 	
 	setActiveRows(activeRows){
-		this.activeRows = this.activeRows = Object.assign([], activeRows);
+		this.activeRows = activeRows.slice();
 		this.activeRowsCount = this.activeRows.length;
 	}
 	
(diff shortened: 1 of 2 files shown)


lukecotter:

azmy60 (approved):
LGTM

not-night-but (approved):

PR #4905, opened 2026-06-15
perf(events): avoid arguments/forEach allocation in event buses

lukecotter:
### What
- `_dispatch`/`_debugDispatch` (both buses): rest params + indexed loop instead of `Array.from(arguments)` + `forEach`.
- `_chain`: reuse one args array across subscribers; stock semantics preserved (array return passed as one arg; empty subscriber list returns the fallback).

### Behaviour
Unchanged — adds `InternalEventBus._chain` tests; full suite green.

### Performance
Baseline `upstream/master` @ `9539446` · branch @ `f0c6159b`.
- **Isolated** (Node, 250k calls, 5 subscribers): internal `_dispatch` 67 → 9.1 ms (**7.4×**), external `_dispatch` 67 → 7.6 ms (**8.9×**), `_chain` 65 → 10.3 ms (**6.3×**).
- **Grid** (headless Chromium, virtual renderer, 20k rows, median of 9): initial render **13.5 → 8.7 ms (−36%, 1.55×)**.

### DIFF
--- a/src/js/core/tools/ExternalEventBus.js
+++ b/src/js/core/tools/ExternalEventBus.js
@@ -72,34 +72,26 @@ export default class ExternalEventBus {
 		}
 	}
 
-	_dispatch(){
-		var args = Array.from(arguments),
-		key = args.shift(),
-		result;
-
-		if(this.events[key]){
-			this.events[key].forEach((callback, i) => {
-				let callResult = callback.apply(this.table, args);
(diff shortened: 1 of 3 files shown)


lukecotter:

azmy60 (approved):

not-night-but (approved):

PR #4906, opened 2026-06-15
perf(sort): precompute sort keys and fast-path numeric sorter

lukecotter:
### What
- `_sortItems`: extract each row's sort keys and component **once** (decorate-sort-undecorate) instead of calling `getData()` / `getFieldValue()` / `getComponent()` inside every comparison.
- number sorter: return `a - b` directly when both values are finite numbers.

### Behaviour
Ordering unchanged — adds `_sortItems` and number-sorter tests; full suite green.

### Performance
Baseline `upstream/master` @ `9539446` · branch @ `240ff91d`.
- **Isolated** (Node): `_sortItems` 1-col 234 → 90 ms (**2.6×**), 3-col 397 → 321 ms (1.2×); number sorter numeric 20.9 → 4.3 ms (**4.9×**), separators 286 → 120 ms (2.4×).
- **Grid** (headless Chromium, virtual renderer, 20k rows, median of 9): sort **31.2 → 21.6 ms (−31%, 1.44×)**.

### DIFF
--- a/src/js/modules/Sort/Sort.js
+++ b/src/js/modules/Sort/Sort.js
@@ -166,19 +166,20 @@ export default class Sort extends Module{
 				sorters=[],
 				match = false;
 				
-				if(column.modules.sort){
-					if(column.modules.sort.tristate){
-						if(column.modules.sort.dir == "none"){
-							dir = column.modules.sort.startingDir;
+				const sortModule = column.modules.sort;
+				if(sortModule){
+					if(sortModule.tristate){
+						if(sortModule.dir == "none"){
+							dir = sortModule.startingDir;
 						}else{
-							if(column.modules.sort.dir == column.modules.sort.startingDir){
-								dir = column.modules.sort.dir == "asc" ? "desc" : "asc";
+							if(sortModule.dir == sortModule.startingDir){
+								dir = sortModule.dir == "asc" ? "desc" : "asc";
 							}else{
 								dir = "none";
 							}
 						}
 					}else{
-						switch(column.modules.sort.dir){
+						switch(sortModule.dir){
 							case "asc":
 								dir = "desc";
 								break;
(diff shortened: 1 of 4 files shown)


lukecotter:

azmy60 (changes requested):
I'm not sure if I like the helper functions of the unit tests like `makeColumn`, `makeRow`, etc.. We seem to build these internal data by scratch, rather than letting the core system to build them. So if something changes in the system, these tests can behave differently.

I would look at the test/unit/modules/Sort.spec.js, which uses the Tabulator or TabulatorFull class. I'm aware that using the Tabulator class in a unit test would be counter-intuitive, but most modules (if not all) are really coupled so it's probably the cleanest approach for unit testing for now.

lukecotter:
Follow-up pushed addressing the review, plus one maintainer decision to flag.

**Test rewrite (@azmy60's ask):** `SortItems.spec.js` no longer uses hand-built `makeColumn`/`makeRow` fakes
`SortNumberSorter.spec.js` is left as-is: it doesn't reconstruct any row/column state, it just calls the pure `numberSorter`.

**One thing for you to decide - `toReversed()`:** the original commit used `self.sortList.toReversed()`. I reverted it to `.slice().reverse()`:

https://github.com/tabulator-tables/tabulator/blob/5740ccc2566023b7db51cc35d0c8ae32b1b58256/src/js/modules/Sort/Sort.js#L343-L345

`Array.prototype.toReversed()` is ES2023 (Chrome 110 / Firefox 115 / Safari 16 / Node 20). The build's `@babel/preset-env` has no `core-js`/`useBuiltIns`, so runtime methods aren't polyfilled, and the newest features currently in `src/js` are `replaceAll` / optional chaining (~ES2021). So `toReversed()` would silently raise the implicit minimum runtime 

happy to switch back to `toReversed()` if you're  happy with an ES2023 floor. Otherwise `.slice().reverse()` is behaviourally identical and runs once per sort (not in the comparator), so  no major perf cost.

PR #4907, opened 2026-06-15
perf(modules): reduce closure allocation in filter and calc loops

lukecotter:
### What
- `Filter.filterRow` / `filterRecurse`: `for...of` + early return instead of `forEach` updating a flag.
- `ColumnCalcs`: hoist the `modExists("format")` check out of the column loop; cache `column.modules.columnCalcs`.

### Behaviour
Unchanged — full suite green.

### Performance (isolated, Node, 250k rows)
- `filterRow` 3 filters 5.3 → 4.5 ms (1.2×), first-fails 6.3 → 1.7 ms (3.8×); `generateRowData` 34.8 → 17.2 ms (2.0×).

### Grid impact
Negligible — in a real grid (headless Chromium, virtual renderer) the filter/calc op is dominated by re-render, so this doesn't move it measurably. Targeted micro-optimization. Baseline `upstream/master` @ `9539446`.

### DIFF
--- a/src/js/modules/ColumnCalcs/ColumnCalcs.js
+++ b/src/js/modules/ColumnCalcs/ColumnCalcs.js
@@ -449,13 +449,14 @@ export default class ColumnCalcs extends Module{
 			
 			var cells = [];
 			
-			this.table.columnManager.columnsByIndex.forEach((column) => {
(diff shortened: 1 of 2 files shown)


lukecotter:

PR #4908, opened 2026-06-15
perf(datatree): cache module refs and flatten child recursion

lukecotter:
### What
- Cache the filter/sort module refs and `dataTreeFilter` / `dataTreeSort` flags at `initialize()`.
- `getChildren`: recurse into one shared accumulator (`_collectChildren`) — each descendant is pushed once, no per-level array + re-copy.

### Behaviour
Unchanged — full suite green.

### Performance (isolated, Node)
- `getChildren` ~250k nodes 4.8 → 3.0 ms (1.6×).

### Grid impact
Negligible — tree render is dominated by row/cell creation; traversal is a small share. Targeted micro-optimization. Baseline `upstream/master` @ `9539446`.

### DIFF
--- a/src/js/modules/DataTree/DataTree.js
+++ b/src/js/modules/DataTree/DataTree.js
@@ -45,6 +45,12 @@ export default class DataTree extends Module{
 
 	initialize(){
 		if(this.table.options.dataTree){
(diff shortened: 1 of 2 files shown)


lukecotter:

PR #4909, opened 2026-06-15
perf(frozencolumns): reinitialize rows in O(n)

lukecotter:
### What
`FrozenColumns.reinitializeRows`: build a `Set` of the visible rows for the `otherRows` exclusion instead of `Array.includes` in a `filter` — O(rows × visible) → O(rows).

### Behaviour
Unchanged — adds a `reinitializeRows` test; full suite green.

### Performance (isolated, Node, 100k rows / 50 visible)
- `reinitializeRows` 2.17 → 1.44 ms (1.5×).

### Grid impact
Negligible — the saving (<1 ms) is a small fraction of a full redraw (~38 ms at 100k rows), within noise (headless Chromium, virtual renderer). Targeted micro-optimization that scales with rows × rendered-window. Baseline `upstream/master` @ `9539446`.


lukecotter:

PR #4918, opened 2026-07-05
Fix popup viewport positioning

### LINKED ISSUE #4285: headerFilter 'list' position
Is it possible to control the positioning of the input 'list' element used in a headerFilter with **Tabulator 5.5.2**?

I created a very basic [example on codepen](https://codepen.io/dbauszus-glx/pen/RwEaQLq):

```js
new Tabulator(document.getElementById("Table"), {
  layout: "fitColumns",
  columns: [
    {
      title: "Number",
      field: "number"
    },
    {
      title: "Letter",
      field: "letter",
      headerFilter: "list",
      headerFilterFunc: "in",
      headerFilterParams: {
        values: ["A", "B", "C", "D"],
        multiselect: true,
        clearable: true
      }
    }
  ],
  data: data
});
```

The list input is off the screen...
image

nickswalker:
Body-hosted popups were using `document.body.offsetWidth` as their horizontal boundary. When the body was narrower than the viewport, popups near the body edge could be treated as overflowing and repositioned incorrectly, even though they still fit within the visible viewport. The new bounds use viewport coordinates for body-hosted popups, including horizontal scroll offset, so page-coordinate popup positions are compared against the actual visible left/right edges.

Previously, overflow handling cleared `left` and set `right`. That worked for some overflow cases, but it could leave stale `right` positioning behind when the same popup element was reused. The new code normalizes horizontal placement back to an explicit `left` value and clears `right` on each show/fit pass.

The `reversedX` reset is limited to root popups. Child popups keep their reversal state so nested menus can continue opening consistently on the opposite side once they have had to reverse near an edge.

Closes #4285 (hopefully for good!). I don't have a enough understanding of the library to be confident that these changes don't cause problems elsewhere. They work well in my uses.

Before
Screenshot 2026-07-05 at 13 31 20
Screenshot 2026-07-05 at 13 31 59

After
Screenshot 2026-07-05 at 13 29 44
Screenshot 2026-07-05 at 13 32 37

### DIFF
--- a/src/js/core/tools/Popup.js
+++ b/src/js/core/tools/Popup.js
@@ -135,6 +135,10 @@ export default class Popup extends CoreFeature{
 		if(this.destroyed || this.table.destroyed){
 			return this;
 		}
+
+		if(!this.parent){
+			this.reversedX = false;
+		}
 		
 		if(origin instanceof HTMLElement){
 			parentEl = origin;
@@ -153,12 +157,11 @@ export default class Popup extends CoreFeature{
 			
 			x = coords.x;
 			y = coords.y;
-			
-			this.reversedX = false;
 		}
 		
 		this.element.style.top = y + "px";
 		this.element.style.left = x + "px";
+		this.element.style.right = "";
 		
 		this.container.appendChild(this.element);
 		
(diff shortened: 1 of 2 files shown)


rathboma:
@nickswalker Thank you for this fix! We agree it's hard to tell if you broke something else. Azmi has another branch to make this part of the library more testable, so hopefully that should allow us to merge in your changes soon.

rathboma:
Play - Day - merge this with Azmi's testing branch, see if all tests pass, add some more tests (try and break it!) then merge if good.

not-night-but:
Superseded by #4969

Issue #4920, opened 2026-07-09
Giving an ID to tabulator-table-body results in duplicate IDs in pages with multiple tables.


btdonovan:
**Describe the bug**
In src/js/core/RowManager.js the tabulator-table element is given the ID "tabulator-table-body". In pages that render multiple tables that results in duplicate IDs in the DOM. 

IDs must be unique so if this needs to be an ID it should be given a pseudo-random suffix. If this doesn't need to be an id this line should probably be dropped.

It looks like the ID was added so that the div the table is initialized in can aria-own the body.

src/js/core/RowManager.js line 64
src/js/core/Tabulator.js line 235

https://github.com/tabulator-tables/tabulator/blame/795691a07a63ba88cfba457ed6452b99d8d285fc/src/js/core/RowManager.js#L64

**Tabulator Info**
6.5.2

**Working Example**
YOU MUST include a link to a JS Fiddle or Codepen that demonstrates the problem, it is very hard to diagnose an issue from a simple description.

https://jsfiddle.net/fnr8y54q/4/

**To Reproduce**
In a page with two tables, inspect the table bodies in the browser dev tools. Right click the first cell in the first row in the first table (not the table header, the body) and select inspect to open the elements tab in chrome or the inspector in firefox.

The first cell will be nested in the first tabulator-row element which is nested in the tabulator-table-body element. Note that element has a static ID "tabulator-table-body"

Now inspect the first cell in the first row of the second table (not the table header, the body). Note that this table's tablulator-table-body element also reuses the identical ID "tabulator-table-body"

**Expected behavior**
Tables should not reuse identical IDs. If an ID must be used it should be derived from the ID of the div the table is initialized in. Example: If the table is initialized on a div with ID "#myTable" the tabulator-table-body could use ID "#myTable-tabulator-table-body" or it could use a pseudo random suffix such as "#tabulator-table-body-m0t2b" where the m0t2b suffix is generated randomly at initialization.

**Desktop (please complete the following information):**
 - OS: not applicable
 - Browser: Chrome or Firefox
 - Version Chrome 149.0.7827.200, Firefox 152.0.4

**Smartphone (please complete the following information):**
Not tested but again, not applicable.

btdonovan:
Resolved by merge.

PR #4923, opened 2026-07-13
Fix identical IDs in every table instance

btdonovan:
This fixes issue #4920

This bug was introduced in 6.4.0 when a non-random non-unique ID was applied to the tabulator-table-body in RowManager.js. This was done so that the table element could aria-own the tabulator-table-body. This results in non-unique duplicate IDs in the DOM in pages which render more than one tabulator-table.

The fix captures the ID of the table element if it exists and generates a pseudo-random suffix if it does not. The value is stored in this.instanceId.

The value is then appended to the aria-owns attribute in Tabulator.js and to the ID attribute in RowManager. This ensures that the ID will either be based on the ID of the owning table or randomized enough that it will be very unlikely for two tabulator-table-body elements to have the same ID while still allowing the body to be aria-owned by the table element.

### DIFF
--- a/src/js/core/RowManager.js
+++ b/src/js/core/RowManager.js
@@ -61,7 +61,7 @@ export default class RowManager extends CoreFeature{
 		
 		el.classList.add("tabulator-table");
 		el.setAttribute("role", "rowgroup");
-		el.setAttribute("id", "tabulator-table-body");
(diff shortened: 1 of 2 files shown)


rathboma:
Thanks for the fix!

PR #4931, opened 2026-08-01
fix(render): four state-tracking fixes in the vertical virtual renderer

lukecotter:
### What this is

Four state fixes in `VirtualDomVertical`. Originally opened as 4 seperate PRs (#4931–#4934)

Four values in `VirtualDomVertical` are wrong on master. No visually impact putrely correctness but they fix render-window state that #4937 relies on.

| Wrong value | Where | Is | Should be |
| --- | --- | --- | --- |
| `this.rows.length - 1` | `renderers/VirtualDomVertical.js:101` | `-1`. `rows()` (`Renderer.js:94`) takes no parameters, so `Function.prototype.length` is `0` | index of the last row |
| `this.vDomScrollHeight` | read at `:137`, first assigned at `:366` | `undefined` until the first fill, so `undefined - scrollTop` is `NaN` and the `> vDomWindowBuffer` test is always `false` | `0` |
| `vDomTopPad` after a filter rerender | `rerenderRows` | a stale pre-filter `topOffset`, inflating `paddingTop` into a blank strip | recomputed against the post-filter rows |
| `vDomTopPad += paddingAdjust` | `:510` | unclamped, unlike the add path at `:457-466` | `Math.max(v, 0)` |

### Verification

Here is a vide to show the scroll to row fix part.
[scroll-to-row.webm](https://github.com/user-attachments/assets/3a92e3be-f2e4-4ec5-beeb-89f465e28f93)

No Performance impact: tested on 500k rows.

Part of #4938.

### DIFF
--- a/src/js/core/rendering/renderers/VirtualDomVertical.js
+++ b/src/js/core/rendering/renderers/VirtualDomVertical.js
@@ -106,14 +106,14 @@ export default class VirtualDomVertical extends Renderer{
 			//The anchor scan above used the PRE-callback (e.g. pre-filter) window
 			//indices. If that window now points past the new row count, topRow/
 			//topOffset are stale and would inflate vDomTopPad into a blank strip
-			//across the top. In that case (or when no anchor row was found) do a
-			//fresh fill, which resets vDomTopPad to 0.
+			//across the top. In that case do a fresh fill, which resets
+			//vDomTopPad to 0.
 			var windowInvalid = this.vDomTop >= newRows.length || this.vDomBottom >= newRows.length;
 
-			if(topRow === false || windowInvalid){
+			if(windowInvalid){
 				this._virtualRenderFill();
 			}else{
-				this._virtualRenderFill(topRow, true, topOffset || 0);
+				this._virtualRenderFill((topRow === false ? newRows.length - 1 : topRow), true, topOffset || 0);
 			}
 		}else{
 			this.clear();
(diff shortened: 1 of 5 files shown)


rathboma:
@lukecotter love all these PRs, but it's hard for me to evaluate the impact of them all, or to know if you've had any hand in them, or tested them, beyond your AI agent.

Any way you could add a little video to each one showing what each PR does, or add a Luke-written comment with context for the fix and how you've tested it? For example - was this just a fun activity for you? Are you using this in production?

lukecotter:
@rathboma  
 Of course I can do all that and I'll tone down letting ai comment.

Yes I have manually tested them all and run them through a chrome dev tools perf profile.

They are actually all fixes I applied to a custom version of the renderer I wrote to fix some bugs. One example was no rows being shown when the scroll is dragged down then quickly scrolled up.

That new renderer also improved performance quite a bit but it had gaps in behaviour and is obviously a sync and maintenance annoyance not to mention a potential upgrade nightmare so I thought it better to port the fixes so I can switch back to stock. I also opened some prs which improve performance and I have some more improvements that could be migrated from the custom version but I was hoping get some of these in first to avoid bombarding you with too much and make sure they other changes are stable first. Plus I want to re benchmark once these all land.

rathboma:
Thanks for being awesome!

I do “tabulator tuesdays”, so I’ll take a look at all of these on Tuesday :-)

yeah the more info you can provide on each pr of the before and after, the easier it is for me to trust and merge them

PR #4932, opened 2026-08-01
fix(render): use correct fallback row index in rerenderRows

lukecotter:
### Problem

`rerenderRows` falls back to the last row index when `topRow === false`, but writes
`this.rows.length - 1`. `this.rows` is a zero-arity **method**, so `.length` is 0 and
the expression is always `-1`.

### Fix

Use `this.rows().length - 1` — the last display-row index.

### Performance

Neutral. 500k rows, K=5, medians (measured together with the follow-up filter-window
fix, see the dependent PR):

| Metric | Before | After |
| --- | --- | --- |
| initial render (ms) | 97.6 | 103.2 |
| initial render, variable heights (ms) | 105.6 | 108.1 |

### DIFF
--- a/src/js/core/rendering/renderers/VirtualDomVertical.js
+++ b/src/js/core/rendering/renderers/VirtualDomVertical.js
@@ -98,7 +98,9 @@ export default class VirtualDomVertical extends Renderer{
 		}
 
 		if(this.rows().length){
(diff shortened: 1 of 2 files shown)


lukecotter:

PR #4933, opened 2026-08-01
fix(render): keep content flush after a filter rerender

lukecotter:
### Problem

After a filter rerender the rendered window could leave a blank strip at the top or the
bottom of the viewport, because the window was not re-flushed against the new row set.

### Fix

Keep content flush after a filter rerender.

### Depends on

The `this.rows()` fallback-index fix, which is included here as the first commit. Merge
that PR first; this one then reduces to a single commit.

### Test

`test/e2e/rerender-filter.spec.ts` asserts no top or bottom blank strip.

### Performance

Neutral. 500k rows, K=5, medians:

| Metric | Before | After |
| --- | --- | --- |
| initial render (ms) | 97.6 | 103.2 |
| initial render, variable heights (ms) | 105.6 | 108.1 |

### DIFF
--- a/src/js/core/rendering/renderers/VirtualDomVertical.js
+++ b/src/js/core/rendering/renderers/VirtualDomVertical.js
@@ -97,10 +97,21 @@ export default class VirtualDomVertical extends Renderer{
 			callback();
 		}
 
-		if(this.rows().length){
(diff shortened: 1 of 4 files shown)


lukecotter:

PR #4934, opened 2026-08-01
fix(render): clamp vDomTopPad to zero when removing top rows

lukecotter:
### Problem

In `_removeTopRow`, accumulated **actual** row heights can push `vDomTopPad` slightly
negative. That sets a negative `paddingTop` and lets rendered content drift above the
viewport.

### Fix

Clamp `vDomTopPad` to zero after the adjustment. One line.

### A note on test coverage

An e2e test written for this passed **with and without** the fix, so it was deleted
rather than kept as misleading coverage. The failure is a sub-pixel drift that the
existing harness cannot resolve; the reasoning is in the diff comment instead.

### Performance

Neutral. 500k rows, K=5, medians:

| Metric | Before | After |
| --- | --- | --- |
| initial render (ms) | 97.6 | 102.0 |
| initial render, variable heights (ms) | 105.6 | 102.9 |

### DIFF
--- a/src/js/core/rendering/renderers/VirtualDomVertical.js
+++ b/src/js/core/rendering/renderers/VirtualDomVertical.js
@@ -509,6 +509,7 @@ export default class VirtualDomVertical extends Renderer{
 
 		if(paddingAdjust){
 			this.vDomTopPad += paddingAdjust;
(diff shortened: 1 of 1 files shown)


lukecotter:

PR #4971, opened 2026-09-17
fix(render): don't tear down the rendered rows while the table is not visible

Teadeno:
### Summary

A table inside an iframe that is hidden with `display:none` (tab switching in the host page)
comes back blank, partially rendered, or scrolled to the wrong place. It recovers as soon as
the table is redrawn by hand, so the damage is done by the redraw the library triggers itself
while the table has no measurable size - a hidden iframe still receives resize observer
callbacks.

### Root cause

1. `VirtualDomVertical._virtualRenderFill()` removes every rendered row
   (`while(element.firstChild) element.removeChild(element.firstChild)`) *before* it checks
   `Helpers.elVisible(this.elementVertical)`. When that check fails, all rows are gone while
   `vDomTop`/`vDomBottom` still hold the previous window.
2. `rerenderRows()` looks for its anchor by comparing `scrollTop` against
   `rows[i].getElement().offsetTop`. A row that is no longer in the DOM reports
   `offsetTop === 0`, so it is picked as the anchor and the table is rendered against a
   position that does not exist. When no row could be anchored, the window was assumed to be
   past the end of shrunken data and the last row was used, which puts the view at the bottom.

### Changes

- `_virtualRenderFill()` bails out before the teardown when the holder is not visible, so the
  rendered rows and the render window are kept until the table can be laid out again.
- `rerenderRows()` skips detached rows when looking for the anchor. If no live row can be
  anchored while the window still points at live rows, the rows were torn down (data reload,
  or a render skipped while the table was hidden), so the table starts from the top; a window
  past the end of shrunken data still keeps the view at the end.

Two commits, each with its own test, so they can be taken separately.

### Tests

`test/unit/core/rendering/VirtualDomVertical.spec.js` - 2 new cases on the existing jsdom
box-model harness. Both fail on master and pass with these patches.
`npm run test:unit`: 539/539 passing on this branch.

### Notes

Only `src/` and `test/` are touched, no `dist/` changes, no API changes.
An alternative would be to skip `redrawTable()` when the resize observer reports a zero size;
the patch above keeps the renderer responsible for its own state, which also covers the
data-reload case.


src/js/core/rendering/renderers/VirtualDomVertical.js
```diff
@@ -99,9 +107,12 @@ export default class VirtualDomVertical extends Renderer{
...
 			if(topRow === false){
-				//no rendered row to anchor on, start from the top if nothing was rendered before,
-				//otherwise the rendered window is past the end of the shrunken data so anchor on the last row
-				topRow = rows.length ? this.rows().length - 1 : 0;
+				//no rendered row to anchor on. When the window still points at live rows the rendered
+				//rows were torn down (the data was reloaded, or the render was skipped because the
+				//table was hidden) so the window means nothing and the table starts from the top;
+				//otherwise the window is past the end of the shrunken data so keep the view at the end
+				topRow = rows[this.vDomTop] ? 0 : (rows.length ? this.rows().length - 1 : 0);
+				topOffset = 0;
```
not-night-but: I think this top level tertiary could just be an if/else statement.

src/js/core/rendering/renderers/VirtualDomVertical.js
```diff
@@ -250,6 +261,15 @@ export default class VirtualDomVertical extends Renderer{
 		avgRowHeight = this.table.options.rowHeight, 
 		resized = true;
 
+		//the table is not visible so nothing can be measured or rendered: bail out before the
+		//rendered rows are torn down and the render window is updated, otherwise the table is
+		//left empty holding a stale window that the next (visible) redraw anchors on and
+		//mispositions (an iframe hidden with display:none still receives resize observer
+		//callbacks, which trigger a redraw in exactly this state)
```
not-night-but: If we could tone down the verbose comments that would be great. Especially in an instance like this where it feels almost self-explanatory (it makes sense not to render if the element is not visible, y'know?)

